PR #1, opened 2016-10-08
Add build scripts

moodmosaic:
This Pull Request adds [Paket](https://fsprojects.github.io/Paket/) as a dependency manager, and [FAKE - F# Make](http://fsharp.github.io/FAKE/) as the DSL for build tasks.

**Usage**

```
$ ./build.sh
```

or in the Command Prompt (Windows)

```
build.cmd
```


jystic:
Sounds good, let me just check I can still build everything with this setup (I'm on OS/X, using Xamarin Studio, although I'd be happy to use vim if it comes to it, that's what I use for everything else)

jystic:
This works for me after a few changes which I can make after I merge this.

Can I ask what is your setup is? are you running on Windows with msys/mingw for your shell, or just the windows command prompt?

The `build.sh` doesn't work for me unless I tell it to run the executable with `mono`.

moodmosaic:
I'm currently on Windows, with Visual Studio 2015 and Visual Studio Code, using both mingw and cmd.exe 😃

moodmosaic:
Hmm... I think it's normal to run the build via Mono outside Windows? :confused:

jystic:
So I had to change the `build.sh` to this to make it work for me:
```sh
#!/bin/sh -eu

mono .paket/paket.bootstrapper.exe
exit_code=$?
if [ $exit_code -ne 0 ]; then
  exit $exit_code
fi

mono .paket/paket.exe restore
exit_code=$?
if [ $exit_code -ne 0 ]; then
  exit $exit_code
fi

mono packages/FAKE/tools/FAKE.exe
```
But I suspect that will break it for you if you're running `build.sh` on Windows

jystic:
But if you're just using `build.cmd` then all is good, I can make the change for me and it won't affect you

moodmosaic:
I think that should be fine for the beginning. 👍 We may then make `build.sh` more generic as we go along.

jystic:
ok, great 👍  I'll merge this and make my fixes

I did some googling, looks like the environment variable `COMSPEC` should be set on Windows alone, so we can check for that and if so run without mono

moodmosaic:
Sounds good! I'm currently on my phone with limited internet access, but I'll do some googling to this too.

PR #3, opened 2016-10-08
Add marker file so the bootstrapper is only run every 24 hours

jystic:
@moodmosaic would you mind checking if this new `build.sh` works as expected on Windows?

The general idea is that the bootstrapper is only run if it's been more than 24 hours since the last run. Would be nice to know if the `COMSPEC` trick works as well.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -13,4 +13,4 @@ if ! test -f .paket/paket.bootstrapper.run || find .paket/paket.bootstrapper.run
 fi
 
 $MONO .paket/paket.exe restore
-$MONO packages/FAKE/tools/FAKE.exe
+$MONO packages/FAKE/tools/FAKE.exe $@
--- a/build.sh
+++ b/build.sh
@@ -6,6 +6,11 @@ else
   MONO=""
 fi
 
-$MONO .paket/paket.bootstrapper.exe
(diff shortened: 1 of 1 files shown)


jystic:
Oh I changed the shebang to use `sh -eu`, the `-e` means that we don't need to check for errors, the whole script will bomb out as soon as one of the commands in it fails.

moodmosaic:
I'm getting this when running `build.sh` after doing a `git clean`:

```
$ git clean -dxf
$ /build.sh: line 15: .paket/paket.exe: No such file or directory
```

jystic:
ah oops, I committed the `.run` file, moment

jystic:
ok, try the latest master

moodmosaic:
Alright, #4 is merged. I'll try.

moodmosaic:
👍

```
Build succeeded.
    0 Warning(s)
    0 Error(s)

Time Elapsed 00:00:01.35
Finished Target: Build

---------------------------------------------------------------------
Build Time Report
---------------------------------------------------------------------
Target     Duration
------     --------
Build      00:00:01.5171739
Total:     00:00:01.5475552
Status:    Ok
---------------------------------------------------------------------
```

jystic:
excellent 👍

moodmosaic:
Yep! Works fine 👍

PR #7, opened 2016-10-09
Unix fixups

jystic:


.travis.yml
```diff
@@ -0,0 +1,10 @@
+language: csharp
```
moodmosaic: I think we want language _fsharp_ there (?)

jystic: I'm not sure if travis supports that, I copied this from [FAKE](https://github.com/fsharp/FAKE/blob/master/.travis.yml)

moodmosaic: Oh, OK

PR #8, opened 2016-10-09
Rename Jack.UnitTests -> Jack.Tests

jystic:
@moodmosaic hope you don't mind


moodmosaic:
It should be fine, at least as long as we have a single test suite. 😉

jystic:
I think we had multiple I'd prefer `Jack.Tests.X`, `Jack.Tests.Y` (or `Jack.Test.X`, `Jack.Test.Y`)

moodmosaic:
OK, that should work 👍

PR #9, opened 2016-10-09
Implement Gen.bind directly instead of using promote

jystic:
I think we could almost get rid of Random altogether now, I only added it to simplify implementing bind for `Gen` but after some simplifications it's fairly straightforward.

### DIFF
--- a/Jack/Gen.fs
+++ b/Jack/Gen.fs
@@ -27,13 +27,17 @@ module Gen =
         mapRandom (Random.map f) g
 
     let map (f : 'a -> 'b) (g : Gen<'a>) : Gen<'b> =
-        mapTree (Tree.map f) g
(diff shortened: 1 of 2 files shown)


moodmosaic:
Actually, shouldn't `sized`, `resize`, `choose`, and `replicate` belong to the `Gen` module too?

jystic:
Yeah those will be implemented in terms of the equivalent combinators in `Random`, `Gen` already has `choose` implemented using the one from `Random` for example.

But it's possible that `Gen` could simply become `Seed -> Size -> Tree<'a>` instead of being `Random<Tree<'a>>`. I did try that initially but it complicated a few things which maybe aren't a problem anymore. What I call `Random` is traditionally called `Gen` in QuickCheck implementations.

PR #10, opened 2016-10-09
I think Gen.constant and Tree.singleton are clearer names than result

jystic:
Too bad both 'return' and 'pure' are reserved words in F#, oh well

### DIFF
--- a/Jack/Gen.fs
+++ b/Jack/Gen.fs
@@ -17,8 +17,8 @@ module Gen =
     let create (shrink : 'a -> LazyList<'a>) (random : Random<'a>) : Gen<'a> =
         Random.map (Tree.unfold id shrink) random |> ofRandom
 
-    let result (x : 'a) : Gen<'a> =
-        Tree.result x |> Random.result |> ofRandom
+    let constant (x : 'a) : Gen<'a> =
+        Tree.singleton x |> Random.constant |> ofRandom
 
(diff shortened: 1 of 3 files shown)


Jack/Gen.fs
```diff
@@ -17,8 +17,8 @@ module Gen =
     let create (shrink : 'a -> LazyList<'a>) (random : Random<'a>) : Gen<'a> =
         Random.map (Tree.unfold id shrink) random |> ofRandom
 
-    let result (x : 'a) : Gen<'a> =
-        Tree.result x |> Random.result |> ofRandom
+    let constant (x : 'a) : Gen<'a> =
+        Tree.singleton x |> Random.constant |> ofRandom
```
moodmosaic: 👍

moodmosaic: Indeed. And other candidate names could be `inject`, `retn` (`return` is a reserved word), and so on.

jystic: Yeah, so I like `singleton` for the `Tree`, but I'm not sure about the other ones yet. My thinking was that if it can't be the "right" word (pure or return) then I should make it something which makes sense for the given concept. For `Gen`/`Random` I think it means `constant`.

moodmosaic: Yes, `constant` is sounds fine for `Gen` and `Random`.

PR #11, opened 2016-10-09
Bare bones module for checking properties

jystic:
If you run the first part of the `Script.fsx`:
```fs
// specialized computation expression 'forAll' which
// is like 'gen' but keeps track of bound arguments so
// they can be reported in counterexamples
Property.check <| forAll {
    let! x = Gen.choose 1 100
    let! y = Gen.elements [ "a"; "b"; "c"; "d" ]
    return! x < 50 || y = "a"
}
```

You get output like this:
```
 *** Failed! Falsifiable (after 2 tests and 5 shrinks):
 50
 "b"
 val it : bool = false
```

@moodmosaic I think this is nice for scripts, and it works well in a Haskell setting, but I imagine there's a more idiomatic way to report test errors for things like xUnit that I'm not really familiar with. I would appreciate any guidance you have on that matter. I'm full time on Haskell these days, it's been a number of years since I was a C# developer and back then we used NUnit exclusively, but times have changed it seems. 😄

### DIFF
--- a/Jack/Jack.fsproj
+++ b/Jack/Jack.fsproj
@@ -46,6 +46,7 @@
     <Compile Include="Random.fs" />
     <Compile Include="Shrink.fs" />
     <Compile Include="Gen.fs" />
+    <Compile Include="Property.fs" />
   </ItemGroup>
   <ItemGroup>
     <Compile Include="AssemblyInfo.fs" />
--- a/Jack/Property.fs
+++ b/Jack/Property.fs
@@ -1,0 +1,135 @@
+﻿#light
+namespace Jack
+
+open FSharpx.Collections
+
+type Result =
+    | Failure of List<string>
+    | Success
+
+type Property =
+    | Property of Gen<Result>
+
+[<CompilationRepresentation(CompilationRepresentationFlags.ModuleSuffix)>]
+module Result =
+    let isFailure (x : Result) : bool =
+        match x with
+        | Failure _ ->
+            true
+        | Success ->
+            false
+
+    let mapFailure (f : List<string> -> List<string>) (x : Result) : Result =
+        match x with
+        | Failure msgs ->
+            Failure (f msgs)
+        | Success ->
+            Success
+
+[<CompilationRepresentation(CompilationRepresentationFlags.ModuleSuffix)>]
+module Property =
+
+    let ofGen (x : Gen<Result>) : Property =
+        Property x
+
+    let toGen (Property x : Property) : Gen<Result> =
+        x
+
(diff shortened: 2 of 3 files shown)


moodmosaic:
@jystic actually having a `script.fsx` for *exploration* is idiomatic in F# 😃

To next step is to add some test coverage (in test modules). Any tests we add could be either run as a whole test suite, by adding a build test in `build.fsx`, or by using the [xUnit.net Console Runner](http://xunit.github.io/docs/getting-started-desktop#add-xunit-runner-ref), or even by using some tool inside the code editor/IDE.

---

FWIW, I'm using [this](http://www.testdriven.net/) tool, which even allows me to right click on a single test and run it (and even debug it, though I rarely rely on this when writing F# properties). But it's available only as a Visual Studio add-in. It's not even currently available in Visual Studio Code :disappointed:

jystic:
> FWIW, I'm using [this](http://www.testdriven.net/) tool, which even allows you to right click on a single test and run it (and even debug it, though I rarely rely on this when writing F# properties). But it's available only as a Visual Studio add-in. It's not even currently available in Visual Studio Code :disappointed:

Ah neat, yeah I actually used to use that when I was doing C# development

jystic:
> @jystic actually having a `script.fsx` for exploration *is* idiomatic in F# 😃

I imagine that we also need a way of running Jack properties in a build system though, via NUnit or xUnit? I'm thinking of users of this library. Something like a `Jack.XUnit` and `Jack.NUnit` assemblies? /not sure

> The next step could be adding some test coverage (in test modules). So any tests we add could be either run as a whole test suite, by adding a build test in `build.fsx`, or by using the [xUnit.net Console Runner](http://xunit.github.io/docs/getting-started-desktop#add-xunit-runner-ref), or even by using some tool inside the code editor/IDE.

Sounds good to me

moodmosaic:
>I imagine that we also need a way of running Jack properties in a build system though, via NUnit or xUnit? 

Absolutely. I've now opened #12 for this.

moodmosaic:
>Something like a Jack.XUnit and Jack.NUnit assemblies? /not sure

IMHO, we shouldn't need those, unless we want to *integrate* Jack *with* xUnit.net and/or NUnit, by providing some kind of Jack-specific attributes that turn regular xUnit.net/NUnit tests into Jack Properties.

But then, we have to make sure we're compatible with whatever these tools are doing, which means significant maintenance overhead―based on my experience in AutoFixture.Xunit, AutoFixture.Xunit2, AutoFixture.NUnit2, and AutoFixture.NUnit3― so I wouldn't just go into this direction (yet).

PR #12, opened 2016-10-09
Add basic infrastructure for discovering and running tests

moodmosaic:


paket.lock
```diff
@@ -7,3 +7,131 @@ NUGET
       FsControl (>= 2.0.0-CI00075)
     FSharpx.Collections (1.15.2)
       FSharp.Core
+    Microsoft.NETCore.Platforms (1.0.1) - framework: dnxcore50
```
moodmosaic: These added changes tell me that we're on an older version of F# (4.4.0.0). I have a feeling that in the next PR, where I'll be adding some actual tests, we'll have to switch to a newer version of F#, at least 4.4.3.0 or newer.

jystic:
Looks good 👍

PR #13, opened 2016-10-09
Switch to F# 4.4.0

moodmosaic:
because we were on an old version (4.0.0), and we were also referencing an old [NuGet Package](https://www.nuget.org/packages/FSharp.Core/) which isn't updated anymore.

See also [this](https://github.com/jystic/dotnet-jack/pull/12#discussion_r82514415) comment in #12.


jystic:
Nice 👍

jystic:
Rebased and merged

PR #14, opened 2016-10-09
Remove #light directive,

moodmosaic:
since the default syntax is the lightweight syntax.

See also: https://docs.microsoft.com/en-us/dotnet/articles/fsharp/language-reference/verbose-syntax


jystic:
Cool, not sure why I thought I needed this 👍

jystic:
Rebased and merged (shame github doesn't have an option for that, they have all the options except the one I want)

PR #15, opened 2016-10-10
Add an example-based test for Shrink.removes

moodmosaic:
This pull request adds a test module for writing example-based tests, as well as property-based tests, for the functions in the `Shrink` module.

Adding the very first test required setting ['redirects: on' in Paket](https://fsprojects.github.io/Paket/dependencies-file.html#Redirects-option), as well as referencing FSharpx.Collections.

### DIFF
--- a/Jack.Tests/paket.references
+++ b/Jack.Tests/paket.references
@@ -1,3 +1,5 @@
 xunit.runner.console
 xunit.core
 Unquote
+
+FSharpx.Collections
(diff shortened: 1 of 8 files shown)


Jack.Tests/paket.references
```diff
@@ -1,3 +1,5 @@
 xunit.runner.console
 xunit.core
 Unquote
+
+FSharpx.Collections
```
moodmosaic: I'm separating (with a blank-line) the references required for test-discovery and assertions from the references required by the actual test code.

jystic: 👍

jystic:
Fantastic 🌴

PR #20, opened 2016-10-12
Add combinators for shrinking, choice, numbers, strings, collections

jystic:
I decided to diverge from some of the traditional names for things to try an make them fit better into an F# style.

Haskell QuickCheck -> F# Jack
`elements` -> `Gen.item`
`choose` -> `Gen.range`
`oneof` -> `Gen.choice`
`listOf` -> `Gen.list` / `Gen.array` / `Gen.seq`

@moodmosaic Would be interested to know what you think.

### DIFF
--- a/Jack/Gen.fs
+++ b/Jack/Gen.fs
@@ -1,6 +1,7 @@
 ﻿namespace Jack
 
 open FSharpx.Collections
+open Jack.Numeric
 
 /// A generator for values and shrink trees of type 'a.
 type Gen<'a> =
@@ -13,20 +14,14 @@ module Gen =
     let toRandom (Gen r : Gen<'a>) : Random<Tree<'a>> =
         r
 
+    let delay (f : unit -> Gen<'a>) : Gen<'a> =
+        Random.delay (toRandom << f) |> ofRandom
+
     let create (shrink : 'a -> LazyList<'a>) (random : Random<'a>) : Gen<'a> =
         Random.map (Tree.unfold id shrink) random |> ofRandom
 
     let constant (x : 'a) : Gen<'a> =
         Tree.singleton x |> Random.constant |> ofRandom
-
-    let mapRandom (f : Random<Tree<'a>> -> Random<Tree<'b>>) (g : Gen<'a>) : Gen<'b> =
-        toRandom g |> f |> ofRandom
-
-    let mapTree (f : Tree<'a> -> Tree<'b>) (g : Gen<'a>) : Gen<'b> =
-        mapRandom (Random.map f) g
-
-    let map (f : 'a -> 'b) (g : Gen<'a>) : Gen<'b> =
-        mapTree (Tree.map f) g
 
     let private bindRandom (m : Random<Tree<'a>>) (k : 'a -> Random<Tree<'b>>) : Random<Tree<'b>> =
         Random <| fun seed0 size ->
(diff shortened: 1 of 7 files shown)


Jack/Shrink.fs
```diff
@@ -97,11 +98,26 @@ module Shrink =
...
             LazyList.empty
         else
-            let two = LanguagePrimitives.GenericOne + LanguagePrimitives.GenericOne
+            let one : ^a = LanguagePrimitives.GenericOne
+            let two : ^a = one + one
 
             /// We need to halve our operands before subtracting them as they may be using
             /// the full range of the type (i.e. 'MinValue' and 'MaxValue' for 'Int32')
-            let diff = (x / two) - (destination / two)
+            let diff : ^a = (x / two) - (destination / two)
```
jystic: I added these annotations so that the compiler infers them all to be the same unified type rather than generating piles of constraints

Jack/Gen.fs
```diff
@@ -41,18 +36,69 @@ module Gen =
...
+            Node (x, LazyList.empty)
+        mapTree drop g
+
+    /// Apply an additional shrinker to all generated trees.
+    let shrinkLazy (f : 'a -> LazyList<'a>) (g : Gen<'a>) : Gen<'a> =
+        mapTree (Tree.expand f) g
+
+    /// Apply an additional shrinker to all generated trees.
+    let shrink (f : 'a -> List<'a>) (g : Gen<'a>) : Gen<'a>  =
+        shrinkLazy (LazyList.ofList << f) g
```
jystic: This combinator is pretty neat I think 😄 You can apply an extra shrinking function to an existing generator - there's an example in the script.

moodmosaic:
>elements -> Gen.item

Sounds reasonable. What happens though if we want to add something similar to [growingElements](https://hackage.haskell.org/package/QuickCheck-2.9.2/docs/src/Test-QuickCheck-Gen.html#growingElements)?

>choose -> Gen.range

Sounds good. I find `Gen.range` more descriptive. (Otherwise one may need to read the comments.)

>oneof -> Gen.choice

This can be confusing, IMHO, because of [the built-in `Choice<'T1,'T2,..>`](https://msdn.microsoft.com/visualfsharpdocs/conceptual/microsoft.fsharp.core-namespace-%5bfsharp%5d) group of types.

>listOf -> Gen.list / Gen.array / Gen.seq

This makes sense to me, since F# has quite a few (..!) collection types: `list`, `array`, `seq`, `set`, `map`, etc. Maybe we even need to consider adding a `Gen.set` one.

Jack/Gen.fs
moodmosaic: Sounds good! I'm planning to check this and the other added functions of this PR a bit later on 😃

Jack/Numeric.fs
```diff
@@ -0,0 +1,24 @@
...
+    let inline maxValue () : ^a =
+        MaxValue.Invoke ()
+
+    /// Converts from a BigInt to the inferred destination type.
+    let inline fromBigInt (x : bigint) : ^a =
+        FromBigInt.Invoke x
+
+    /// Converts to a BigInt.
+    let inline toBigInt (x : ^a) : bigint =
+        ToBigInt.Invoke x
```
moodmosaic: Pretty much decent example of how FsControl (via statically resolved type params) can be useful IRL 😃 👍

moodmosaic:
Great job! 👍

jystic:
> >elements -> Gen.item

> Sounds reasonable. What happens though if we want to add something similar to [growingElements](https://hackage.haskell.org/package/QuickCheck-2.9.2/docs/src/Test-QuickCheck-Gen.html#growingElements)?

I guess `growingItem` wouldn't be so bad, `elements` isn't such a bad name in any case I suppose. I went for `item` as it was similar in function to `List.item` / `Array.item` / `Seq.item`.

Do you have a use case for `growingElements` by chance? I have never used it myself so I can't imagine the scenario.

> >oneof -> Gen.choice

> This can be confusing, IMHO, because of [the built-in `Choice<'T1,'T2,..>`](https://msdn.microsoft.com/visualfsharpdocs/conceptual/core.choice%5b%27t1%2c%27t2%5d-union-%5bfsharp%5d) group of types.

I named this one after `choice` in FParsec, it has similar behaviour in that it takes `seq<'m<'a>> -> 'm<'a>`.

In some ways I think it could fit in with the built-in choice, as we could have combinators like `choice2 : Gen<'a> -> Gen<'b> -> Gen<Choice<'a, 'b>>`, etc:

```fs
let choice2 gx gy =
    choice [
        Gen.map Choice1Of2 gx
        Gen.map Choice2Of2 gy
    ]
```

> >listOf -> Gen.list / Gen.array / Gen.seq

> This makes sense to me, since F# has quite a few (..!) collection types: `list`, `array`, `seq`, `set`, `map`, etc. Maybe we even need to consider adding a `Gen.set` one.

It's too bad `Gen.map` will cause a conflict, not sure what to do about that. We could append `Of` to all of them, but I actually really like `Gen.list` instead of `Gen.listOf`. The way one pipelines things from right to left in F#, having the `Of` on the wrong side is a bit weird:

```fs
Gen.range 0 100 |> Gen.list
//vs
Gen.range 0 100 |> Gen.listOf
```

moodmosaic:
>I guess `growingItem` wouldn't be so bad, `elements` isn't such a bad name in any case I suppose. I went for item as it was similar in function to `List.item` / `Array.item` / `Seq.item`.

👍 

>Do you have a use case for `growingElements` by chance? I have never used it myself so I can't imagine the scenario.

I don't. I've only seen it in both Haskell QuickCheck and Erlang QuickCheck.

>I named this one after `choice` in FParsec, it has similar behaviour in that it takes `seq<'m<'a>> -> 'm<'a>`.

>In some ways I think it could fit in with the built-in choice, as we could have combinators like `choice2 : Gen<'a> -> Gen<'b> -> Gen<Choice<'a, 'b>>`, etc:

I didn't think of FParsec, and I didn't also think about having a hypothetical `Gen.choice2`. Good points 👍 

>It's too bad `Gen.map` will cause a conflict, not sure what to do about that.

Probably just don't add support for generating immutable dictionaries? In favor of:

```f#
Gen.range 0 100 |> Gen.list
// vs
Gen.range 0 100 |> Gen.listOf
```

If a user wants to create an immutable dictionary he or she may use the existing combinators, I guess...

jystic:
> If a user wants to create an immutable dictionary he or she may use the existing combinators, I guess...

That's true, it's pretty easy:
```fs
Gen.zip Gen.string Gen.int |> Gen.list |> Map.ofList
```

moodmosaic:
👍

PR #21, opened 2016-10-13
Add the standard "hello-world" property in script.fsx,

moodmosaic:
because I think that's what most users look for when learning a new property-based testing tool.


jystic:
Cool 🍔

PR #22, opened 2016-10-13
Conditional and Lazy Properties

moodmosaic:
Implemented as in QuickCheck 1.2.0.1, and as described in the [manual](http://www.cse.chalmers.se/~rjmh/QuickCheck/manual.html
). The important thing here is that even though properties may take the form

```
<condition> ==> <property>
```

F# has strict semantics, and so the use of a lazy computation can be required in cases we shouldn't evaluate the property at the right of the condition.

For example, the following property isn't checked correctly without using a `lazy` computation:

```f#
fun n -> n <> 0 ==> (1 / n = 1 / n)
|> Property.forAll Gen.int
|> Property.check

System.DivideByZeroException: Attempted to divide by zero.
>    at FSI_0045.it@92-64.Invoke(Int32 n) in ..
Stopped due to error
```

Instead, we need to tell Jack to *not* evaluate the computation immediately and instead evaluate it only when the result is needed:

```f#
fun n -> n <> 0 ==> lazy (1 / n = 1 / n)
|> Property.forAll Gen.int
|> Property.check

+++ OK, passed 100 tests.
val it : bool = true
>
```

### DIFF
--- a/Jack/Property.fs
+++ b/Jack/Property.fs
@@ -43,6 +43,13 @@ module Property =
         else
             Failure [] |> ofResult
 
+    let private convert (value : 'T) =
+        match box value with
+        | :? Lazy<bool> as b -> ofBool b.Value
+        | :? Property   as p -> p
+        | :? bool       as b -> ofBool b
+        | _                  -> ofBool true
+
     let mapGen (f : Gen<Result> -> Gen<Result>) (Property x : Property) : Property =
         Property <| f x
 
@@ -54,6 +61,15 @@ module Property =
             counterexample (sprintf "%A" x) (f x) |> toGen
         Gen.bind gen prepend |> ofGen
 
+    //
+    // Conditional Properties
+    //
+
+    let implies (x : bool) (candidate : 'TProperty) =
+        if x then
+            convert candidate
+        else
+            convert ()
 
     //
     // Runner
@@ -132,3 +148,7 @@ module ForAllBuilder =
             Property.forAll m k
 
     let forAll = Builder ()
+
+[<AutoOpen>]
+module PropertyOperators =
+    let (==>) = Property.implies
--- a/Jack/Script.fsx
+++ b/Jack/Script.fsx
@@ -78,6 +78,22 @@ Property.check <| forAll {
 }
 
 //
+// Conditional Properties
+//
+
+fun (x, y) -> (x > 0 && y > 0) ==> (x * y > 0)
+|> Property.forAll (Gen.zip Gen.int Gen.int)
+|> Property.check
+
+//
+// Lazy Properties
+//
+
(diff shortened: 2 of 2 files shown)


Jack/Script.fsx
```diff
@@ -78,6 +78,22 @@ Property.check <| forAll {
 }
 
 //
+// Conditional Properties
+//
+
+fun (x, y) -> (x > 0 && y > 0) ==> (x * y > 0)
+|> Property.forAll (Gen.zip Gen.int Gen.int)
+|> Property.check
```
moodmosaic: When I execute this in F# Interactive, I'm getting back:

```
*** Failed! Falsifiable (after 45 tests and 29 shrinks):
(879, 2443099)
val it : bool = false
> 
```

Most likely I did something wrong, otherwise it can be a bug (?) :confused:

jystic: I think you have discovered the limits of 32-bit integers 😃 

```
2147483647 = 2^31 - 1 (max 32-bit number)
2147484021 = 879 * 2443099
```

Only `374` higher than the actual maximum 32-bit number, so the shrinking works well 👍

Jack/Property.fs
```diff
@@ -43,6 +43,13 @@ module Property =
         else
             Failure [] |> ofResult
 
+    let private convert (value : 'T) =
+        match box value with
+        | :? Lazy<bool> as b -> ofBool b.Value
+        | :? Property   as p -> p
+        | :? bool       as b -> ofBool b
+        | _                  -> ofBool true
```
jystic: Sorry, but I don't like these dynamic checks, if we implement an implication operator then it has to be using the FsControl style ad-hoc polymorphism. The fact that everything still type checks even if I pass an `int` to `convert` is a problem for me.

I'm not sure that this is necessarily required though, I have something in mind for implication which uses computation expressions and should work for both your examples, I'll put up a PR and see what you think?

jystic: I actually was at the ML Workshop to see that work being presented, it's very nice!

moodmosaic: Yes, it'll make things easier when it gets released 😃

jystic: I suspect that as always, the main hurdle is not the engineering work, but convincing the F# team at Microsoft that's it's a good fit. Apparently the best thing we can do to help is to get lots of people to say that they really want this feature.

moodmosaic: It looks like [some work was done](https://github.com/Microsoft/visualfsharp/compare/master...CaptainHayashi:hackathon-vs) on this... There's also [this](https://fslang.uservoice.com/forums/245727-f-language/suggestions/5762135-support-for-type-classes-or-implicits) UserVoice entry. Let's see what happens.

PR #23, opened 2016-10-13
Support for conditional generators and conditional properties

jystic:
By implementing `Delay` and `Zero` on the `forAll` builder, it means that if statements work as expected:

```fs
Property.check <| forAll {
    let! n = Gen.int
    if n <> 0 then
        return 1 / n = 1 / n
}
// +++ OK, passed 100 tests.
```

### DIFF
--- a/Jack/Gen.fs
+++ b/Jack/Gen.fs
@@ -67,6 +67,36 @@ module Gen =
     let zip3 (gx : Gen<'a>) (gy : Gen<'b>) (gz : Gen<'c>) : Gen<'a * 'b * 'c> =
         map3 (fun x y z -> x, y, z) gx gy gz
 
+    /// More or less the same logic as suchThatMaybe from QuickCheck, except
+    /// modified to ensure that the shrinks also obey the predicate.
+    let private tryRandom (p : 'a -> bool) (r0 : Random<Tree<'a>>) : Random<Option<Tree<'a>>> =
+        let rec tryN k = function
+            | 0 ->
+                Random.constant None
+            | n ->
+                let r = Random.resize (2 * k + n) r0
+                Random.bind r <| fun x ->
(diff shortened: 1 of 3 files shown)


jystic:
In the "hidden" else case, `Zero` is called. This discards the current attempt and we repeat the test with new values.

moodmosaic:
Awesome 👍

Jack/Gen.fs
```diff
@@ -67,6 +67,36 @@ module Gen =
...
+                Random.bind r <| fun x ->
+                    if p (Tree.outcome x) then
+                        Tree.filter p x |> Some |> Random.constant
+                    else
+                        tryN (k + 1) (n - 1)
+
+        Random.sized (tryN 0 << max 1)
+
+    /// Generates a value that satisfies a predicate.
+    let suchThat (p : 'a -> bool) (g : Gen<'a>) : Gen<'a> =
```
moodmosaic: How about `where` or `filter`?

jystic: Oh yeah, hah I didn't even think about this, it's a straight port from Haskell.

`filter` seems the most F#-like, I'll change it

jystic: > And perhaps even return a `Gen<'a option>` instead?

In the Haskell version there's combinators for both cases:

```hs
suchThat :: Gen a -> (a -> Bool) -> Gen a
suchThatMaybe :: Gen a -> (a -> Bool) -> Gen (Maybe a)
```

so what about:

```fsharp
val filter : ('a -> bool) -> Gen<'a> -> Gen<'a>
val tryFilter : ('a -> bool) -> Gen<'a> -> Gen<'a option>
```

moodmosaic: I don't see why not 👍

Issue #25, opened 2016-10-14
Observing Test Case Distribution


moodmosaic:
Taken from QuickCheck [manual](http://www.cse.chalmers.se/~rjmh/QuickCheck/manual_body.html#8):

>It is important to be aware of the distribution of test cases: if test data is not well distributed then conclusions drawn from the test results may be invalid.

Thus, we could consider adding [`label`](https://hackage.haskell.org/package/QuickCheck-2.9.2/docs/Test-QuickCheck-Property.html#v:label), [`classify`](https://hackage.haskell.org/package/QuickCheck-2.9.2/docs/Test-QuickCheck-Property.html#v:classify), and [`collect`](https://hackage.haskell.org/package/QuickCheck-2.9.2/docs/Test-QuickCheck-Property.html#v:collect).

jystic:
I have never used this kind of thing myself, but I can see the appeal, I usually test generators using `printSample` and eye-balling the result, perhaps this is better though

moodmosaic:
https://github.com/hedgehogqa/haskell-hedgehog/issues/127#issuecomment-341927624

cmeeren:
I think this would be a great addition to Hedgehog for the obvious reasons mentioned by @moodmosaic. I miss them (coming from FsCheck) and would certainly use them if there were present.

A related suggestion is to enable tests to fail if the number of tests in certain classifications are above or below a given percentage, or absolute number, or both (and perhaps even to generate more test cases until a sufficient number is met). As the feature is implemented in FsCheck, you have to manually check the output of each test to make sure the percentages are OK.

Alxandr:
I would also really love this feature 👍. I've been looking at the at the code (and the haskell package https://github.com/qfpl/tasty-hedgehog-coverage), but as far as I understood what they do is wrap the whole `Prop` in an outer monad that does some funky stuff that went way over my head (it's too late at night for me to figure out nesting of monads at this point). My current thinking is that is should probably be part of the journal (just like the counter examples), the issue with that is dealing with cases where you have 0 "hits". The registry of possible cases needs to be stored somewhere. Any thoughts?

moodmosaic:
[...]

PR #29, opened 2016-10-15
Support for 'where' and 'counterexample' custom operations

jystic:
I also renamed the `forAll` builder to `property` as I think it's more correct. I think this name is better, not completely sure.

/cc @moodmosaic

### DIFF
--- a/Jack/Property.fs
+++ b/Jack/Property.fs
@@ -2,83 +2,153 @@
 
 open FSharpx.Collections
 
-type Result =
-    | Failure of List<string>
-    | Success
-    | Discard
+type Report =
+    | Report of List<string>
 
-type Property =
-    | Property of Gen<Result>
+type Result<'a> =
+    | Failure
+    | Discard
+    | Success of 'a
+
+type Property<'a> =
+    | Property of Gen<Report * Result<'a>>
+
+module Tuple =
+    let first (f : 'a -> 'c) (x : 'a, y : 'b) : 'c * 'b =
+        f x, y
+
+    let second (f : 'b -> 'c) (x : 'a, y : 'b) : 'a * 'c =
+        x, f y
 
 [<CompilationRepresentation(CompilationRepresentationFlags.ModuleSuffix)>]
 module Result =
-    let isFailure (x : Result) : bool =
-        match x with
-        | Failure _ ->
-            true
-        | Success ->
-            false
-        | Discard ->
-            false
-
-    let mapFailure (f : List<string> -> List<string>) (x : Result) : Result =
-        match x with
-        | Failure msgs ->
-            Failure (f msgs)
-        | Success ->
-            Success
+    let map (f : 'a -> 'b) (r : Result<'a>) : Result<'b> =
+        match r with
+        | Failure ->
+            Failure
         | Discard ->
             Discard
+        | Success x ->
+            Success (f x)
+
+    let filter (f : 'a -> bool) (r : Result<'a>) : Result<'a> =
+        match r with
+        | Failure ->
+            Failure
+        | Discard ->
+            Discard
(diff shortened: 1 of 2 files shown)


jystic:
An example of the new operations in action:
```fs
Property.check <| property {
    let! x = Gen.int
    let! y = Gen.int
    where (x > 0 && y > 0)
    counterexample (sprintf "x * y = %d" <| x * y)
    return x * y > 0
}
```

Jack/Property.fs
```diff
@@ -2,83 +2,153 @@
...
+type Report =
+    | Report of List<string>
+
+type Result<'a> =
+    | Failure
     | Discard
+    | Success of 'a
+
+type Property<'a> =
+    | Property of Gen<Report * Result<'a>>
```
moodmosaic: `Report` is there to support #25, right?

Jack/Property.fs
```diff
@@ -2,83 +2,153 @@
...
+
+module Tuple =
```
moodmosaic: How about marking `Tuple` as `private`? I don't think we care about exposing it to the users. :wink:

Jack/Property.fs
```diff
@@ -2,83 +2,153 @@
...
+
+    let addFailure (msg : string) (x : Report) : Report =
+        map (List.cons msg) x
+
+    let append (xs : Report) (ys : Report) : Report =
+        toList xs @ toList ys |> ofList
 
 [<CompilationRepresentation(CompilationRepresentationFlags.ModuleSuffix)>]
 module Property =
+    open Tuple
```
moodmosaic: If we mark `Tuple` as `private`, we may then `[<AutoOpen>]` it, and then we can get rid of that `open` directive.

Jack/Property.fs
```diff
@@ -133,33 +203,47 @@ module Property =
...
+            List.map (printfn "%s") (Report.toList report) |> ignore
             false
 
-    let check (p : Property) : bool =
+    let check (p : Property<unit>) : bool =
         check' 100 p
 
 [<AutoOpen>]
-module ForAllBuilder =
+module PropertyBuilder =
```
moodmosaic: Yes! I find `Property` a better name, since we may not only support universal quantification in the future. 😉 

One day, if we decide to support existentials (à la [SmallCheck](https://github.com/feuerbach/smallcheck)), we may still do that through some custom operation(s), and then we won't have to rename the builder itself – which may be a breaking change if we follow [SemVer](http://semver.org/).

moodmosaic:
It's also great that we don't have to use a `lazy` computation in this case! It seems to be just working out of the box.

```f#
Property.check <| property {
    let! n = Gen.int
    where (n <> 0)
    return 1 / n = 1 / n
}

//+++ OK, passed 100 tests.
//val it : bool = true
```

moodmosaic:
👍 This is looking good!

* `where` and `counterexample` are impressive
* [`property` looks good](https://github.com/jystic/dotnet-jack/pull/29#discussion_r83532584)
* I've opened #30, based on the same (`topic/zero-not-ideal`) branch

Jack/Property.fs
jystic: Sounds good, will check out your PR

Jack/Property.fs
jystic: It was just for failure messages, but it could be used to support #25 or any other types of things that we might want to track during a test run.

jystic:
> It's also great that we don't have to use a `lazy` computation in this case! It seems to be just working out of the box.

The key here is that `PropertyBuilder` implements `Delay`, so every line of the computation expression gets wrapped in `Gen.delay` and hence is lazy.

PR #30, opened 2016-10-15
Mark Tuple as private and [<AutoOpen>],

moodmosaic:
in order to make its let-bound values available when users open its namespace.


jystic:
Rebased and merged 🎱

PR #31, opened 2016-10-16
Do we care about making Shrink.double more generic?

moodmosaic:
If so, this pull request turns `Shrink.double` into a generic `Shrink.number` function:

```f#
val number : x:'a -> LazyList<'a>
	(requires member get_Zero
		  and member ( ~- )
		  and member ( - )
		  and member ( / )
		  and member ( + )
		  and member get_One
		  and comparison)

// > Shrink.number 10.m;;
// val it : FSharpx.Collections.LazyList<decimal> =
//   seq [0M; 5M; 7.5M; 8.75M; ...]

// > Shrink.number 10.0;;
// val it : FSharpx.Collections.LazyList<float> =
//	 seq [0.0; 5.0; 7.5; 8.75; ...]

// > Shrink.number 10s;;
// val it : FSharpx.Collections.LazyList<int16> =
//   seq [0s; 5s; 8s; 9s]

// > Shrink.number 10L;;
// val it : FSharpx.Collections.LazyList<int64> = 
//   seq [0L; 5L; 8L; 9L]

// > Shrink.number 10;;
// val it : FSharpx.Collections.LazyList<int> = 
//   seq [0; 5; 8; 9]
```

### DIFF
--- a/Jack/Shrink.fs
+++ b/Jack/Shrink.fs
@@ -108,16 +108,15 @@ module Shrink =
             LazyList.consNub destination <|
             LazyList.map (fun y -> x - y) (halves diff)
 
-    /// Shrink a floating point number.
-    let double (x : double) : LazyList<double> =
+    /// Shrink a number.
+    let inline number x : LazyList<'a> =
         let positive =
-            if x < 0.0 then
+            if x < LanguagePrimitives.GenericZero then
                 LazyList.singleton (-x)
             else
                 LazyList.empty
 
         let integrals =
-            towards 0I (bigint x)
-            |> LazyList.map double
+            towards LanguagePrimitives.GenericZero x
 
         LazyList.append positive integrals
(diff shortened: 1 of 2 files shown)


Jack/Shrink.fs
```diff
@@ -108,16 +108,15 @@ module Shrink =
...
-            if x < 0.0 then
+            if x < LanguagePrimitives.GenericZero then
                 LazyList.singleton (-x)
             else
                 LazyList.empty
 
         let integrals =
-            towards 0I (bigint x)
-            |> LazyList.map double
+            towards LanguagePrimitives.GenericZero x
```
jystic: So, the old code was deliberately forcing the value to an integer (which is what the current version of QuickCheck does in Haskell now), because I think floating point math can put us in a situation where shrinking doesn't terminate.

jystic:
Sorry for the delayed reply on this, I've been busy the last few days

moodmosaic:
No worries!

PR #32, opened 2016-10-19
Add example-based tests for Shrink.elems and Shrink.double

moodmosaic:


jystic:
Nice 🐎 rebased and merged

Issue #34, opened 2016-10-20
Property.check' could throw a System.Exception,


moodmosaic:
instead of returning `bool` and printing to stdout, because otherwise xUnit.net (and I believe also NUnit and MSTest) can't figure out whether a property fails or not:

```f#
[<Fact>]
let ``the reverse of the reverse of a list is the list itself`` () =
    Property.check <| property {
        let! xs = Gen.list Gen.int
        return List.rev xs = xs // Fails, on purpose.
    }
```
Output:

```
------ Test started: Assembly: Jack.Tests.dll ------

1 passed, 0 failed, 0 skipped, took 0.69 seconds (xUnit.net 2.1.0 build 3179).
```

Expected output:

```
------ Test started: Assembly: Jack.Tests.dll ------

Test 'Jack.Tests.ShrinkTests.the reverse of the reverse of a list is the list itself' 
failed: System.Exception : *** Failed! Falsifiable (after 5 tests and 3 shrinks):
[1; 0]

	at Microsoft.FSharp.Core.Operators.FailWith[T](String message)
	..\Property.fs(215,0): at Jack.PropertyModule.check'(Int32 n, Property`1 p)
	..\Property.fs(218,0): at Jack.PropertyModule.check(Property`1 p)
	..\PropertyTests.fs(126,0):
		at Jack.Tests.ShrinkTests.the reverse of the reverse of a list is the list itself()

0 passed, 1 failed, 0 skipped, took 2.86 seconds (xUnit.net 2.1.0 build 3179).
```

That, or we could add another function in the `Property` module that throws a System.Exception instead. – Let me know and I'd happily send a pull request :smiley:

jystic:
This is fixed by https://github.com/jystic/dotnet-jack/pull/35

moodmosaic:
Awesome! I'll check #35 in the next couple of hours, hopefully.

PR #35, opened 2016-10-22
Property.check now throws exceptions for better test runner integration

jystic:
`Property.print`, and the lower level `Property.report`, are available for experimentation in scripts.

This PR has a number of changes:
- `Property.check` becomes `Property.report` and returns a rich data type instead of a `bool`
- `Property.print` gives more or less the same behaviour as the previous `Property.check` although it doesn't return a `bool`, maybe it should
- `Property.check` throws the new exceptions `GaveUpException` and `FailedException` where appropriate, this allows for better integration with test runners, as per issue #34 
- Added `MinimalTests.fs` to check that shrinking gives a minimal counterexample and to check that the test runner integration works well
- Added some more operations to the `property` builder, you can now loop over sequences, do arbitrary effects, and `try/catch` inside properties. I have no examples of using this yet but I needed it for some model testing stuff I was working on.

*edit: I also removed the dependency on FSharpPlus as we were only using FsControl, it might be nice if we can take the bits we need (to/from bigint) from FsControl so we don't need to depend on it either*

### DIFF
--- a/Jack/Property.fs
+++ b/Jack/Property.fs
@@ -2,8 +2,8 @@
 
 open FSharpx.Collections
 
-type Report =
-    | Report of List<string>
+type Journal =
+    | Journal of List<string>
 
 type Result<'a> =
     | Failure
@@ -11,7 +11,22 @@ type Result<'a> =
     | Success of 'a
 
 type Property<'a> =
-    | Property of Gen<Report * Result<'a>>
+    | Property of Gen<Journal * Result<'a>>
+
+[<Measure>] type tests
+[<Measure>] type discards
+[<Measure>] type shrinks
+
+type Status =
+    | Failed of int<shrinks> * Journal
+    | GaveUp
+    | OK
+
+type Report = {
+    tests : int<tests>
+    discards : int<discards>
+    status : Status
+}
 
 [<AutoOpen>]
 module private Tuple =
@@ -20,6 +35,26 @@ module private Tuple =
 
     let second (f : 'b -> 'c) (x : 'a, y : 'b) : 'a * 'c =
         x, f y
+
+[<CompilationRepresentation(CompilationRepresentationFlags.ModuleSuffix)>]
+module Journal =
+    let ofList (xs : List<string>) : Journal =
+        Journal xs
+
+    let toList (Journal xs : Journal) : List<string> =
+        xs
+
+    let empty : Journal =
+        List.empty |> ofList
+
+    let map (f : List<string> -> List<string>) (xs : Journal) : Journal =
+        toList xs |> f |> ofList
+
+    let addFailure (msg : string) (x : Journal) : Journal =
+        map (List.cons msg) x
+
+    let append (xs : Journal) (ys : Journal) : Journal =
+        toList xs @ toList ys |> ofList
 
 [<CompilationRepresentation(CompilationRepresentationFlags.ModuleSuffix)>]
 module Result =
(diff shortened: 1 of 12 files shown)


Jack/Property.fs
```diff
@@ -221,28 +343,44 @@ module PropertyBuilder =
...
+        member __.ReturnFrom(m : Property<'a>) : Property<'a> =
+            m
+
         member __.Delay(f : unit -> Property<'a>) : Property<'a> =
-            Gen.delay (Property.toGen << f) |> Property.ofGen
+            Property.delay f
 
         member __.Zero() : Property<unit> =
-            Property.discard
+            Property.success ()
```
jystic: Turns out this has to be `success` instead of `discard` otherwise doing effects causes the property to be discarded.

An example of something you can do with this is:
```fs
Property.check <| property {
    // setup
    for n in names do
        try Registry.unregister n with _ -> ()

    // do actual property stuff..
}
```

Jack/Property.fs
```diff
@@ -221,28 +343,44 @@ module PropertyBuilder =
         member __.For(m : Property<'a>, k : 'a -> Property<'b>) : Property<'b> =
             Property.bind m k
 
+        member __.For(xs : seq<'a>, k : 'a -> Property<unit>) : Property<unit> =
+            let join m n =
+                Property.bind m (fun _ -> n)
+            let init =
+                Property.success ()
+            xs |> Seq.map k |> Seq.fold join init
```
jystic: I think this isn't safe yet as we don't dispose the `IEnumerator` unless the whole list is looped over, will fix in a later PR though

Jack/Property.fs
```diff
@@ -2,16 +2,31 @@
...
     | Discard
     | Success of 'a
 
 type Property<'a> =
-    | Property of Gen<Report * Result<'a>>
+    | Property of Gen<Journal * Result<'a>>
+
+[<Measure>] type tests
+[<Measure>] type discards
+[<Measure>] type shrinks
```
moodmosaic: Good idea! Units of Measure will make these much more clear now.

Jack/Property.fs
moodmosaic: I *think* you're right.

Jack/Property.fs
jystic: yeah that's right, effects would be of type `Property<unit>` for example:
```fs
property { printfn "foo" } : Property<unit>
```

moodmosaic:
>`Property.print`, and the lower level `Property.report`, are available for experimentation in scripts.

👍 

>`Property.check` becomes `Property.report` and returns a rich data type instead of a `bool`

Awesome. This is much better now, more flexible, and easier to reason about. 🎉 

>`Property.print` gives more or less the same behaviour as the previous `Property.check` although it doesn't return a `bool`, maybe it should

No, I think it should not. 

>`Property.check` throws the new exceptions `GaveUpException` and `FailedException` where appropriate, this allows for better integration with test runners, as per issue #34

👍 🎉 

>I also removed the dependency on FSharpPlus as we were only using FsControl, it might be nice if we can take the bits we need (to/from bigint) from FsControl so we don't need to depend on it either

👍 It'd be nice if we don't have *any* external dependency. This way we can be much more flexible. Thus, I'd open a PR (if you want) with the `to`/`from` `bigint` bits we need from FsControl, so that we don't have to reference it.

---

>Added `MinimalTests.fs` to check that shrinking gives a minimal counterexample and to check that the test runner integration works well

>Added some more operations to the `property` builder, you can now loop over sequences, do arbitrary effects, and `try/catch` inside properties. I have no examples of using this yet but I needed it for some model testing stuff I was working on.

I haven't tried these out yet. Once I'll do (soon) I might be able to provide some feedback.

---

I've opened #36 while looking at this pull request.

Jack/Property.fs
jystic: I've spent a bunch of time today on loops today, and I have something workable, but it can still blow the stack, we might need to build a trampoline in to Property/Gen/Random, not sure.

jystic:
> 👍 It'd be nice if we don't have *any* external dependency. This way we can be much more flexible. Thus, I'd open a PR (if you want) with the `to`/`from` `bigint` bits we need from FsControl, so that we don't have to reference it.

I agree, if you'd like to do this that would be great, maybe put it in `Numeric.fs`?

moodmosaic:
>I agree, if you'd like to do this that would be great, maybe put it in `Numeric.fs`?

Sounds like a plan.

Jack/Property.fs
moodmosaic: Not sure either without first taking a closer look.

moodmosaic:
>I agree, if you'd like to do this that would be great, maybe put it in `Numeric.fs`?
>>Sounds like a plan.

Done now in #38.

PR #37, opened 2016-10-22
Loops now work somewhat properly

jystic:
Unfortunately they still grow the stack each iteration, this might be fixable with a trampoline, but I was surprised how far we can get even with this implementation.

### DIFF
--- a/Jack/Property.fs
+++ b/Jack/Property.fs
@@ -1,6 +1,7 @@
 ﻿namespace Jack
 
 open FSharpx.Collections
+open System
 
 type Journal =
     | Journal of List<string>
@@ -46,6 +47,9 @@ module Journal =
 
     let empty : Journal =
         List.empty |> ofList
+
+    let singleton (x : string) : Journal =
+        List.singleton x |> ofList
 
     let map (f : List<string> -> List<string>) (xs : Journal) : Journal =
         toList xs |> f |> ofList
@@ -150,7 +154,7 @@ module private Pretty =
 
 [<AbstractClass>]
 type JackException (message : string) =
-    inherit System.Exception (message)
+    inherit Exception (message)
 
 type GaveUpException (tests : int<tests>, discards : int<discards>) =
     inherit JackException (renderGaveUp tests discards)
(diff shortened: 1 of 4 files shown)


Jack/Property.fs
```diff
@@ -229,6 +247,9 @@ module Property =
         else
             failure
 
+    let counterexample (msg : string) : Property<unit> =
+        Gen.constant (Journal.singleton msg, Success ()) |> ofGen
```
jystic: I'm not 100% on this, but I changed `counterexample` to simply be a `Property<unit>` rather than a function `Property<'a> -> Property<'a>` because it makes calling it inside a computation expression possible. You would think this doesn't matter because we have the custom operation `counterexample`, but for some reason custom operations can't be used at the same time as control flow operations like `for` and `while`.

Jack/Script.fsx
```diff
@@ -112,6 +112,30 @@ Property.print <| property {
...
+        return x <> 'w'
+        return x <> 'z'
+
+    let mutable n = 0
+    while n < 10 do
+        n <- n + 1
+        let! k = Gen.range 0 n
+        return! Property.counterexample (sprintf "n = %d" n)
+        return! Property.counterexample (sprintf "k = %d" k)
+        return k <> 5
```
jystic: This shows the crazy stuff which you can do with loops, I'm not sure if this is actually useful yet, we'll see.

It boggles the mind to think what is happening under the hood with the rose tree.

moodmosaic:
Hm... I'm curious to look at the generated IL when compiling this with [`--tailcalls`](https://docs.microsoft.com/en-us/dotnet/articles/fsharp/language-reference/compiler-options) on, or when compiling in Release mode (e.g. by running `./build.sh` and loading the .DLLs in [ILSpy](http://ilspy.net/) or similar).

Jack/Property.fs
```diff
@@ -337,27 +355,43 @@ module Property =
...
-            let join m n =
-                Property.bind m (fun _ -> n)
-            let init =
-                Property.success ()
-            xs |> Seq.map k |> Seq.fold join init
+            let xse = xs.GetEnumerator ()
+            Property.using xse <| fun xse ->
+                let mv = xse.MoveNext
+                let kc = Property.delay (fun () -> k xse.Current)
+                loop mv kc
```
jystic: Yeah I was thinking maybe `Random` should be `Seed -> Size -> Trampoline<'a>`, will do some experimentation when I get a chance

Jack/Script.fsx
```diff
@@ -112,6 +112,30 @@ Property.print <| property {
...
+Property.print <| property {
+    for x in "abcd" do
+        // Custom operations (i.e. counterexample) can't be used in computation
+        // expressions which have control flow :( we can fake it using return!
+        // however.
+        return! Property.counterexample (sprintf "x = %A" x)
+
+        // Note, return can be used multiple times, its a bit like 'assert'.
+        return x <> 'w'
+        return x <> 'z'
```
jystic: @moodmosaic I'm not sure that I like this, I have some ideas to make `return` one shot only which might be better

moodmosaic: @jystic I tend to not like this either, but haven't yet spend enough time on this, so I don't have some useful feedback for the time being.

PR #38, opened 2016-10-24
Removed the dependency on FsControl,

moodmosaic:
as discussed in #35.


moodmosaic:
090faff broke the build, since I forgot to update `Script.fsx`, so 828018b fixes it.

jystic:
Fantastic 🎉

PR #39, opened 2016-10-24
Support for generating classes with overloaded constructors

moodmosaic:


Jack/Gen.fs
```diff
@@ -67,12 +67,26 @@ module Gen =
...
 
     let zip3 (gx : Gen<'a>) (gy : Gen<'b>) (gz : Gen<'c>) : Gen<'a * 'b * 'c> =
         map3 (fun x y z -> x, y, z) gx gy gz
 
+    let zip4 (gx : Gen<'a>) (gy : Gen<'b>) (gz : Gen<'c>) (gw : Gen<'d>) : Gen<'a * 'b * 'c * 'd> =
+        map4 (fun x y z w -> x, y, z, w) gx gy gz gw
+
+    let two   (g : Gen<'a>) : Gen<'a * 'a>           = zip  g g
+    let three (g : Gen<'a>) : Gen<'a * 'a * 'a>      = zip3 g g g
+    let four  (g : Gen<'a>) : Gen<'a * 'a * 'a * 'a> = zip4 g g g g
```
jystic: These names are a bit inconsistent with the other n-arity functions we have, I'd prefer something with a numeric suffix. Maybe `tuple`, `tuple3` .. `tupleN` or `dup`, `dup3` .. `dupN`, open to suggestions though.

Also a bit of a style nitpick, I'm not a fan of aligning things, as it creates noise in PRs. So, I'd prefer to adopt a style where alignment isn't required for readability. To this end, I think top level functions should have their body wrapped like:
```fs
let two (g : Gen<'a>) : Gen<'a * 'a> =
    zip g g
```
Hope that's ok :)

moodmosaic: I've just added 7cddf8d and 9871179 to fix both the names and the alignments. Thanks for the review! 👍

jystic:
Perfect 🍰

PR #40, opened 2016-10-25
Remove aligning from the ad-hoc polymorphic things we got from FsControl,

moodmosaic:
related to what we're discussing [here](https://github.com/jystic/dotnet-jack/pull/39#discussion_r84802315).


jystic:
Nice, yeah this is much better 👾

jystic:
Rebased and merged

Issue #26, opened 2016-10-14
SplitMix or TFGen as a better fast splittable random generator


moodmosaic:
Perhaps, dotnet-jack could benefit from using [SplitMix](https://www.researchgate.net/profile/Doug_Lea/publication/273188325_Fast_Splittable_Pseudorandom_Number_Generators/links/57bcd2e008ae37ee394a6f92.pdf) (or even [TFGen](http://publications.lib.chalmers.se/records/fulltext/183348/local_183348.pdf)<sup>1</sup>?), since the current splittable random number generator seems to have known issues, as shown [here](https://www.reddit.com/r/haskell/comments/3x15sm/why_is_the_first_random_value_produced_from_a/), [here](https://github.com/haskell/random/issues/31), and [here](https://ghc.haskell.org/trac/summer-of-code/ticket/1674).

---

<sup>1</sup> In fact, [QuickCheck 2 provides a wrapper around either TFGen or StdGen](https://hackage.haskell.org/package/QuickCheck-2.9.2/docs/src/Test-QuickCheck-Random.html#QCGen).

jystic:
I actually started porting [tf-random](https://hackage.haskell.org/package/tf-random) when I first started on this but I it seemed like it was a pretty big task in and of itself.

Would definitely like to move to something like TFGen in the long run though!

moodmosaic:
I think I'll give SplitMix a shot, sooner or later, by porting it from the [original paper](https://www.researchgate.net/profile/Doug_Lea/publication/273188325_Fast_Splittable_Pseudorandom_Number_Generators/links/57bcd2e008ae37ee394a6f92.pdf), or maybe from [here](https://github.com/nkartashov/SplitMix).

jystic:
Whether it's SplitMix or TFGen, I am keen!

I'm going to be focusing on porting the rest of the functionality I have in haskell/purescript and making sure it works nicely in F#, so I'd be happy to take a PR whether you want to do it in a separate library or in this one, up to you.

moodmosaic:
It'll be nice having it in this one, if the implementation is reasonable small. One less external dependency 😃

jystic:
I'm actually surprised the core of that Haskell SplitMix implementation is in C, looking at the code it seems to be things that should run fast enough in Haskell.

[...]

PR #41, opened 2016-10-28
Make functions in Seed module private,

moodmosaic:
if they aren't used outside of the module.

---

Basically, I came across this while working on #26 (for which I _do_ have something locally that seems to work, _yay!_) 😃

### DIFF
--- a/Jack/Seed.fs
+++ b/Jack/Seed.fs
@@ -26,7 +26,7 @@ module Seed =
             failwith msg
 
     /// Create a new 'Seed' from a 32-bit integer.
-    let ofInt32 (s0 : int) : Seed =
+    let private ofInt32 (s0 : int) : Seed =
         // We want a non-negative number, but we can't just take the abs
         // of s0 as -Int32.MinValue == Int32.MinValue.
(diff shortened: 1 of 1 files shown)


jystic:
I'm not so sure about this, it seems perfectly reasonable that someone might want to use these lower level functions? what's your motivation? just API compatibility?

> for which I do have something locally that seems to work, yay!

Awesome!

moodmosaic:
>it seems perfectly reasonable that someone might want to use these lower level functions?

The more we expose, the more we'd have to maintain.

If these functions aren't used outside of the `Seed` module perhaps we could just make them private. After all, the interface of a Seed is

* the function that returns a new Seed
* one or more *nextXyz* functions that generate numbers
* the split function that returns a new, different, Seed

The rest are just implementation details in most cases...

moodmosaic:
After we reach `v1`, it'll be a breaking change, according to [SemVer](http://semver.org/), if we change (API) functions from public to private. But, right now, these proposed change could be considered 'safe'. 😉

moodmosaic:
Closing this because of #44. We can reopen it in the future, if necessary.

PR #44, opened 2016-11-03
Replace GHC's System.Random ported implementation with SplitMix

### LINKED ISSUE #26: SplitMix or TFGen as a better fast splittable random generator

moodmosaic:
Resolves #26 

---

**Edit**: These measurements are now outdated, since new commits are added to the pull request.

`./build.sh` times on `splitmix` branch:

```
=== TEST EXECUTION SUMMARY ===
   Jack.Tests  Total: 49, Errors: 0, Failed: 0, Skipped: 0, Time: 39.751s
Finished Target: Test

---------------------------------------------------------------------
Build Time Report
---------------------------------------------------------------------
Target     Duration
------     --------
Build      00:00:04.8662755
Test       00:00:40.8340264
Total:     00:00:45.7305498
Status:    Ok
---------------------------------------------------------------------
```

`./build.sh` times on `master` branch:

```
=== TEST EXECUTION SUMMARY ===
   Jack.Tests  Total: 49, Errors: 0, Failed: 0, Skipped: 0, Time: 48.736s
Finished Target: Test

---------------------------------------------------------------------
Build Time Report
---------------------------------------------------------------------
Target     Duration
------     --------
Build      00:00:05.1305200
Test       00:00:49.8303248
Total:     00:00:54.9957778
Status:    Ok
---------------------------------------------------------------------
```

### DIFF
--- a/Jack/Seed.fs
+++ b/Jack/Seed.fs
@@ -109,9 +109,17 @@ module Seed =
             nextBigInt hi lo seed
         else
             let rec loop hilo (v0, seed0) =
+                // hilo is the size of the exclusive range: 'hi - lo + 1'.
                 if hilo > bigint.Zero then
+                    // The range is a positive number – return a
+                    // number in range using modular arithmetic:
+                    // - http://codereview.stackexchange.com/a/29110/10541
+                    // - http://stackoverflow.com/a/1202706/467754
                     lo + v0 % hilo, seed0
                 else if v0 < lo || v0 >= hi then
+                    // The range is a negative number – this shouldn't
+                    // happen (unless there is an arithmetic overflow)
+                    // so in this case keep going.
                     loop hilo (next seed0)
                 else v0, seed0
             loop <| hi - lo + bigint.One <| next seed
--- a/Jack/Seed.fs
+++ b/Jack/Seed.fs
@@ -114,39 +114,39 @@ module Seed =
         { s0 with Value = s0.Value + s0.Gamma }
 
     /// Returns the next pseudo-random number in the sequence, and a new seed.
-    let next (s0 : Seed) : int * Seed =
+    let next (s0 : Seed) : int64 * Seed =
         let s = nextSeed s0
-        let n = mix32 s.Value
+        let n = mix64 s.Value
         n, nextSeed s
-    
+
     /// Generate a random bigint in the specified range.
     let rec nextBigInt (lo : bigint) (hi : bigint) (seed : Seed) : bigint * Seed =
         if lo > hi then
             nextBigInt hi lo seed
         else
-            let s1 = nextSeed seed
-            let mutable r = bigint (mix32 s1.Value)
-            if (lo < hi) then
-                let n = hi - lo
-                let m = n - (bigint 1)
-                if (n &&& m) = (bigint 0) then (r &&& m) + lo, s1
-                elif n > (bigint 0) then
-                    let mutable u = r >>> 1
-                    let mutable s = nextSeed s1
-                    while u + m - (r <- u % n
-                                   r)
-                          < bigint 0 do
-                        u <- bigint (mix32 s.Value >>> 1)
-                        r <- r + lo
-                        s <- nextSeed s
-                    r, s
-                else
-                    let mutable s = nextSeed s1
-                    while (r < lo || r >= hi) do
-                        r <- bigint (mix32 s.Value)
-                        s <- nextSeed s
-                    r, s
-            else r, s1
+            let m = hi - lo
+            let n = m + bigint.One
+            let r, seed = next seed
+            let mutable candidate' = bigint r
+            let mutable s' = seed
+            let mutable result = System.Nullable<_>()
+            if (n > bigint 0) then
+                while not result.HasValue do
+                    let pcand = bigint (int64 (uint64 candidate' >>> 1))
+                    let offset = pcand % n
+                    if pcand + m - offset < bigint 0 then
+                        let r, seed = next seed
+                        candidate' <- bigint r
+                        s' <- seed
+                    else result <- System.Nullable<_>(lo + offset)
+            else
+                while not result.HasValue do
+                    if candidate' < lo || candidate' >= hi then
+                        let r, seed = next seed
+                        candidate' <- bigint r
+                        s' <- seed
+                    else result <- System.Nullable<_> candidate'
+            result.Value, s'
 
     /// Splits a random number generator in to two.
     let split (s0 : Seed) : Seed * Seed =
--- a/Jack/Seed.fs
+++ b/Jack/Seed.fs
@@ -100,8 +100,8 @@ module Seed =
           Gamma = mixGamma s + goldenGamma }
 
     /// Returns the next pseudo-random number in the sequence, and a new seed.
-    let next (s : Seed) : bigint * Seed =
-        bigint (mix64 s.Value), nextSeed s
+    let next (s : Seed) : int64 * Seed =
+        mix64 s.Value, nextSeed s
 
     /// Generate a random bigint in the specified range.
     let rec nextBigInt (lo : bigint) (hi : bigint) (seed : Seed) : bigint * Seed =
@@ -109,6 +109,7 @@ module Seed =
             nextBigInt hi lo seed
         else
             let rec loop hilo (v0, seed0) =
+                let v0 = bigint (v0 : int64)
                 // hilo is the size of the exclusive range: 'hi - lo + 1'.
                 if hilo > bigint.Zero then
                     // The range is a positive number – return a
--- a/Jack/Seed.fs
+++ b/Jack/Seed.fs
@@ -101,10 +101,6 @@ module Seed =
         let s = System.DateTimeOffset.UtcNow.Ticks + 2L * goldenGamma
         { Value = mix64 s
           Gamma = mixGamma s + goldenGamma }
-
-    /// The possible range of values returned from 'next'.
-    let range : int * int =
-        1, 2147483562
     
     /// Mix the bits of a 64-bit arg to produce a result, computing a
     /// bijective function on 64-bit values.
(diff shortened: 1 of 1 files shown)


Jack/Seed.fs
```diff
@@ -32,87 +58,69 @@ module Seed =
...
+        s ^^^ (s >>> 31)
+
+    let private bitCount (s0 : int64) : int =
+        let s = s0 - ((s0 >>> 1) &&& 0x5555555555555555L)
+        let s = (s &&& 0x3333333333333333L) + ((s >>> 2) &&& 0x3333333333333333L)
+        let s = (s + (s >>> 4)) &&& 0x0f0f0f0f0f0f0f0fL
+        let s = s + (s >>> 8)
+        let s = s + (s >>> 16)
+        let s = s + (s >>> 32)
+        (int s) &&& 0x7f
```
jystic: It's a shame there is no .NET intrinsic for `POPCNT`, this should be a single instruction with SSE4

Jack/Seed.fs
```diff
@@ -32,87 +58,69 @@ module Seed =
...
-            crashUnless (v >= 0I) "v >= 0I"
-            crashUnless (k >= 0I) "k >= 0I"
-            lo + v % k, seedN
+            let rec loop hilo (v0, seed0) =
+                if hilo > bigint.Zero then
+                    lo + v0 % hilo, seed0
+                else if v0 < lo || v0 >= hi then
+                    loop hilo (next seed0)
+                else v0, seed0
+            loop <| hi - lo + bigint.One <| next seed
```
jystic: Can you explain how this works? I don't quite follow how it generates a number between the entire range `lo` to `hi`

Jack/Seed.fs
```diff
@@ -32,87 +58,69 @@ module Seed =
...
-    
-        let z    = s1'' - s2''
-        let z'   = if z < 1 then z + 2147483562 else z
-    
-        z', Seed (s1'', s2'')
-    
+    let next (s0 : Seed) : bigint * Seed =
+        let s = nextSeed s0
+        let n = mix64 s.Value
+        bigint n, nextSeed s
```
jystic: Is this supposed to cycle through the seed twice? Also it should really return only an `int64` as that is the range of the random value which you have.

I would have expected something like:
```fs
    let next (s : Seed) : int64 * Seed =
        mix64 s.Value, nextSeed s
```

jystic:
This is pure awesome, I just ha a few comments on things I'm not sure about

Jack/Seed.fs
moodmosaic: Yeah. There's [this](https://github.com/dotnet/corefx/issues/2209) issue but I'm not sure if it'll make it outside of [corefx](https://github.com/dotnet/corefx)...

Jack/Seed.fs
moodmosaic: >Is this supposed to cycle through the seed twice?

Oh, you're right – it shouldn't have to. I'll change it.

moodmosaic: >Also it should really return only an `int64` as that is the range of the random value which you have.

Well, yeah, that's what I originally had it to return, but then `nextBigInt` works with `bigint`... I guess I'll change it. 👍

moodmosaic: >Also it should really return only an `int64` as that is the range of the random value which you have.
>>Well, yeah, that's what I originally had it to return, but then `nextBigInt` works with `bigint`... I guess I'll change it. 👍

Hmm... Too many casts going on after returning `int64` from `next`:

```diff
$ git diff
diff --git a/Jack/Seed.fs b/Jack/Seed.fs
index c18de54..b3bbbc4 100644
--- a/Jack/Seed.fs
+++ b/Jack/Seed.fs
@@ -100,20 +100,20 @@ module Seed =
           Gamma = mixGamma s + goldenGamma }

     /// Returns the next pseudo-random number in the sequence, and a new seed.
-    let next (s : Seed) : bigint * Seed =
-        bigint (mix64 s.Value), nextSeed s
+    let next (s : Seed) : int64 * Seed =
+        mix64 s.Value, nextSeed s

     /// Generate a random bigint in the specified range.
     let rec nextBigInt (lo : bigint) (hi : bigint) (seed : Seed) : bigint * Seed =
         if lo > hi then
             nextBigInt hi lo seed
         else
-            let rec loop hilo (v0, seed0) =
+            let rec loop hilo (v0 : int64, seed0) =
                 if hilo > bigint.Zero then
-                    lo + v0 % hilo, seed0
-                else if v0 < lo || v0 >= hi then
+                    lo + (bigint v0) % hilo, seed0
+                else if (bigint v0) < lo || (bigint v0) >= hi then
                     loop hilo (next seed0)
-                else v0, seed0
+                else (bigint v0), seed0
             loop <| hi - lo + bigint.One <| next seed

     /// Splits a random number generator in to two.
```

Jack/Seed.fs
moodmosaic: I just wrote some comments which perhaps we should commit (?)

```diff
diff --git a/Jack/Seed.fs b/Jack/Seed.fs
index c18de54..c9e6769 100644
--- a/Jack/Seed.fs
+++ b/Jack/Seed.fs
@@ -109,11 +109,22 @@ module Seed =
             nextBigInt hi lo seed
         else
             let rec loop hilo (v0, seed0) =
+                // hilo is the size of the exclusive range (hi - lo + 1)
                 if hilo > bigint.Zero then
+                    // The range is a positive number – return a
+                    // number in range using modular arithmetic:
+                    // - http://codereview.stackexchange.com/a/29110/10541
+                    // - http://stackoverflow.com/a/1202706/467754
                     lo + v0 % hilo, seed0
-                else if v0 < lo || v0 >= hi then
-                    loop hilo (next seed0)
-                else v0, seed0
+                else
+                    if v0 < lo || v0 >= hi then
+                        // The range is a negative number – this shouldn't
+                        // happen (unless there is an arithmetic overflow)
+                        // so in this case keep going.
+                        loop hilo (next seed0)
+                    else
+                        // We're in range, return gen'ed number and seed.
+                        v0, seed0
             loop <| hi - lo + bigint.One <| next seed

     /// Splits a random number generator in to two.
```

moodmosaic:
Thanks for the feedback! I've added some commits that (hopefully) address all the comments above.

Jack/Seed.fs
```diff
@@ -32,87 +58,76 @@ module Seed =
...
-            lo + v % k, seedN
+            let rec loop hilo (v0, seed0) =
+                let v0 = bigint (v0 : int64)
+                // hilo is the size of the exclusive range: 'hi - lo + 1'.
+                if hilo > bigint.Zero then
+                    // The range is a positive number – return a
+                    // number in range using modular arithmetic:
+                    // - http://codereview.stackexchange.com/a/29110/10541
+                    // - http://stackoverflow.com/a/1202706/467754
+                    lo + v0 % hilo, seed0
```
jystic: So the issue here that I was worried about (and having the cast above is good as it makes it even clearer) is that `v0` will only ever be in the range `-2^63` to `2^63-1` because it's just a 64-bit number. So if `hilo` is greater than `2^64` then we'll never be able to generate numbers which cover the entire range of values.

moodmosaic: Exactly.

moodmosaic: Yes, I think so, too. Let's switch to old code for `nextBigInt`.

moodmosaic:
Based on your [comment](https://github.com/jystic/dotnet-jack/pull/44#discussion_r86903281) I've now added https://github.com/jystic/dotnet-jack/pull/44/commits/cec7cb8c756751d4206f9ceddb3bd7f92db5a587, so that we don't have to reinvent the wheel...

moodmosaic:
I've now rerun the tests:

`./build.sh` times on `splitmix` branch:

```
=== TEST EXECUTION SUMMARY ===
   Jack.Tests  Total: 49, Errors: 0, Failed: 0, Skipped: 0, Time: 41.557s
Finished Target: Test

---------------------------------------------------------------------
Build Time Report
---------------------------------------------------------------------
Target     Duration
------     --------
Build      00:00:04.5533561
Test       00:00:42.6447674
Total:     00:00:47.1981235
Status:    Ok
---------------------------------------------------------------------
```

`./build.sh` times on `master` branch:

```
=== TEST EXECUTION SUMMARY ===
   Jack.Tests  Total: 49, Errors: 0, Failed: 0, Skipped: 0, Time: 41.152s
Finished Target: Test

---------------------------------------------------------------------
Build Time Report
---------------------------------------------------------------------
Target     Duration
------     --------
Build      00:00:04.9401624
Test       00:00:42.2770417
Total:     00:00:47.2172041
Status:    Ok
---------------------------------------------------------------------
```

jystic:
Fantastic, this is really great 🎉 will rebase and merge

moodmosaic:
Thanks for the great feedback and review!

moodmosaic:
...Maybe we should do the same for Jack [in Haskell](https://github.com/ambiata/disorder.hs/tree/master/disorder-jack) [and PureScript](https://github.com/jystic/purescript-jack)? :smiley:

jystic:
Yeah definitely keen to have this in PureScript! We'd need to make some tweaks, PureScript doesn't have 64-bit integers for example. Do you think this can be implemented using 32-bit integers alone?

Haskell is a tricky one, QuickCheck is so pervasive in the Haskell community that it's useful to build on top of it rather than creating a separate library, although having a ground up re-implementation similar to what we have here would certainly have its benefits.

PR #56, opened 2016-12-11
Use 'Integrated' instead of 'Automatic' when talking about Shrinking

moodmosaic:


moodmosaic:
Closing this because of #57.

PR #61, opened 2016-12-16
Use Jack instead of `dotnet-jack`,

moodmosaic:
because the _dotnet_-part can be confusing after .NET Standard becomes mainstream:
https://github.com/dotnet/standard/blob/master/docs/faq.md

Any thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,9 @@
 
 # dotnet-jack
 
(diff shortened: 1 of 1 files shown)


jystic:
It would be good to differentiate it from the other Jack implementations in some way. I'm ok referring to it as just Jack in most of the readme, but probably the title should mention .NET? /not sure

moodmosaic:
>but probably the title should mention .NET? /not sure

Agreed.

In fact, now that I've had another look into the readme, I think it makes sense to keep the title as `dotnet-jack` (the way it was).

So I've just updated the PR (forced update via `git push --force`).

jystic:
Cool, sounds good 👍

jystic:
Rebased and merged

PR #67, opened 2017-01-24
Add an example of generating regex-constrained strings

moodmosaic:


moodmosaic:
A preview of how it looks like is shown [here](https://github.com/moodmosaic/dotnet-jack/tree/topic/readme-integrations-fare#integrations).

jystic:
Awesome 🍰

PR #68, opened 2017-01-24
Remove Jack/LICENSE since it's stored in /LICENSE

moodmosaic:
We already store the license here in [/LICENSE](https://github.com/jystic/dotnet-jack/blob/master/LICENSE).


jystic:
Rebased and merged ⚡️

PR #75, opened 2017-03-23
Change dice image URL

moodmosaic:
So that it shows up correctly on the README. Right now it points to the old repo, which no longer exists.


jystic:
👍 merged and rebased

Issue #69, opened 2017-02-18
Shrink README


moodmosaic:
Maybe we should have a more compact README, like [the one in TypeScript](https://github.com/Microsoft/TypeScript/blob/0212319b170e0501b97a50170eb05ad5a4da5fa3/README.md), or similar? I think, right now it's a bit [TL;DR](https://en.wikipedia.org/wiki/TL;DR). Any content that we take out could be moved to specific `docs/Xyz.md` files.

jystic:
Yeah I think that's a good idea 👍

Issue #80, opened 2017-03-29
Add the ability to discover and run all properties


moodmosaic:
in a file, or a module<del>, using [Erased Provided Types](https://docs.microsoft.com/en-us/dotnet/articles/fsharp/tutorials/type-providers/creating-a-type-provider#details-about-erased-provided-types)</del>.

I haven't thoroughly think about this, but it should act similarly to adding the Template Haskell splice which runs all the properties in Haskell.

jystic:
Yeah that would be pretty cool 👍

moodmosaic:
I've looked into this and I think a TP seems like an overkill. At the end, I might end up just using Reflection.

ploeh:
As a knee-jerk reaction, I think this should be the responsibility of a testing framework/library. If I correctly understand this suggestion, this is what Gerard Meszaros calls _Test Discovery_ in [xUnit Test Patterns](http://bit.ly/xunitpatterns).

Various libraries have different ways they perform test discovery. For example, [xUnit.net](https://xunit.github.io) uses reflection, whereas [Expecto](https://github.com/haf/expecto) uses lists of tests as first-class citizens.

If such a feature was added to Hedgehog, could it still accommodate such divergent approaches by third parties?

moodmosaic:
I've been going back and forth on that one. On one hand, [I agree that test discovery is the responsibility of a testing framework/library](https://twitter.com/search?l=&q=test%20discovery%20from%3Anikosbaxevanis&src=typd) like xUnit.net or Expecto.

On the other hand, I was thinking that, perhaps, the public APIs of the Haskell and F# version ought to be similar (?)

(In fact, that's what we've been doing with `Seed`, `Shrink`, and `Range`, which are modeled closely to Haskell version. This is also the plan for `Tree` and `Gen`, in order to have a solid foundation to work on #110.)

If we go and look [Hedgehog's public API on Hackage](https://hackage.haskell.org/package/hedgehog/docs/Hedgehog.html), at the time of this writing (v4.0.1), we'll notice that test discovery is part of the public API. It's basically for this reason that I've opened this item...

---

Perhaps, we could skip test discovery from the F# version—after all, I assume that most F# Hedgehog users won't be using this tool from command-line, but they'll be using it with Expecto, xUnit.net, or similar.

I *do* like however the pretty printing and diff output of the Haskell version. Should we have something similar with Expecto or xUnit.net, that'd be great.

[...]

PR #81, opened 2017-03-30
Compact README.md

moodmosaic:
Perhaps we <del>should</del>could have a more compact README (#69). So I've basically
- moved the current README.md to doc/TUTORIAL.md
- added CONTRIBUTING.md
- compacted README.md and added links to above files

A preview of how it looks like is shown [here](https://github.com/moodmosaic/dotnet-hedgehog/blob/topic/compact-readme/README.md).

---

**Edit**:  I *suck* at writing READMEs, so for this reason I took inspiration from [here](https://github.com/Microsoft/TypeScript/blob/0212319b170e0501b97a50170eb05ad5a4da5fa3/README.md) :satisfied:


moodmosaic:
We've decided to close this and take a minimalistic approach, similar to [haskell-hedgehog](https://github.com/hedgehogqa/haskell-hedgehog)'s [README](https://github.com/hedgehogqa/haskell-hedgehog/blob/master/README.md). As soon as we're done with #80 and #85, we could more or less copy that README here.

Issue #83, opened 2017-04-04
We should probably rename this to fsharp-hedgehog


moodmosaic:
Specially if I'm done with what I'm working on #80, this project will be more like an F# project, than a general .NET one.

After all, computation expressions like `gen` and `property` (and what I'll probably come up with #80) aren't consumable outside of F#.

I think that should be a fair rename, like we have haskell-hedgehog (instead of ghc-hedgehog) and scala-hedgehog (instead of jvm-hedgehog).

jystic:
I had hoped that we could have a subset of this library which would be usable from C#, but I could be convinced otherwise.

I don't think that it would ever be possible for the scala-hedgehog to be used from Java on the other hand, as the features we're using of Scala's type system are far away from what Java supports.

moodmosaic:
>I had hoped that we could have a subset of this library which would be usable from C#

This would require a bit of extra maintenance from our side, and I can't convince myself that it's worth it...

Basically, we'd have to create and maintain some sort of [façade](https://en.wikipedia.org/wiki/Facade_pattern), to make it easier for C# developers to call in to our various F# functions in the `Gen` and `Property` modules.

I'm not saying that it's not possible to use Hedgehog from C# as-is, just that it can easily cause a headache:

```c#
using Hedgehog;
using Microsoft.FSharp.Collections;
using Microsoft.FSharp.Core;
using System;
using System.Linq;

class Program
{
    static void Main()
    {
        var g =
            Gen.list(Gen.@int);
        var k =
            FSharpFunc<FSharpList<int>, Property<Unit>>.FromConverter(
                new Converter<FSharpList<int>, Property<Unit>>(xs =>
                    PropertyModule.ofBool(
                        Enumerable.Reverse(Enumerable.Reverse(xs)).SequenceEqual(
                            xs))));
        var p =
            PropertyModule.forAll(
                g, k);

        PropertyModule.print(p);
        // +++ OK, passed 100 tests.
    }
}
```

The above program, can be written 'natively' as:

```fsharp
module Proram

open Hedgehog

let g =
    Gen.list Gen.int
let p =
    fun xs ->
        List.rev (List.rev xs) = xs
        |> Property.ofBool
    |> Property.forAll g
Property.print p
// +++ OK, passed 100 tests.
```

Of course, the idiomatic way of using Hedgehog is:

```f#
module Proram

open Hedgehog

let g =
    Gen.list Gen.int
let p =
    property {
        let! xs = g
[...]

PR #84, opened 2017-04-08
Fix up all references to the old location in Travis, NuGet, and README

moodmosaic:
See [comment](https://github.com/hedgehogqa/fsharp-hedgehog/issues/83#issuecomment-292697405) in #83.


moodmosaic:
FWIW, I've just added 4d3e48c, which replaces the image we have on the README with the Hedgehog logo, without the hassle of doing this on a separate PR

moodmosaic:
A preview of how it looks like is shown [here](https://github.com/moodmosaic/fsharp-hedgehog/blob/topic/fix-up-references/README.md).

jystic:
The image doesn't seem like it is properly crisp? The edges are a bit soft, maybe it needs to be re-exported again or something? /not sure

moodmosaic:
Yeah, I noticed that, but wasn't sure how to fix it. Should we get rid of the image entirely until we find out why it's happening? It looks fine in the mobile version though.
img_7641

moodmosaic:
OK, I was able to fix it also on the desktop version, using HTML (as explained [here](http://stackoverflow.com/a/12118349/467754)) 🎉

capture

jystic:
Maybe I'm crazy but I still think the edges on the logo / text look a bit soft, I'm going to merge this anyway to push the content through, but it would be good to fix it at some point.

The other [logos](https://github.com/hedgehogqa/haskell-hedgehog/blob/master/img/hedgehog-logo.png) you sent me by email look very sharp by comparison.

moodmosaic:
That's weird. I made two versions, transparent and non-transparent. So I will try with the non-transparent one.

moodmosaic:
Yes, I see what you mean now. It looks blurry on screens that have a higher pixel density (e.g. on my MacBook Pro). Looking at this.

jystic:
Ah that must be what it is, I've been looking at it on a MacBook Pro also

If you look at the Haskell readme you can see I've set the width of the image tag to half the size of the actual image, that seems to work for retina displays.

moodmosaic:
Yeah, I've fixed it, will open a PR in a few minutes 😃

moodmosaic:
OK, I think it should be fixed if/after we merge #86.

This was the issue:

capture

So now I know that I suck *both* at writing READMEs and doing logos 😄

PR #86, opened 2017-04-09
Use higher resolution logo

moodmosaic:
so that it doesn't look blurry on screens that have a higher pixel density.


moodmosaic:
Does this one looks better? 😃

jystic:
Yeah that's heaps better 💥

moodmosaic:
Should we merge it, then? 😃

jystic:
Yep, sorry 🚀

PR #90, opened 2017-04-13
Improve the Visual Studio dev experience

Porges:
This allows you to run the tests in Visual Studio using the built-in Test Explorer window.

image

Doing this required taking the explicit dependency on FSharp.Core 4.0.0.1, since we have to have a version in the output directory of the test project for them to be able to run. I've made the dependency have a lower bound but no upper, so it won't force any particular version onto the consumers of the nuget package. (I was going to do this anyway for the C# bits.)

I've also included the various paket files into the solution to make it easier to view them.

### DIFF
--- a/Hedgehog.sln
+++ b/Hedgehog.sln
@@ -1,8 +1,16 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2012
-Project("{f2a71f9b-5d33-465a-a702-920d77279786}") = "Hedgehog", "src\Hedgehog\Hedgehog.fsproj", "{B338B8C4-1C98-4F42-9C45-56BD7EF2AD0A}"
+# Visual Studio 15
+VisualStudioVersion = 15.0.26403.3
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "Hedgehog", "src\Hedgehog\Hedgehog.fsproj", "{B338B8C4-1C98-4F42-9C45-56BD7EF2AD0A}"
 EndProject
 Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "Hedgehog.Tests", "tests\Hedgehog.Tests\Hedgehog.Tests.fsproj", "{788F88B7-5516-4F89-8652-20F14576E1DD}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution items", "Solution items", "{76532829-C6DB-48F1-ABA8-835E3B718023}"
+	ProjectSection(SolutionItems) = preProject
+		paket.dependencies = paket.dependencies
+		README.md = README.md
+	EndProjectSection
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
--- a/paket.dependencies
+++ b/paket.dependencies
@@ -1,6 +1,8 @@
 redirects: on
 source https://www.nuget.org/api/v2/
 
+nuget FSharp.Core 4.0.0.1
+
 nuget FAKE 4.41.1
 nuget xunit.runner.console 2.1.0
 
--- a/paket.dependencies
+++ b/paket.dependencies
@@ -1,7 +1,15 @@
 redirects: on
 source https://www.nuget.org/api/v2/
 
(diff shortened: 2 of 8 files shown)


Porges:
Wow, that image came out bigger than expected 😀

tests/Hedgehog.Tests/Hedgehog.Tests.fsproj
```diff
@@ -33,29 +33,6 @@
...
-    <Content Include="App.config" />
-  </ItemGroup>
-  <ItemGroup>
-    <ProjectReference Include="..\..\src\Hedgehog\Hedgehog.fsproj">
-      <Name>Hedgehog</Name>
-      <Project>{b338b8c4-1c98-4f42-9c45-56bd7ef2ad0a}</Project>
-      <Private>True</Private>
-    </ProjectReference>
-  </ItemGroup>
   <PropertyGroup>
```
moodmosaic: I guess these items where moved automatically? It's hard to spot if there are differences here on not 😕

src/Hedgehog/Hedgehog.fsproj
```diff
@@ -18,32 +18,27 @@
...
     <FSharpTargetsPath>$(MSBuildExtensionsPath32)\..\Microsoft SDKs\F#\3.0\Framework\v4.0\Microsoft.FSharp.Targets</FSharpTargetsPath>
   </PropertyGroup>
-  <ItemGroup>
-    <Reference Include="mscorlib" />
-    <Reference Include="FSharp.Core, Version=$(TargetFSharpCoreVersion), Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a">
-      <Private>True</Private>
-    </Reference>
-    <Reference Include="System" />
-    <Reference Include="System.Core" />
-    <Reference Include="System.Numerics" />
```
moodmosaic: I guess these items where moved automatically? It's hard to spot if there are differences here on not 😕

paket.dependencies
```diff
@@ -1,12 +1,23 @@
...
+// We want to depend on FSharp.Core 4.0.0.1, but not require
+// that version in the generated nuget package (see paket.template).
+//
+// If we just specify "FSharp.Core 4.0.0.1" then the nuget package
+// has an dependency on that *exact* version.
+//
+// So, we specify 4.0.0.1 as a minimum bound, and set lowest_matching
+// so it won't be updated automatically to a newer one by 'paket update'.
+nuget FSharp.Core >= 4.0.0.1 lowest_matching: true
+
```
moodmosaic: 👍 That's pretty awesome.

Hedgehog.sln
```diff
@@ -1,9 +1,17 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2012
-Project("{f2a71f9b-5d33-465a-a702-920d77279786}") = "Hedgehog", "src\Hedgehog\Hedgehog.fsproj", "{B338B8C4-1C98-4F42-9C45-56BD7EF2AD0A}"
+# Visual Studio 15
```
moodmosaic: I think this change was made automatically by Visual Studio(?)

moodmosaic:
I think this is helpful to those who run their tests using the Visual Studio built-in test runner. After all, these tests are written using xUnit.net, so I don't see why not taking up a dependency on its Visual Studio runner.

Would it be possible to move those lines in the *.fsproj back to their original position? (See [this](https://github.com/hedgehogqa/fsharp-hedgehog/pull/90#discussion_r111318725) and [this](https://github.com/hedgehogqa/fsharp-hedgehog/pull/90/files/5879a9e0029cee507d30daaa4dca5708ebe1ed0e#diff-726cfb13c07cca41c2b35d5eea95b8ed) comment.)

Porges:
It was actually paket that moved all the lines around. It seems to want to put them in the same place every time, so I think if they're moved once they won't move again in future...

Hedgehog.sln
Porges: Yes. It shouldn't prevent older VS's from opening the solution.

paket.dependencies
moodmosaic: FWIW, I just run a build with the NuGet flag to create a package locally. Indeed, it's now requiring FSharp.Core >= 4.0.0.1

image

👍

jystic:
👍 thanks for the contribution, have rebased and merged

Issue #85, opened 2017-04-08
Add Range combinators


moodmosaic:
for full control over the scope of generated numbers and collections.

moodmosaic:
Resolved via #91.

Issue #89, opened 2017-04-13
Keep Gen module in line with the Haskell version


moodmosaic:
It'd be nice to keep `Gen` (and `Range` after we're done with <del>#85</del>) in sync with [Gen.hs](https://github.com/hedgehogqa/haskell-hedgehog/blob/master/hedgehog/src/Hedgehog/Gen.hs) and [Range.hs](https://github.com/hedgehogqa/haskell-hedgehog/blob/master/hedgehog/src/Hedgehog/Range.hs).

Related to https://github.com/hedgehogqa/fsharp-hedgehog/issues/110.

jystic:
I'm 💯 for this, would be great if people could pick up the library in any of the languages and have it be more or less the same at its core

moodmosaic:
Now that #91 is merged and we have the Range combinators in place, I've started modifying the Gen stuff. I think, now with the Range combinators, the stuff in `Random.fs` might not be needed, but I might be wrong.

jystic:
The main reason for the distinction originally was so that I could implement `Gen.list` with better than default shrinking. But I've figured out how to do that now staying within `Gen` by adding the [Gen.freeze](http://hackage.haskell.org/package/hedgehog-0.1/docs/Hedgehog-Gen.html#v:freeze) combinator, which is also used to get subterm shrinking, something we don't have in F# yet.

So yes, I suspect eventually we can get rid of `Random.fs`

moodmosaic:
>eventually

Hmm...

Okay, then this means, for now, we should probably keep `Random.fs` and just modify it so that it takes a `Range<'a>` where possible. (Just like what needs to be done in `Gen.fs` as well.)

jystic:
Yeah I guess the fact that all the size stuff happens in `Random.fs` is problematic, it probably is a bit of a rejig to get rid of `Random.fs` entirely and bring things in line with the Haskell implementation.

jystic:
Although, we still have `Gen.sized`, so maybe you can use that to get the size in `Gen.fs` and just call `Random.range` (and delete `Random.sizedRange`)

moodmosaic:
Hmm... Yeah, maybe that'll work with `Gen.sized` (I'll have to try it).

And once we have the stuff in `Gen.fs` ready to use the Range combinators we may more easily decide whether/how we should get rid of `Random.fs`.

jystic:
[...]

PR #91, opened 2017-04-14
Add Range type and combinators

moodmosaic:
Related to #85.

### DIFF
--- a/src/Hedgehog/Range.fs
+++ b/src/Hedgehog/Range.fs
@@ -18,8 +18,8 @@ module Range =
     //
     // Combinators - Range
     //
-    /// Get the origin of a range. This might be the mid-point or the lower bound,
-    /// depending on what the range represents.
+    /// Get the origin of a range. This might be the mid-point or the lower
+    /// bound, depending on what the range represents.
     ///
     /// The 'bounds' of a range are scaled around this value when using the
     /// 'linear' family of combinators.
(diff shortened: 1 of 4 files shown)


src/Hedgehog/Range.fs
```diff
@@ -0,0 +1,132 @@
...
+    //
+
+    [<AutoOpen>]
+    module Internal =
+        // The functions in this module where initially marked as internal
+        // but then the F# compiler complained with the following message:
+        //
+        // The value 'linearFrom' was marked inline but its implementation
+        // makes use of an internal or private function which is not
+        // sufficiently accessible.
```
jystic: I actually prefer to expose internal functions as long as they are marked as such. I think SemVer is too strict here. It would be good if we could add a caveat to our SemVer compliance to say "any internal functions may change at any time, even in a patch release."

This allows our users more flexibility if they want to experiment with something that requires access to internals. They're taking the risk of breaking changes upon themselves if they do so. You'll see I've structured the Haskell library around this concept.

jystic:
💯

src/Hedgehog/Range.fs
moodmosaic: Oh, I see what you mean.

Actually, these two are exposed too. But to access them, you have to type (or `open`) Hedgehog.Range.Internal instead of Hedgehog.Range.

(But yes, in the beginning I was about to hide them with the `internal` keyword.)

moodmosaic: >It would be good if we could add a caveat to our SemVer compliance to say "any internal functions may change at any time, even in a patch release."

This works nice with #89 too. I'll add it on the README in the next days.

jystic: > Actually, these two are exposed too. But to access them, you have to type (or open) Hedgehog.Range.Internal instead of Hedgehog.Range.

Yeah this is perfect imo

PR #92, opened 2017-04-19
Modify Gen combinators so that they take a Range

moodmosaic:
This pull request is part of a series of pull requests on bringing `Gen` in line with the Haskell version (#89).

### DIFF
--- a/src/Hedgehog/Gen.fs
+++ b/src/Hedgehog/Gen.fs
@@ -420,7 +420,7 @@ module Gen =
         let yMax = System.DateTime.MaxValue.Year
         gen {
             let! y =
-                integral <| Range.constantFrom 2000 yMin yMax
+                integral <| Range.linearFrom 2000 yMin yMax
             let! m =
                 integral <| Range.constant 1 12
             let! d =
--- a/src/Hedgehog/Gen.fs
+++ b/src/Hedgehog/Gen.fs
@@ -165,13 +165,6 @@ module Gen =
     let inline integral (range : Range<'a>) : Gen<'a> =
         create (Shrink.towards <| Range.origin range) (Random.integral range)
 
-    /// Generates a random number from the whole range of the numeric type.
-    let inline bounded () : Gen<'a> =
-        let zero = LanguagePrimitives.GenericZero
-        let range = Range.constantBounded ()
-
-        create (Shrink.towards zero) (Random.integral range)
-
     //
     // Combinators - Choice
     //
@@ -445,7 +438,7 @@ module Gen =
 
     /// Generates a random globally unique identifier.
     let guid : Gen<System.Guid> =
-        gen { let! bs = array' 16 16 <| bounded ()
+        gen { let! bs = array' 16 16 (byte <| Range.constantBounded ())
               return System.Guid bs }
 
     /// Generates a random instant in time expressed as a date and time of day.
--- a/src/Hedgehog/Gen.fs
+++ b/src/Hedgehog/Gen.fs
@@ -317,26 +317,26 @@ module Gen =
     //
 
     // Generates a random character in the specified range.
-    let charRange (lo : char) (hi : char) : Gen<char> =
+    let char (lo : char) (hi : char) : Gen<char> =
         integral <| Range.constant (int lo) (int hi) |> map char
 
     /// Generates a random character.
-    let char : Gen<char> =
+    let charBounded : Gen<char> =
         let lo = System.Char.MinValue
         let hi = System.Char.MaxValue
-        charRange lo hi
+        char lo hi
 
     // Generates a random digit.
     let digit : Gen<char> =
-        charRange '0' '9'
+        char '0' '9'
 
     // Generates a random lowercase character.
     let lower : Gen<char> =
-        charRange 'a' 'z'
+        char 'a' 'z'
 
     // Generates a random uppercase character.
     let upper : Gen<char> =
-        charRange 'A' 'Z'
+        char 'A' 'Z'
 
     // Generates a random alpha character.
     let alpha : Gen<char> =
@@ -354,7 +354,7 @@ module Gen =
 
     /// Generates a random string.
     let string : Gen<string> =
-        choice [alphaNum; char] |> string'
+        choice [alphaNum; charBounded] |> string'
 
     //
     // Combinators - Primitives
--- a/src/Hedgehog/Gen.fs
+++ b/src/Hedgehog/Gen.fs
@@ -162,30 +162,15 @@ module Gen =
     //
 
     /// Generates a random number in the given inclusive range.
-    let inline range (lo : ^a) (hi : ^a) : Gen<'a> =
-        create (Shrink.towards lo) (Random.range lo hi)
+    let inline integral (range : Range<'a>) : Gen<'a> =
+        create (Shrink.towards <| Range.origin range) (Random.integral range)
 
     /// Generates a random number from the whole range of the numeric type.
     let inline bounded () : Gen<'a> =
-        let lo = minValue ()
-        let hi = maxValue ()
         let zero = LanguagePrimitives.GenericZero
+        let range = Range.constantBounded ()
 
-        create (Shrink.towards zero) (Random.range lo hi)
-
-    /// Generates a random number in the given inclusive range, but smaller
-    /// numbers are generated more often than bigger ones.
-    let inline sizedRange (lo : ^a) (hi : ^a) : Gen<'a> =
-        create (Shrink.towards lo) (Random.sizedRange lo hi)
-
-    /// Generates a random number from the whole range of the numeric type, but
-    /// smaller numbers are generated more often than bigger ones.
-    let inline sizedBounded () : Gen<'a> =
-        let lo = minValue ()
-        let hi = maxValue ()
-        let zero = LanguagePrimitives.GenericZero
-
-        create (Shrink.towards zero) (Random.sizedRange lo hi)
+        create (Shrink.towards zero) (Random.integral range)
 
     //
     // Combinators - Choice
@@ -201,7 +186,7 @@ module Gen =
         if Array.isEmpty xs then
             return crashEmpty "xs"
         else
-            let! ix = range 0 (Array.length xs - 1)
+            let! ix = integral <| Range.constant 0 (Array.length xs - 1)
             return Array.item ix xs
     }
 
@@ -223,7 +208,7 @@ module Gen =
                 else
                     pick (n - k) ys
 
-        let! n = range 1 total
+        let! n = integral <| Range.constant 1 total
         return! pick n xs
     }
 
@@ -234,7 +219,7 @@ module Gen =
         if Array.isEmpty xs then
             return crashEmpty "xs" xs
         else
-            let! ix = range 0 (Array.length xs - 1)
+            let! ix = integral <| Range.constant 0 (Array.length xs - 1)
             return! Array.item ix xs
     }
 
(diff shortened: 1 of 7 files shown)


src/Hedgehog/Gen.fs
```diff
@@ -413,36 +400,36 @@ module Gen =
...
-    let byte : Gen<byte> =
-        sizedBounded ()
+    let byte (range : Range<byte>) : Gen<byte> =
+        integral range
 
     /// Generates a random signed byte.
-    let sbyte : Gen<sbyte> =
-        sizedBounded ()
+    let sbyte (range : Range<sbyte>) : Gen<sbyte> =
+        integral range
```
moodmosaic: So, for example here, if I rename `integral` back to `range` the F# compiler wouldn't allow me to apply `range` to `range`.

src/Hedgehog/Gen.fs
```diff
@@ -458,20 +445,29 @@ module Gen =
 
     /// Generates a random globally unique identifier.
     let guid : Gen<System.Guid> =
-        gen { let! bs = byte |> array' 16 16
+        gen { let! bs = array' 16 16 <| bounded ()
```
moodmosaic: Nitpick: I could have used a type annotation here to make it clear we're asking for random bytes (?)

src/Hedgehog/Script.fsx
```diff
@@ -35,7 +36,7 @@ Property.print <| property {
 //
 
 Property.print <| property {
-    let! xs = Gen.list Gen.int
+    let! xs = Gen.list <| Gen.bounded ()
```
moodmosaic: Nitpick: I could have used a type annotation here to make it clear we're asking for random integers. Same for a few other places below where I used `Gen.bounded ()`.

src/Hedgehog/Script.fsx
```diff
@@ -57,8 +58,8 @@ Gen.printSample genLeapYear
 
 // Fails due to integer overflow
 Property.print <| property {
-    let! x = Gen.int
-    let! y = Gen.int
+    let! x = Gen.int <| Range.constantBounded ()
+    let! y = Gen.int <| Range.constantBounded ()
```
moodmosaic: Later, we may also provide some convenience 'default' combinators in `Gen` so that it's still possible to write, say `Gen.int`, without having to pass a `Range`. Something like `Gen.Default.int`.

Hypothetically, these convenience 'default' combinators will use some bounded range internally. This might actually be a good idea also for the Haskell version (happy to add them in both places).

src/Hedgehog/Gen.fs
```diff
@@ -162,30 +162,15 @@ module Gen =
     //
 
     /// Generates a random number in the given inclusive range.
-    let inline range (lo : ^a) (hi : ^a) : Gen<'a> =
-        create (Shrink.towards lo) (Random.range lo hi)
+    let inline integral (range : Range<'a>) : Gen<'a> =
+        create (Shrink.towards <| Range.origin range) (Random.integral range)
 
     /// Generates a random number from the whole range of the numeric type.
     let inline bounded () : Gen<'a> =
```
moodmosaic: Later on, I consider moving `Gen.bounded` to an `Internal` submodule, so it'll be something like `Gen.Internal.bounded` just so that we're in line with the Haskell version for `Gen` and `Range`.

src/Hedgehog/Gen.fs
```diff
@@ -315,28 +300,30 @@ module Gen =
...
     /// size parameter.
     let list (g : Gen<'a>) : Gen<List<'a>> =
-        sized (fun size -> list' 0 size g)
+        sized (fun size -> list' (Range.constant 0 size) g)
 
     /// Generates a non-empty list of random length. The maximum length depends
     /// on the size parameter.
     let list1 (g : Gen<'a>) : Gen<List<'a>> =
-        sized (fun size -> list' 1 size g)
+        sized (fun size -> list' (Range.constant 1 size) g)
```
jystic: I think we should remove `list` and `list1` and rename `list'` to `list`, as the range combinators give a much richer way to generate all kinds of lists.

src/Hedgehog/Gen.fs
```diff
@@ -315,28 +300,30 @@ module Gen =
...
 
     /// Generates an array between 'n' and 'm' in length.
     let array' (n : int) (m : int) (g : Gen<'a>) : Gen<array<'a>> =
-        list' n m g |> map Array.ofList
+        list' (Range.constant n m) g |> map Array.ofList
```
jystic: I think this should take a `Range` instead, and be renamed to `array` (same with all the other collection combinators)

src/Hedgehog/Gen.fs
jystic: I think `bounded` is kind of redundant now that we have `Range.constantBounded` as a combinator

src/Hedgehog/Gen.fs
```diff
@@ -368,13 +355,13 @@ module Gen =
...
     // Generates a random character in the specified range.
     let charRange (lo : char) (hi : char) : Gen<char> =
-        range (int lo) (int hi) |> map char
+        integral <| Range.constant (int lo) (int hi) |> map char
 
     /// Generates a random character.
     let char : Gen<char> =
         let lo = System.Char.MinValue
         let hi = System.Char.MaxValue
-        sizedRange (int lo) (int hi) |> map char
```
jystic: This has changed the semantics, as it used to generate different characters depending on the size

src/Hedgehog/Gen.fs
```diff
@@ -162,30 +162,15 @@ module Gen =
     //
 
     /// Generates a random number in the given inclusive range.
-    let inline range (lo : ^a) (hi : ^a) : Gen<'a> =
-        create (Shrink.towards lo) (Random.range lo hi)
+    let inline integral (range : Range<'a>) : Gen<'a> =
+        create (Shrink.towards <| Range.origin range) (Random.integral range)
```
jystic: The reason I called this `integral` in Haskell is because it's named after the `Integral` type class. The general idea with the combinators in `Gen` (that I took in Haskell, and would like to apply here) is that they should be named exactly after types. F# ad-hoc polymorphism does work a bit differently though, so perhaps `integral` is OK here.

src/Hedgehog/Gen.fs
```diff
@@ -162,30 +162,15 @@ module Gen =
...
-
-    /// Generates a random number from the whole range of the numeric type, but
-    /// smaller numbers are generated more often than bigger ones.
-    let inline sizedBounded () : Gen<'a> =
-        let lo = minValue ()
-        let hi = maxValue ()
-        let zero = LanguagePrimitives.GenericZero
-
-        create (Shrink.towards zero) (Random.sizedRange lo hi)
+        create (Shrink.towards zero) (Random.integral range)
```
jystic: Yeah anything to do with sized/bounded is really covered by `Range` now and shouldn't be here, perhaps with the exception of the `char` combinators because I'm not sure if they can be done with `Range`?

Of course we still need `sized`/`resize` here, just not the ones that generate data based on size.

src/Hedgehog/Gen.fs
jystic: I don't think `bounded` should exist anymore now we have `Range`, so this should be:
```fs
array 16 16 <| byte <| Range.constantBounded ()
```

src/Hedgehog/Gen.fs
```diff
@@ -458,20 +445,29 @@ module Gen =
...
-        gen { let! y = create (Shrink.towards 2000) (Random.range yMin yMax)
-              let! m = range 1 12
-              let! d = range 1 (System.DateTime.DaysInMonth (y, m))
-              let! h = range 0 23
-              let! min = range 0 59
-              let! sec = range 0 59
-              return System.DateTime (y, m, d, h, min, sec) }
+        gen {
+            let! y =
+                integral <| Range.constantFrom 2000 yMin yMax
```
jystic: This might be nicer as `Range.linearFrom`

src/Hedgehog/Gen.fs
jystic: I'm curious, does `char` work with the `Range` combinators? or is it too different from an integer?
If it does, then we should delete `charRange` and change this to:
```fs
char (range : Range<char>)
```
If we can't use characters with `Range` then we should have:
```fs
char (lo : char) (hi : char)
charBounded : Gen<char>
```
to match the `enum` / `enumBounded` combinators in Haskell.

src/Hedgehog/Gen.fs
jystic: This might be nicer as `Range.linearFrom` so we gradually try a larger range of year, but start around `2000`

src/Hedgehog/Script.fsx
jystic: As mentioned above, I think we should delete `Gen.bounded` now we have the `Range` combinators, so this should be:
```fs
let! xs = Gen.list <| Gen.int <| Range.constantBounded ()
```

src/Hedgehog/Script.fsx
jystic: I want to hold off on this as long as possible until the path forward is clear.

I like the minimalism we have in the Haskell version (one combinator per type) and it's not clear to me what the default should necessarily be. In QuickCheck, the default for integers is equivalent to `Gen.int <| Range.linear 0 100`.

It's easy for people to experiment with their own default combinators by combining the primitives that we currently offer (i.e. `Gen.int` and the `Range` combinators)

src/Hedgehog/Gen.fs
moodmosaic: Done now in 438cd3d.

src/Hedgehog/Gen.fs
moodmosaic: Yep. The `char` combinators, based on the comments [below](https://github.com/hedgehogqa/fsharp-hedgehog/pull/92#discussion_r113351766), are now done in 41cde3c.

src/Hedgehog/Gen.fs
moodmosaic: `char` doesn't seem to work with `Range`

image

So I've changed our existing ones to match the ones we have in Haskell. Done in 41cde3c.

src/Hedgehog/Gen.fs
moodmosaic: Done in 7fc57a1.

src/Hedgehog/Script.fsx
moodmosaic: Done in 535abb7.

src/Hedgehog/Script.fsx
moodmosaic: Yeah, I was also going back and forth on that one. Let's make `Gen` similar to Haskell and we may come back to this one later.

src/Hedgehog/Gen.fs
moodmosaic: Done in 535abb7.

src/Hedgehog/Gen.fs
moodmosaic: Done in 7fc57a1 and f7782ca.

src/Hedgehog/Gen.fs
moodmosaic: Done in 14b035a.

moodmosaic:
Great feedback! I've gone through each one of the comments and add some commits to address them.

jystic:
:1st_place_medal: good stuff, sorry for the delay on this one

moodmosaic:
No worries! Thanks for the feedback 👍

moodmosaic:
Hmm... it looks like the Travis CI build fails with 

>System.BadImageFormatException : Method has zero rva

Probably a false negative? Would it be possible to trigger a new build?

PR #94, opened 2017-05-05
Add doc/tutorial, make smaller README

### LINKED ISSUE #69: Shrink README

moodmosaic:
This resolves #69.

It basically

* adds a doc/tutorial.md which contains most of the stuff from the old README.md, updated to use the Range combinators
* ports README.md from Haskell version, with any features we don't currently support being removed for now

You can see a preview [here](https://github.com/moodmosaic/fsharp-hedgehog/blob/topic/readme/README.md).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,84 @@
+hedgehog [![Hackage][hackage-shield]][hackage] [![Travis][travis-shield]][travis]
+========
+
+> Hedgehog will eat all your bugs.
+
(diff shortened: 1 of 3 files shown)


jystic:
🎮

PR #95, opened 2017-05-05
Increment minor version

moodmosaic:
so we can push it on NuGet as Hedgehog 0.2.0.

In version 0.2.0, `Gen` combinators have been modified so they can take a `Range` type, and they are now closer to the Haskell version of Hedgehog.


jystic:
👍 rebased and merged

PR #96, opened 2017-05-10
Add ascii/latin1/unicode character generators

moodmosaic:
In line with https://github.com/hedgehogqa/haskell-hedgehog/pull/73/.

### DIFF
--- a/src/Hedgehog/Gen.fs
+++ b/src/Hedgehog/Gen.fs
@@ -320,8 +320,9 @@ module Gen =
     let char (lo : char) (hi : char) : Gen<char> =
         integral <| Range.constant (int lo) (int hi) |> map char
 
-    /// Generates a random character.
-    let charBounded : Gen<char> =
+    /// Generates a Unicode character, including invalid standalone surrogates:
+    /// '\000'..'\65535'
+    let unicodeAll : Gen<char> =
         let lo = System.Char.MinValue
         let hi = System.Char.MaxValue
         char lo hi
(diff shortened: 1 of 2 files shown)


src/Hedgehog/Gen.fs
```diff
@@ -354,7 +368,7 @@ module Gen =
 
     /// Generates a random string.
     let string : Gen<string> =
-        choice [alphaNum; charBounded] |> string'
+        choice [alphaNum; unicodeAll] |> string'
```
jystic: Maybe this should be just `unicode` given the surrogates won't be a legal string unless they're combined correctly? (sorry I don't know that much about unicode surrogates, but I get the impression this wouldn't be a valid string as such)

moodmosaic: Hmm... True, `unicode` is better default, I'll change it.

And, I think, we need something similar in the Haskell version? Right now it uses min/max bounds of char, no?

jystic: I don't think there is a haskell version, the haskell `Gen.string` takes a `Gen<char>` as an argument

moodmosaic: Done now in 4979fb5.

jystic: hmm, I wonder if we should just rename `string'` -> `string` and delete this, let people make their own choices for characters?

moodmosaic: *That* was exactly what I was thinking right now... We're going to have `Gen.fs` in line with the Haskell one anyway, so good idea. I'll undo 4979fb5 also.

moodmosaic: OK, I just pushed f39c328.

jystic:
👍 sorry for the delay

PR #102, opened 2017-05-24
Add F# interactive examples to the Shrink module

moodmosaic:
These examples are ported from [Shrink.hs](https://github.com/hedgehogqa/haskell-hedgehog/blob/d3e2d75e1141e2f734f3a037cd635946726b9003/hedgehog/src/Hedgehog/Internal/Shrink.hs
), and run automatically when the project is build with

    ./build.sh

or manually with

    ./build.sh Doctest


jystic:
Nice ⛱

PR #103, opened 2017-05-26
Shrink floating binary point types similar to the Haskell version

moodmosaic:
Based on Hedgehog 0.2.2 `Shrink.hs` module's [towardsFloat](
https://github.com/hedgehogqa/haskell-hedgehog/blob/51ebefc972235d07f25a462b6edbfe3ce507ccbc/hedgehog/src/Hedgehog/Internal/Shrink.hs#L38-L62).

### DIFF
--- a/src/Hedgehog/Gen.fs
+++ b/src/Hedgehog/Gen.fs
@@ -409,7 +409,7 @@ module Gen =
 
     /// Generates a random 64-bit floating point number.
     let double : Gen<double> =
-        create Shrink.double Random.sizedDouble
+        create (Shrink.towardsDouble 0.0) Random.sizedDouble
 
     /// Generates a random 64-bit floating point number.
     let float : Gen<float> =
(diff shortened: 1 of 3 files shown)


src/Hedgehog/Gen.fs
```diff
@@ -409,7 +409,7 @@ module Gen =
 
     /// Generates a random 64-bit floating point number.
     let double : Gen<double> =
-        create Shrink.double Random.sizedDouble
+        create (Shrink.towardsDouble 0.0) Random.sizedDouble
```
moodmosaic: I know that at some point we need to change `Gen.double` so that it takes a `Range`, similar to the other primitive combinators. I'd prefer to do this on a separate PR because `Random.sizedDouble` must be changed too and `Seed.nextDouble` must be added first. (Same also for `Gen.float`.)

moodmosaic:
`Range` is in line with the Haskell version, and with this PR `Shrink` too. Next is `Random`, `Seed`, and `Gen`, not necessarily in this order 😃

src/Hedgehog/Gen.fs
jystic: 👍

jystic:
Nice 👍

PR #104, opened 2017-05-31
Change Gen.double and Gen.float so that they take a Range type

moodmosaic:


moodmosaic:
`Range` and `Shrink` are in line with the Haskell version, and with this PR `Seed` too. Next is `Gen` 😃

src/Hedgehog/Seed.fs
```diff
@@ -144,6 +144,20 @@ module Seed =
...
+    let rec nextDouble (lo : double) (hi : double) (seed : Seed) : double * Seed =
+        if lo > hi then
+            nextDouble hi lo seed
+        else
+            let x, seed' =
+                seed |> nextBigInt
+                    (bigint Int32.MinValue)
+                    (bigint Int32.MaxValue)
+            let scaled_x =
+                  (0.5 * lo + 0.5 * hi) + ((0.5 * hi - 0.5 * lo) / (0.5 * 4294967296.0)) * float x
```
jystic: Curious where this algorithm came from / what is the thinking behind it?

moodmosaic: This is `randomIvalDouble` ported from Haskell [System.Random](https://hackage.haskell.org/package/random-1.1/docs/src/System-Random.html), similar to the existing `Seed.nextBigInt` which is really `randomIvalInteger` ported from that same module.

jystic: ah perfect 👍

jystic:
✌️

src/Hedgehog/Seed.fs
moodmosaic: :neckbeard: 😃

PR #105, opened 2017-06-17
Add Range exponential combinators

moodmosaic:
Implementation + doctests are based on the [current (v3.0) Haskell version](https://github.com/hedgehogqa/haskell-hedgehog/blob/0.3/hedgehog/src/Hedgehog/Range.hs).

### DIFF
--- a/src/Hedgehog/Range.fs
+++ b/src/Hedgehog/Range.fs
@@ -164,6 +164,22 @@ module Range =
 
             fromBigInt (z + diff)
 
+        /// Scale an integral exponentially with the size parameter.
+        let inline scaleExponential (sz0 : Size) (z0 : 'a) (n0 : 'a) : 'a =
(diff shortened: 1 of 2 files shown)


src/Hedgehog/Range.fs
```diff
@@ -164,6 +164,22 @@ module Range =
 
             fromBigInt (z + diff)
 
+        /// Scale a floating-point number exponentially with the size parameter.
```
moodmosaic: Oh, this should rather be

```
Scale an integral exponentially with the size parameter.
```

Fixing it 😄

moodmosaic: Done

moodmosaic:
Knock knock, anyone there? 🤠

jystic:
🎉

PR #107, opened 2017-06-18
Increment version to 0.3.0

moodmosaic:
Should only be merged after we merge #105 (and optionally #106) so we can do a release 🤠


jystic:
👍 rebased and merged

Issue #88, opened 2017-04-13
C# support


Porges:
Quick question first: would you accept a PR for C# support in the F# repository, or prefer that it's a separate package?

Otherwise, these are the things (that I know of...) that are needed for nice C# integration:
- [x] Potentially: depend on the FSharp.Core nuget package so that Hedgehog works without additional setup when installed into a C# project.
- [ ] Use `CompiledNameAttribute` to make the real names of the functions more usable from C# (e.g. the byte combinator is currently `Gen.@byte`!) This will be a binary-breaking change.
- [ ] Add extension methods for `Select{Many,}` etc, to allow use of query syntax with `Gen`.
- [ ] Figure out exactly how to expose `Property` to C# code. Assertions won't work in query syntax, so might need to work on something with a boolean return value.

jystic:
There was a bit of discussion here related to that: https://github.com/hedgehogqa/fsharp-hedgehog/issues/83

It is something I'd like to see eventually

moodmosaic:
We've had a similar discussion with @jystic in #83 (see [this](https://github.com/hedgehogqa/fsharp-hedgehog/issues/83#issuecomment-292650360) comment, for example). In short, this is a good idea, but we've had no plans to have such a [façade](https://en.wikipedia.org/wiki/Facade_pattern) anytime soon.

moodmosaic:
Yep, so go for it 😃 

>Use `CompiledNameAttribute` to make the real names of the functions more usable from C# (e.g. the byte combinator is currently `Gen.@byte`!) This will be a binary-breaking change.

A breaking change is fine for now. This project follows [SemVer 2.0.0](http://semver.org/spec/v2.0.0.html), and so according to that, until version 1.0.0 is released, major version zero (0.y.z) is for initial development. Anything may change at any time.

moodmosaic:
@Porges, you might want to split this in 2 pull requests, so it gets reviewed faster:

1. A pull request that adds the `CompiledNameAttribute` attribute on the `Gen` and `Property` combinators
2. A pull request that adds support for the LINQ syntax (including the [FSharp.Core 4.0.0.1](https://www.nuget.org/packages/FSharp.Core/4.0.0.1) NuGet Package)

(That's only a suggestion though 😃)

Porges:
FSharp.Core package is addressed by #90 since I had to add it to get the tests running in VS.

Porges:
[...]

Issue #45, opened 2016-11-11
Support .NET Standard


moodmosaic:
Perhaps, it will allow broader reach if we make sure that the F# version can be used with both .NET <del>Core</del>Standard and Mono. ― A quick comparison between Mono and .NET Core can be found [here](https://docs.microsoft.com/en-us/dotnet/articles/core/#comparison-with-mono).

jystic:
It should work with Mono, I've been developing this on a Mac 👍 

But yes, I'd love support for .NET Core

moodmosaic:
Talking about .NET Core, here's a decent FAQ about [what .NET Standard means for .NET Core](https://github.com/dotnet/standard/blob/master/docs/faq.md) and more.

moodmosaic:
.NET Framework 4.7.1 will have built-in support for .NET Standard 2.0. This means that you don't have to deploy any additional files or use binding redirects to use .NET Standard 2.0 libraries.

There's more at: https://github.com/dotnet/announcements/issues/32

moodmosaic:
Some places to look at:

- https://github.com/dotnet/cli/issues/4874
- https://github.com/SwensenSoftware/unquote/pull/136

Porges:
Latest VS now has proper support for this, so it should be much easier.

Porges:
Hedgehog can probably target something lower than 2.0, as well. It doesn't have much interaction with the outside world!

moodmosaic:
Indeed, basically its only external dependency is `System` and `FSharp.Core`.

I'm still not sure what's the proper way of adding support for .NET Standard, other than starting with a new project and copying all files from the old to the new one.

cmeeren:
> starting with a new project and copying all files from the old to the new one.

I think that would be best.

Porges:
See https://github.com/hedgehogqa/fsharp-hedgehog/pull/153 for a first take.

moodmosaic:
I'm closing this, since we've now merged #153.

PR #109, opened 2017-07-04
Exclude FSharp.Core NuGet dependency

ploeh:
This is per the official F# guidelines at
https://fsharp.github.io/2015/04/18/fsharp-core-notes.html#do-not-bundle-fsharpcore-with-a-library,
the gist of which is:

> Do not include a copy of FSharp.Core with your library or package. If
> you do, you will create havoc for users of your library.
>
> The decision about which FSharp.Core a library binds to is up to the
> application hosting of the library. The library and/or library package
> can place constraints on this, but it doesn’t decide it.

### DIFF
--- a/src/Hedgehog/paket.template
+++ b/src/Hedgehog/paket.template
@@ -18,3 +18,5 @@ iconUrl
     https://github.com/hedgehogqa/fsharp-hedgehog/raw/master/img/SQUARE_hedgehog_300x300.png
 tags
     fsharp, f#, testing
+excludeddependencies
+    FSharp.Core


moodmosaic:
👍 💯

Porges:
I believe this change is incorrect. The guidelines are warning against _including a copy_ of the FSharp.Core DLL in your package. However, this is not what happens. FSharp.Core is only referenced as a dependency, it is not inside the package itself.

With this change to exclude the FSharp.Core dependency, it is not installed for C# projects that reference the nuget package, and they subsequently get a _run-time_ failure due to not having the required DLL:

image

moodmosaic:
@Porges, thank you for looking at this! This is interesting...

So

```
nuget FSharp.Core >= 4.0.0.1 lowest_matching: true
```

in `.paket.dependencies`, which is what we currently have (thanks to #90), is all we need then?

Porges:
Yes, if we reverted 3f7f34c it would be as I described.

I'd like to make sure @ploeh agrees first though 🙂

moodmosaic:
The official F# guidelines around FSharp.Core and the various options in Paket don't help that much 😕 

Perhaps, what's inside `.paket.dependencies` is enough:

```
nuget FSharp.Core >= 4.0.0.1 lowest_matching: true
```

Currently, with the change in the `.paket.template`

```
excludeddependencies
    FSharp.Core
```

if we build a NuGet package out of the `master` branch, the FSharp.Core assembly isn't referenced at all, which causes a compilation error if one tries to consume Hedgehog from C# (see #88, #103).

---

Unless there's a better approach, perhaps we should revert 3f7f34c5e6d77e5d333ec7b0a0f8aaaf78aca573?

ploeh:
AFAICT, this is a situation where it's necessary to make a choice regarding which scenario you want to support the best. You can't have the best of both worlds.

- Add `FSharp.Core` as a dependency, and you'll make development difficult for F# developers.
- Omit `FSharp.Core` as a dependency, and you'll make development difficult for C# (and, I suppose, VB) developers.

Which audience does Hedgehog want to cater to? F# developers, or C# developers?

> The official F# guidelines around FSharp.Core and the various options in Paket don't help that much

I disagree; I think it's fairly clear-cut. I've already quoted an important part above, but [further down, it elaborates](https://fsharp.github.io/2015/04/18/fsharp-core-notes.html#the-fsharpcore-nuget-package):

> If you make the FSharp.Core NuGet package a dependency of your own NuGet package, you will induce a transitive dependency that NuGet package for any users of your package,. This forces them to manage their FSharp.Core dependency more manually via NuGet commands rather than via Visual Studio or other IDE tooling. Use caution if necessary.

Over the years, I've given multiple F# workshops, and I can report that this problem is real. The most common obstacle that everyone runs into is caused by libraries that pull in `FSharp.Core` as a dependency.

This is only exacerbated by [good library design](https://fsharp.github.io/2015/04/18/fsharp-core-notes.html#libraries-target-lower-versions-of-fsharpcore):

> Libraries target lower versions of FSharp.Core

> F# ecosystem libraries should generally target the earliest, most portable version of FSharp.Core feasible.

> If your library is part of an ecosystem, it can be helpful to target the earliest, most widespread language version and the earliest (4.0+) and most portable profiles of the .NET Framework feasible.

This is basically [Postel's law](https://en.wikipedia.org/wiki/Robustness_principle) applied. I don't know what Hedgehog currently does, but I think it ought to follow this principle.

The earlier a version referenced, however, the more likely it is that the code consuming a library like Hedgehog will run into versioning problems. These are much easier to deal with if FSharp.Core isn't explicitly included as a dependency in NuGet packages.

Yes: excluding FSharp.Core will guarantee that Hedehog will fail out of the box in C# code bases. C# developer can, however, still add FSharp.Core explicitly, which is [also documented](https://fsharp.github.io/2015/04/18/fsharp-core-notes.html#a-c-project-referencing-an-f-dll-or-nuget-package-may-need-to-also-have-a-reference-to-fsharpcoredll).

So the question is: where do you currently expect Hedgehog adoption to come from? F# developers? Or C# developers?

Due to the situation related to FSharp.Core, you can't have _seamless_ support for both audiences.

moodmosaic:
As discussed in #83, fsharp-hedgehog primarily targets F# developers, so the way forward is clear now. Thank you @ploeh for the above detailed explanation and highlights. 👍

ploeh:
I don't know if this is a good idea or not, but perhaps it'd help to _also_ publish a NuGet package called e.g. _Hedgehog.CSharp_, which is identical to _Hedgehog_ with the only exception that it also includes a NuGet package dependency in _FSharp.Core_...

Porges:
Thanks for the additional information @ploeh. Do you know what exactly the issues are that this causes? I've used FsCheck (which takes a dependency on FSharp.Core) a fair bit on hobby F# projects without running into any issues (but I haven't used it on anything *large*). Is it something to do with older F# installations?

ploeh:
It changes all the time, which doesn't make things easier.

As an example, FsCheck has a transitive dependency on the _FSharp.Core_ NuGet package. In some releases, the dependency was on an older version of _FSharp.Core_. As an example, [FsCheck 2.8.2](https://www.nuget.org/packages/FsCheck/2.8.2) depends on [FSharp.Core 3.1.2.5](https://www.nuget.org/packages/FSharp.Core/3.1.2.5). That's F# 3.1, IIRC.

When you pull in F# 3.1 via NuGet, it'll override your project settings. If you go to the Visual Studio project settings page, you can no longer edit the F# language version.

This can confuse new users if they try to use e.g. `List.distinct`, which was added in F# 4.

It gets even worse if you pull in another library that depends on another version of _FSharp.Core_.

You can try to update _FSharp.Core_ to a newer NuGet package, but since the NuGet team long ago broke the _binding-redirect_ feature for F#, you'll now have versioning issues, which you'll have to manually fix by hand-editing `app.config` files.

...unless you use Paket to manage your NuGet packages, but even Paket isn't perfect. Particularly, it seems to have a hard time understanding that it should add binding redirects to unit test libraries.

Porges:
Thanks for the longer explanation 👍 

Maybe the nuget tooling situation will get better now that F# is in all the Core stuff...

ploeh:
@Porges I don't know if you found it convincing, but my guess is that it's similar issues that caused The F# Core Engineering Group to publish the now-oft-quoted guidelines...

Porges:
If nuget would always add the correct binding redirect I think it would be fine to take the dependency.

But as you say it doesn't, so I'm happy to drop it and require C# users to install Hedgehog.CSharp.


I suspect that to get nuget to always install the redirect, the FSharp.Core package would have to be installed via nuget in the default project configuration (much like how ASP.NET moved from machine-installed packages to nuget-installed).

Porges:
Bringing up this old issue again. With the new-style F# projects, FSharp.Core is consumed via a nuget package automatically. (Essentially everything I asked for in my last comment came true! 😀)

moodmosaic:
>With the new-style F# projects

Sounds interesting. Do you have any links where we can look at? :smiley: 

>Essentially everything I asked for in my last comment came true! :grinning:

It'd be nice if we can 'fix' this, then, so we keep both C# and F# developers happy at the same time.

cmeeren:
New-style fsproj projects are AFAIK just like new-style csproj projects. In the latest VS, just create a new F# .NET Standard library and you're good to go. :)

moodmosaic:
Oh, I see what you (and @Porges) mean.

Porges:
I'm going to put any further discussion on #45 so we don't split this :)

Issue #108, opened 2017-06-23
build.sh fails with Visual Studio 2017


moodmosaic:
I setup a Virtual Machine with Windows 10, Visual Studio 2017 and making sure to select F# in the installer.

Even though I can compile the solution in Visual Studio 2017, when I run `./build.sh` in Git Bash it fails:

```
$ ./build.sh
Paket version 5.1.8
Performance:
 - Disk IO: 92 milliseconds
 - Runtime: 1 second
Building project with version: LocalBuild
Shortened DependencyGraph for Target Test:
<== Test
   <== Doctest
      <== Build

The resulting target order is:
 - Build
 - Doctest
 - Test
Starting Target: Build
Building project: C:\Users\Nikos\Documents\Hedgehog\fsharp-hedgehog\Hedgehog.sln
  c:\Windows\Microsoft.NET\Framework\v4.0.30319\MSBuild.exe  C:\Users\Nikos\Documents\Hedgehog\fsharp-hedgehog\Hedgehog.sln /t:Rebuild /m      /p:RestorePackages="False" /p:Configuration="Release" /logger:Fake.MsBuildLogger+ErrorLogger,"C:\Users\Nikos\Documents\Hedgehog\fsharp-hedgehog\packages\FAKE\tools\FakeLib.dll"
Microsoft (R) Build Engine version 4.7.2046.0
[Microsoft .NET Framework, version 4.0.30319.42000]
Copyright (C) Microsoft Corporation. All rights reserved.

Build started 6/23/2017 2:16:16 PM.
     1>Project "C:\Users\Nikos\Documents\Hedgehog\fsharp-hedgehog\Hedgehog.sln" on node 1 (Rebuild target(s)).
     1>ValidateSolutionConfiguration:
         Building solution configuration "Release|Any CPU".
     1>Project "C:\Users\Nikos\Documents\Hedgehog\fsharp-hedgehog\Hedgehog.sln" (1) is building "C:\Users\Nikos\Documen
       ts\Hedgehog\fsharp-hedgehog\src\Hedgehog\Hedgehog.fsproj" (2) on node 1 (Rebuild target(s)).
     2>C:\Users\Nikos\Documents\Hedgehog\fsharp-hedgehog\src\Hedgehog\Hedgehog.fsproj(42,3): error MSB4019: The importe
[...]

PR #113, opened 2017-07-10
Sketch of C# support

Porges:
Here's a sketch of supporting LINQ for `Gen` and `Property` in C#.

What I've done here:

1. Add a test that all Gen/Property/Range combinators have standard naming (just tests for capital letter at the moment). Some of the names I've chosen are a little different to the F# ones (I'll summarize these later).
2. Add LINQ methods 
3. Add tests using those methods in C#

I'm not that happy with the `ForAll` needed to wrap all `Gens` in C# at the moment.

It might be better that all `Property`s are written direct-style, like:

    Property.ForAll(gen1, x => ... )

Rather than with LINQ syntax, and reserve the LINQ syntax for writing your own `Gen`s.

Related to https://github.com/hedgehogqa/fsharp-hedgehog/issues/88.


Porges:
Will need to fix CI build to work with new-style .csproj ...

tests/Hedgehog.CSharp.Tests/Hedgehog.CSharp.Tests/Hedgehog.CSharp.Tests.csproj
```diff
@@ -0,0 +1,9 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
moodmosaic: This project won't load on Visual Studio 2015. And there's #108 regarding Visual Studio 2017 😕

src/Hedgehog/Gen.fs
```diff
@@ -221,6 +252,7 @@ module Gen =
     /// is halved. When the size gets to one or less, selections are no longer made
     /// from the recursive list.
     /// <i>The first argument (i.e. the non-recursive input list) must be non-empty.</i>
+    [<CompiledName("ChoiceRecursive")>]
```
moodmosaic: Here we're deviating a bit, but a think it's okay (?) 😃 /cc @jystic

src/Hedgehog/Gen.fs
```diff
@@ -305,63 +342,76 @@ module Gen =
...
 
+    [<CompiledName("Char")>]
     // Generates a random character in the specified range.
     let char (lo : char) (hi : char) : Gen<char> =
         integral <| Range.constant (int lo) (int hi) |> map char
 
     /// Generates a Unicode character, including invalid standalone surrogates:
     /// '\000'..'\65535'
+    [<CompiledName("CharAll")>]
     let unicodeAll : Gen<char> =
```
moodmosaic: I think this could be `UnicodeAll` (?)

src/Hedgehog/Gen.fs
```diff
@@ -305,63 +342,76 @@ module Gen =
...
         let lo = System.Char.MinValue
         let hi = System.Char.MaxValue
         char lo hi
 
     // Generates a random digit.
+    [<CompiledName("CharDigit")>]
```
moodmosaic: Do we need to prefix this with `Char`?

src/Hedgehog/Gen.fs
```diff
@@ -305,63 +342,76 @@ module Gen =
...
     let digit : Gen<char> =
         char '0' '9'
 
     // Generates a random lowercase character.
+    [<CompiledName("CharLower")>]
```
moodmosaic: Do we need to prefix this with `Char`?

src/Hedgehog/Gen.fs
```diff
@@ -305,63 +342,76 @@ module Gen =
...
     let lower : Gen<char> =
         char 'a' 'z'
 
     // Generates a random uppercase character.
+    [<CompiledName("CharUpper")>]
```
moodmosaic: Do we need to prefix this with `Char`?

src/Hedgehog/Gen.fs
```diff
@@ -305,63 +342,76 @@ module Gen =
...
     let upper : Gen<char> =
         char 'A' 'Z'
 
     /// Generates an ASCII character: '\000'..'\127'
+    [<CompiledName("CharAscii")>]
```
moodmosaic: Do we need to prefix this with `Char`?

src/Hedgehog/Gen.fs
```diff
@@ -305,63 +342,76 @@ module Gen =
...
     let ascii : Gen<char> =
         char '\000' '\127'
 
     /// Generates a Latin-1 character: '\000'..'\255'
+    [<CompiledName("CharLatin1")>]
```
moodmosaic: Do we need to prefix this with `Char`?

src/Hedgehog/Gen.fs
```diff
@@ -305,63 +342,76 @@ module Gen =
...
     let latin1 : Gen<char> =
         char '\000' '\255'
 
     /// Generates a Unicode character, excluding invalid standalone surrogates:
     /// '\000'..'\65535' (excluding '\55296'..'\57343')
+    [<CompiledName("Char")>]
```
moodmosaic: Why not just `Unicode`?

src/Hedgehog/Gen.fs
```diff
@@ -305,63 +342,76 @@ module Gen =
...
     let unicode : Gen<char> =
         filter (not << System.Char.IsSurrogate) unicodeAll
 
     // Generates a random alpha character.
+    [<CompiledName("CharAlpha")>]
```
moodmosaic: Do we need to prefix this with `Char`?

src/Hedgehog/Gen.fs
```diff
@@ -305,63 +342,76 @@ module Gen =
...
     let alpha : Gen<char> =
         choice [lower; upper]
 
+    [<CompiledName("CharAlphaNumeric")>]
     // Generates a random alpha-numeric character.
```
moodmosaic: Do we need to prefix this with `Char`?

src/Hedgehog/Gen.fs
```diff
@@ -479,6 +546,7 @@ module Gen =
 
 [<AutoOpen>]
 module GenBuilder =
+    [<CompiledName("Gen")>]
     let gen = Gen.Builder ()
```
moodmosaic: Do we really need a `[<CompiledName>]` here? Maybe this should be `private`?

src/Hedgehog/Range.fs
```diff
@@ -212,8 +222,9 @@ module Range =
     /// >>> Range.bounds 99 <| Range.linear 0 10
     /// (0, 10)
     ///
-    let inline linear (x : 'a) : ('a -> Range<'a>) =
-      linearFrom x x
+    [<CompiledName("Linear")>]
+    let inline linear (x : 'a) (y : 'a) : Range<'a> =
+      linearFrom x x y
```
moodmosaic: What's the motivation for this change? 😃 The idea here is that this, ported, code tries to be as close as possible with the Haskell version; we even ported the doctests 🐻

tests/Hedgehog.CSharp.Tests/Hedgehog.CSharp.Tests/LinqTests.cs
```diff
@@ -0,0 +1,126 @@
...
+        //}
+
+        //[Fact]
+        //public void ShrinkTestt()
+        //{
+        //    Check(
+        //        from s in ForAll(Gen.String(Range.Constant(1, 100), Gen.CharAlpha))
+        //        from s2 in ForAll(Gen.String(Range.Constant(1, 10), Gen.CharAlpha))
+        //        select !s.Contains(s2));
+        //}
```
moodmosaic: Maybe we should remove this commented-out blocks of code and just add some `FIXME`s?

The problem here is that these might get stale, and after some time we might not know what to do with them; and they might end up staying there, commented-out, forever.

tests/Hedgehog.CSharp.Tests/Hedgehog.CSharp.Tests/NameTests.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        }
+
+        [Theory]
+        [MemberData(nameof(AllPublicMembers))]
+        public void AllPublicMembersFollowDotNetNamingGuidelines(MemberInfo mi)
+        {
+            var startsWithUppercaseChar = char.IsUpper(mi.Name.First());
+            Assert.True(startsWithUppercaseChar, $"{mi.Name} should start with uppercase letter");
+        }
```
moodmosaic: I think that's pretty neat 💯 It actually looks similar to what AutoFixture.Idioms does 👍

src/Hedgehog/LinqSupport.fs
```diff
@@ -0,0 +1,70 @@
...
+    let select (g : Gen<'a>) (f : Func<'a, 'b>) : Gen<'b> =
+        Gen.map (fun x -> f.Invoke x) g
+
+    [<Extension>]
+    [<CompiledName("SelectMany")>]
+    let bind2 (g : Gen<'a>) (f : Func<'a, Gen<'b>>) (proj : Func<'a, 'b, 'c>) : Gen<'c> =
+        gen {
+            let! a = g
+            let! b = f.Invoke a
+            return proj.Invoke(a, b)
```
moodmosaic: Nitpick: Throughout the F# codebase we add a space between the member/function and its argument. :neckbeard: (near `proj.Invoke`)

src/Hedgehog/LinqSupport.fs
```diff
@@ -0,0 +1,70 @@
...
+
+    [<Extension>]
+    [<CompiledName("Select")>]
+    let select (p : Property<'a>) (f : Func<'a, 'b>) : Property<'b> =
+        Property.map (fun x -> f.Invoke x) p
+
+    [<Extension>]
+    [<CompiledName("SelectMany")>]
+    let bind2 (pa : Property<'a>) (f : Func<'a, Property<'b>>) (proj : Func<'a, 'b, 'c>) : Property<'c> =
+        Property.bind pa (fun a -> Property.bind (f.Invoke a) (fun b -> Property.success (proj.Invoke(a, b))))
```
moodmosaic: Nitpick: Throughout the F# codebase we add a space between the member/function and its argument. :neckbeard: (near `proj.Invoke`)

src/Hedgehog/Property.fs
```diff
@@ -270,14 +287,19 @@ module Property =
...
         bindGen (toGen m) (toGen << k) |> ofGen
 
+    [<CompiledName("ForAll")>]
     let forAll (gen : Gen<'a>) (k : 'a -> Property<'b>) : Property<'b> =
         let prepend (x : 'a) =
             bind (counterexample (sprintf "%A" x)) (fun _ -> k x) |> toGen
         Gen.bind gen prepend |> ofGen
 
+    [<CompiledName("ForAll")>]
+    let forAll' (gen : Gen<'a>) : Property<'a> = forAll gen success
```
moodmosaic: Nitpick^2:

```f#
let forAll' (gen : Gen<'a>) : Property<'a> = forAll gen success
```

might as well be written as

```f#
let forAll' (gen : Gen<'a>) : Property<'a> =
    forAll gen success
```

so that it's consistent with the rest of the function/member declarations in the codebase. I think it's the same in Haskell version too. /cc @jystic

src/Hedgehog/Gen.fs
Porges: Yes I should remove this, leftover from earlier version

src/Hedgehog/Range.fs
Porges: This is so we don't end up with `FSharpFunc` in the API, otherwise it's tricky for C# to call.

src/Hedgehog/Gen.fs
Porges: I'll follow up about the naming later

tests/Hedgehog.CSharp.Tests/Hedgehog.CSharp.Tests/LinqTests.cs
Porges: I'm going to turn these into proper tests, just thought I'd get some feedback on the PR while I'm in progress ☺️

src/Hedgehog/Gen.fs
jystic: In the Haskell version I actually split this in to two separate combinators [choice](http://hackage.haskell.org/package/hedgehog-0.4.1/docs/Hedgehog-Gen.html#v:choice) + [recursive](http://hackage.haskell.org/package/hedgehog-0.4.1/docs/Hedgehog-Gen.html#v:recursive) which you combine together, so I think this is fine for now.

src/Hedgehog/Property.fs
jystic: Yep, @moodmosaic is correct regarding the style we use.

src/Hedgehog/Range.fs
jystic: I think this is a good change, it won't make a difference to the F# usage and the C# usage will be nicer.

tests/Hedgehog.CSharp.Tests/Hedgehog.CSharp.Tests/LinqTests.cs
```diff
@@ -0,0 +1,126 @@
+﻿using Xunit;
+
+// Import Check and ForAll:
+using static Hedgehog.Property;
```
jystic: Wow, I've been out of the C# loop for a while, it's awesome that you can do this now 💯

jystic:
I really like the direction here, if you can address @moodmosaic's comments then I'd be more than happy to go ahead with this.

jack-pappas:
@Porges Any update on this? It'd be handy to be able to utilize Hedgehog from C# as well as F#, as one can currently do with [FsCheck](https://github.com/fscheck/FsCheck).

Porges:
@jack-pappas yes I do intend to clean this up soon

Porges:
... now that I'm back from vacation 😁

moodmosaic:
@Porges, if you're busy, that's fine 🎉 🚀 Perhaps I can give it a shot so we can merge it?

Porges:
Updated to fix the stylistic issues. I'm not sure if the build will pass though... I may need to downgrade the .csproj (will see in a moment).

Porges:
Okay, I'll need to downgrade the .csproj. I've been fighting for a while to try and get build.cmd working locally so that I can check the CI build... 🤦‍♂️

There's also something I alluded to in the first post, about not having to write your tests in LINQ syntax, since the body could be quite long (and imperative!), which I'll add before we get this completed. One more day :grin:

src/Hedgehog/LinqSupport.fs
```diff
@@ -0,0 +1,70 @@
...
+open System.Runtime.CompilerServices
+
+open Hedgehog
+
+[<Extension>]
+module GenLinqSupport =
+
+    // Support for `where`:
+    [<Extension>]
+    [<CompiledName("Where")>]
```
moodmosaic: Nitpick: This can also be expressed as `[<CompiledName "Where>]` without parenthesis, but I can go ahead and do this after we merge the PR 😅

Porges: That looks a bit weird to me 👽

moodmosaic: Ugh, nevermind then 😅

Porges: happy to change it though, but I haven't seen other F# code doing that... 🙂

moodmosaic: No, don't let this small thing slow you down 😃 

---

It's just that, the closest to our existing code style would really be

```f#
[<CompiledName ("Where")>] // Notice the space.  But then, the space isn't needed.
```

Porges:
I've added a `CheckForAll` method so that you can write bigger properties from C# than is possible in Linq syntax.

It does feel a little clunky, it would be better if there were just a single method that took one `Gen` and if you wanted multiple, then you would pass `Gen.Tuple(Gen.One, Gen.Two, Gen.Three)`... but C# doesn't have destructuring of tuples in lambda arguments, so you'd end up with:

    CheckForAll(
        Gen.Tuple(Gen.One, Gen.Two, Gen.Three),
        tuple => {
            var (one, two, three) = tuple;
            // proceed with the real test...
        });

... which is a whole different clunkiness.

moodmosaic:
@Porges, we now have [fsharp-hedgehog-experimental](https://github.com/hedgehogqa/fsharp-hedgehog-experimental) for trying out new things, like those in https://github.com/hedgehogqa/fsharp-hedgehog/pull/113/commits/ed75041fe59082153a0a659128c41cfabee2db10 😃

The build still fails, so I can't test it locally 😕

moodmosaic:
I mean, it'd be great if we can focus just on the essential stuff to enable support for C#/LINQ and we can tackle all the rest separately, here or in [fsharp-hedgehog-experimental](https://github.com/hedgehogqa/fsharp-hedgehog-experimental).

Porges:
Putting it into `experimental` is a good idea, it would be nice to get some real-world usage to figure out what would work best. It does mean you have to put the main body of the test into another method or a local method if you want to write multiple lines of test, but that's not _too_ onerous 😉 

I've backed out `CheckForAll` for now and downgraded the test project, hopefully this fixes it 🙂

moodmosaic:
Weird, the build server is green but the build fails locally (Visual Studio 2015, F# 4.0):

```
Build FAILED.

       "E:\Notes\fsharp-hedgehog\Hedgehog.sln" (Rebuild target) (1) ->
       "E:\Notes\fsharp-hedgehog\tests\Hedgehog.CSharp.Tests\Hedgehog.CSharp.Te
       sts.csproj" (Rebuild target) (4) ->
       (CoreCompile target) ->
         NameTests.cs(32,48): error CS1026: ) expected [E:\Notes\fsharp-hedgeho
       g\tests\Hedgehog.CSharp.Tests\Hedgehog.CSharp.Tests.csproj]
         NameTests.cs(32,68): error CS1002: ; expected [E:\Notes\fsharp-hedgeho
       g\tests\Hedgehog.CSharp.Tests\Hedgehog.CSharp.Tests.csproj]
         NameTests.cs(32,68): error CS1513: } expected [E:\Notes\fsharp-hedgeho
       g\tests\Hedgehog.CSharp.Tests\Hedgehog.CSharp.Tests.csproj]
         NameTests.cs(37,46): error CS1026: ) expected [E:\Notes\fsharp-hedgeho
       g\tests\Hedgehog.CSharp.Tests\Hedgehog.CSharp.Tests.csproj]
         NameTests.cs(38,90): error CS1002: ; expected [E:\Notes\fsharp-hedgeho
       g\tests\Hedgehog.CSharp.Tests\Hedgehog.CSharp.Tests.csproj]
         NameTests.cs(38,90): error CS1513: } expected [E:\Notes\fsharp-hedgeho
       g\tests\Hedgehog.CSharp.Tests\Hedgehog.CSharp.Tests.csproj]

    0 Warning(s)
    6 Error(s)
```

moodmosaic:
Hmm, it's because I don't have the most recent version of the C# compiler installed, I guess.

moodmosaic:
According to [this](https://stackoverflow.com/a/43048660/467754) answer, here's what needs to be done to support also VS2015:

```diff
$ git diff
diff --git a/paket.dependencies b/paket.dependencies
index fe00929..2a46851 100644
--- a/paket.dependencies
+++ b/paket.dependencies
@@ -10,6 +10,7 @@ source https://www.nuget.org/api/v2/
 // So, we specify 4.0.0.1 as a minimum bound, and set lowest_matching
 // so it won't be updated automatically to a newer one by 'paket update'.
 nuget FSharp.Core >= 4.0.0.1 lowest_matching: true
+nuget Microsoft.Net.Compilers 2.4.0

 nuget FAKE 4.41.1
 nuget xunit.runner.console 2.1.0
diff --git a/tests/Hedgehog.CSharp.Tests/paket.references b/tests/Hedgehog.CSharp.Tests/paket.references
index b023a6b..e355b18 100644
--- a/tests/Hedgehog.CSharp.Tests/paket.references
+++ b/tests/Hedgehog.CSharp.Tests/paket.references
@@ -3,3 +3,4 @@ xunit.core
 xunit.runner.console
 xunit.runner.visualstudio
 FSharp.Core
+Microsoft.Net.Compilers
\ No newline at end of file
```

tests/Hedgehog.CSharp.Tests/NameTests.cs
```diff
@@ -0,0 +1,57 @@
...
+                BindingFlags.Static |
+                BindingFlags.DeclaredOnly;
+
+            foreach (var type in _publicApiTypes)
+            {
+                var members = type.GetMembers(bindingFlags);
+
+                foreach (var member in members)
+                {
+                    // Ignore special members like indexers etc.
```
moodmosaic: The body the `foreach` loop requires features not found in the C# compiler that ships with Visual Studio 2015. We could either refactor the body of this loop to plain-old C# or do something like [this](https://github.com/hedgehogqa/fsharp-hedgehog/pull/113#issuecomment-341953738), so that the project compiles.

Porges: Changed this to C#6-compatible code and forced the project to only use C#6.

moodmosaic: @Porges, @jystic runs into this on his local:

```
LinqTests.cs(4,23): error CS0305: Using the generic type `Hedgehog.Property<a>' requires `1' type argument(s)
```

It's basically that `using static Hedgehog.Property;` one we were talking about yesterday. You were right that we have to get rid of it (it *does* compile in VS2015 but has issue with Mono)...

moodmosaic: I guess, we should probably leave this as-is, and have @jystic install C# 6 on his machine. I think it's fair to have C# 6 as base...

Porges: C#6 is 2015 so yes 😁

PR #115, opened 2017-08-23
Add changelog

moodmosaic:
This is similar to https://github.com/hedgehogqa/haskell-hedgehog/pull/71/. I've noticed that Microsoft updated the NuGet Gallery UI and so the idea here is that we can show a changelog in [Hedgehog NuGet page](https://www.nuget.org/packages/Hedgehog/).


jystic:
👍 great stuff

Issue #118, opened 2017-09-21
Switch to F# Formatting


moodmosaic:
as a better alternative to our custom doctest implementation.

**Benefits**

1. It removes the noise from the tooltips in the API docs on all major F# IDE's and code editors.
2. It has the ability to execute code examples as part of the build process.
3. It makes it easier to have some sort of HTML documentation.

Right now, with our custom doctest implementation, we're tackling only the 2nd item.

---

See also https://github.com/hedgehogqa/fsharp-hedgehog/issues/117.

kurtschelfthout:
Note https://github.com/fscheck/FsCheck/issues/384

jwChung:
Should we use FAKE to solve this issue?

moodmosaic:
Perhaps, FAKE [can make this easier](http://fsprojects.github.io/ProjectScaffold/writing-docs.html#API-docs) but I don't think it's *required*(?) IIRC, @Porges removed it in #153 since all we needed was doable with the `dotnet` command―no need to maintain pesky `build.fsx` files.

FTR, there's another tool, DocFx which is created by Microsoft, it's open source, and I think it's the one being used in https://docs.microsoft.com/.

Sample: [F# Formatting](https://fake.build/apidocs/v5/index.html#Fake.Api)
Sample: [DocFx](https://dotnet.github.io/docfx/api/Microsoft.DocAsCode.ContentPairingInfo.html)

jwChung:
@moodmosaic  I see. Thank you for your explanation.

jwChung:
[As DocFx says as below, ](https://dotnet.github.io/docfx/tutorial/docfx_getting_started.html) it is unlikely to be a solution.

> DocFX is an API documentation generator for .NET, which currently supports C# and VB.

1. If we use [the FSharp.MetadataFormat library](https://fsprojects.github.io/FSharp.Formatting/metadata.html) for API documentation, does it have the ability to execute code examples(Can it check compile errors)?

2. Not using FAKE, which executable program do we use to execute the `FSharp.MetadataFormat.dll` library?

moodmosaic:
>1.
Yes, it looks like [F# support for DocFx](https://github.com/dotnet/docfx/issues/159) isn't there yet. Good catch. So, the [library documentation tool](https://fsprojects.github.io/FSharp.Formatting/metadata.html) of F# Formatting seems like the only choice right now.

>2.
[...]

Issue #117, opened 2017-09-20
IDE-readable API documentation


ploeh:
Is there a special reason that the API documentation is written as it is?

In the Visual Studio IDE, the documentation is hard to read:

![image](https://user-images.githubusercontent.com/242165/30661918-ac78340c-9e45-11e7-9b28-d0ad228826d8.png)

Perhaps the formatting is written like that because it's readable in some other editor..?

moodmosaic:
I must admit, this looks horrible in Visual Studio 2015, and perhaps also in Visual Studio 2017 😮 This is a mistake on my part; I should have checked how they render there.

When I was porting the [doctest](http://hackage.haskell.org/package/doctest) part of [examples in source code comments](https://github.com/hedgehogqa/haskell-hedgehog/blob/891b67c85d637441e76a0151ac02f1eb63006a05/hedgehog/src/Hedgehog/Internal/Range.hs#L134-L146) from Haskell, such a tool wasn't available in F#, so [I first wrote my own‒quick and dirty‒version of doctest in F#](https://github.com/moodmosaic/doctest).

At this time, I was using F# mostly from Visual Studio Code and Atom, and I don't remember having *doctests* rendered in such a horrible way...

moodmosaic:
It's not perfect in Visual Studio Code either, but not as horrible as it renders in Visual Studio...

![image](https://user-images.githubusercontent.com/287532/30683706-2d89e056-9eb7-11e7-8b64-a1a32956c2fb.png)

For now, this should definitely be fixed from the F# Hedgehog's side.

ploeh:
Does it make sense to use a Haskell documentation system for F#?

FWIW, the API documentation for FsCheck normally looks fine in Visual Studio. Would you be interested in one or more pull requests that modifies the Hedgehog F# code comments to that format?

moodmosaic:
These examples, in Hedgehog's documentation, are being actually [executed as part of the build process](https://github.com/hedgehogqa/fsharp-hedgehog/blob/67e3921ec5ea632e6fd02216cba0388c7ff1c138/build.fsx#L22-L34). If you can do similar to the API documentation used by FsCheck then that'll be great. 👍

One or more PRs that modifies the Hedgehog F# code comments to that format will be appreciated.

ploeh:
The 'normal' way to do this with F#, I believe, is to use [F# Formatting](http://fsprojects.github.io/FSharp.Formatting).

[...]

PR #122, opened 2017-10-09
IDE-readable API documentation

moodmosaic:
Addresses #117 by removing `doctests` in favor of F# Formatting (#118). Any `doctests` that don't exist already as example-based tests are added in to the *Hedgehog.Tests* project.

The XML documentation shouldn't be hard to read now on Visual Studio and other tools supporting F#:

image

/cc @jystic @ploeh


jystic:
👍

Issue #135, opened 2017-10-11
Generation is order of magnitude slower than FsCheck


cmeeren:
Hedgehog generates strings (haven't tested other types) more than an order of magnitude slower than FsCheck.

I discovered this after converting tests from FsCheck to Hedgehog and noticing a drastic slowdown of the test suite.

Code to test:

```f#
module H =

  open Hedgehog

  let s = Gen.string (Range.singleton 10) Gen.alphaNum

  let check () = s |> Gen.sample 0 10000

module F =

  open FsCheck

  let alphaNum = ['0'..'9'] @ ['A'..'Z'] @ ['a'..'z']

  let s = Gen.elements alphaNum
          |> Gen.listOfLength 10
          |> Gen.map (fun cs -> System.String.Concat(Array.ofList(cs)))

  let check () = s |> Gen.sample 0 10000
  
```

Result of interactive execution using `#time`:

```
> F.check();;
> Real: 00:00:00.057, CPU: 00:00:00.062, GC gen0: 11, gen1: 0, gen2: 0
val it : string list =
  ["mj3bVSmKEY"; "smj3bVSmKE"; "Ksmj3bVSmK"; "dKsmj3bVSm"; "OdKsmj3bVS";
   "LOdKsmj3bV"; "VLOdKsmj3b"; "7VLOdKsmj3"; "X7VLOdKsmj"; "xX7VLOdKsm"; ...]

> H.check();;
> Real: 00:00:01.838, CPU: 00:00:01.890, GC gen0: 52, gen1: 14, gen2: 3
val it : List<string> =
  ["ffmhtVdB5q"; "283b0i3P88"; "RZUYtob07L"; "PSEp4C9Q1c"; "PkPv63NQ83";
   "b9Gnw9ZN7N"; "3HFgXJB5tl"; "TcwRo7AjON"; "HfZGkhHb0h"; "Oj6Fg0RyBw"; ...]
```

Notice the insane difference: 57 ms for FsCheck vs. 1838 ms for Hedgehog.

cmeeren:
Actually it happens with simple ints too. See below (I have assumed that by default, FsCheck generates ints from -100 to 100 at max size):

module H =

  open Hedgehog

  let i = Gen.int (Range.constant -100 100)
  let check () = i |> Gen.sample 100 100000

module F =

  open FsCheck

  let i = Arb.generate<int>
  let check () = i |> Gen.sample 100 100000
  

Result:

```
[...]

PR #140, opened 2017-10-23
Add convenience combinators

### LINKED ISSUE #139: Include batteries
One of Hedgehog's main strengths, IMHO, is its flexibility due to all the `Gen` combinators. However, practical use entirely depends on the user to define commonly used combinations themselves to avoid too much tedious repetition in the tests, and then either repeat these definitions for each test project, or create a separate solution/assembly with common usecases and add that to all test projects. I think most people will just copy-paste the same common combinations across all of their test projects (copy-paste).

I'd like Hedgehog to have a bit more "batteries included" approach. This would make the framework a much easier sell, so to speak. This could either be done by including common shortcuts in Hedgehog itself, or it could be released as a separate nuget (Hedgehog.Batteries?) maintained by Hedgehog maintainers or someone else. If it's a separate library maintained by someone else, it should preferably be officially endorsed by Hedgehog.

Here are examples of what I'm talking about (I'm not saying everything here must be included):

```f#
namespace Hedgehog.Batteries

module Gen =

  open Hedgehog

  /// Generates an int using sensible defaults (Range.exponentialBounded)
  let dInt = Gen.int <| Range.exponentialBounded()

  /// Generates a string using sensible defaults (Range.exponential 0 50, latin1)
  let dString = Gen.string (Range.exponential 0 50) Gen.latin1

  /// Generates a list using sensible defaults (Range.exponential 0 50)
  let dList g =
    g |> Gen.list (Range.exponential 0 50)

  /// Generates a list with length determined using Range.exponential lower upper
  let eList lower upper =
    Gen.list (Range.exponential lower upper)

  /// Generates a list with length determined using Range.linear lower upper
  let lList lower upper =
    Gen.list (Range.linear lower upper)

  /// Generates a list with length determined using Range.constant lower upper
  let cList lower upper =
    Gen.list (Range.constant lower upper)

  /// Generates null part of the time
  let withNull g =
    g |> Gen.option |> Gen.map (fun xOpt ->
      match xOpt with Some x -> x | None -> null)

  /// Generates a value that is not null
  let noNull g =
    g |> Gen.filter (not << isNull)

  /// Generates a value that is not equal to another value
  let notEqual other =
    Gen.filter ((<>) other)

  /// Generates a value that is not equal to another option-wrapped value
  let notEqualOpt other =
    Gen.filter (fun x -> match other with Some o -> x <> o | None -> true)


  /// Generates a 2-tuple with sorted elements.
  let sorted2 g =
    g |> Gen.map (fun (x1, x2) -> 
           let l = [x1; x2] |> List.sort
           (l.Item 0, l.Item 1))

  /// Generates a 3-tuple with sorted elements.
  let sorted3 g =
    g |> Gen.map (fun (x1, x2, x3) -> 
           let l = [x1; x2; x3] |> List.sort
           (l.Item 0, l.Item 1, l.Item 2))

  /// Generates a 4-tuple with sorted elements.
  let sorted4 g =
    g |> Gen.map (fun (x1, x2, x3, x4) -> 
           let l = [x1; x2; x3; x4] |> List.sort
           (l.Item 0, l.Item 1, l.Item 2, l.Item 3))

  /// Generates a 2-tuple with distinct elements.
  let distinct2 g =
    g |> Gen.filter (fun (x1, x2) -> x1 <> x2)

  /// Generates a 3-tuple with distinct elements.
  let distinct3 g =
    g |> Gen.filter (fun (x1, x2, x3) -> 
           [x1; x2; x3] |> List.distinct = [x1; x2; x3])

  /// Generates a 4-tuple with distinct elements.
  let distinct4 g =
    g |> Gen.filter (fun (x1, x2, x3, x4) -> 
           [x1; x2; x3; x4] |> List.distinct = [x1; x2; x3; x4])

  /// Generates a 2-tuple with strictly increasing elements
  let increasing2 g =
    g |> sorted2 |> distinct2

  /// Generates a 3-tuple with strictly increasing elements
  let increasing3 g =
    g |> sorted3 |> distinct3

  /// Generates a 4-tuple with strictly increasing elements
  let increasing4 g =
    g |> sorted4 |> distinct4

  /// Generates a tuple of datetimes where the range determines the number
  /// of days apart (positive numbers -> increasing order and vice versa).
  let dateInterval dayRange =
    gen {
      let tickRange = 
        dayRange 
        |> Range.map (fun days -> int64 days * System.TimeSpan.TicksPerDay)
      let! ticksApart = Gen.integral tickRange
      let! dt1 = Gen.dateTime |> Gen.filter (fun dt -> 
        dt.Ticks + ticksApart > System.DateTime.MinValue.Ticks
        && dt.Ticks + ticksApart < System.DateTime.MaxValue.Ticks)
      let dt2 = dt1.AddTicks ticksApart
      return dt1, dt2
    }


  /// Generates a list using inpGen together with a function that maps each of
  /// the distinct elements in the list to values generated by outGen. Distinct
  /// elements in the input list may possibly map to the same values. The generated
  /// function throws if called with values not in the input list.
  let withMapTo (outGen:Gen<'b>) (inpGen:Gen<'a list>) : Gen<'a list * ('a->'b)> =
    gen {
      let! inputs = inpGen
      let! outputs = outGen |> Gen.list (Range.singleton inputs.Length)
      let inOutMap = List.zip inputs outputs |> Map.ofList
      return inputs, (fun x -> inOutMap.Item x)
    }

  /// Generates a list using inpGen together with a function that maps each of
  /// the distinct elements in the list to values generated by outGen. Distinct
  /// elements in the input list are guaranteed to map to distinct values. Only
  /// use this if the output space is large enough that a list of distinct output
  /// values is likely to be generated. The generated function throws if called
  /// with values not in the input list.
  let withDistinctMapTo (outGen:Gen<'b>) (inpGen:Gen<'a list>) : Gen<'a list * ('a->'b)> =
    gen {
      let! inputs = inpGen
      let! outputs =
        outGen
        |> Gen.list (Range.singleton inputs.Length)
        |> Gen.filter (fun lst -> lst = List.distinct lst)
      let inOutMap = List.zip inputs outputs |> Map.ofList
      return inputs, (fun x -> inOutMap.Item x)
    }

  /// Inserts the given element at a random place in the list
  let addElement (x:'a) (g:Gen<'a list>) =
    gen {
      let! xs = g
      let! i = Gen.integral (Range.constant 0 xs.Length)
      let l1, l2 = xs |> List.splitAt i
      return List.concat [l1; [x]; l2]
    }
```

cmeeren:
Closes #139.

In addition to combinators mentioned in #139, I decided to add several others I have found helpful. Let me know if any should be removed.

### DIFF
--- a/src/Hedgehog/Gen.fs
+++ b/src/Hedgehog/Gen.fs
@@ -545,7 +545,7 @@ module Gen =
     /// Generates null part of the time.
     let withNull (g : Gen<'a>) : Gen<'a> =
         g |> option |> map (fun xOpt ->
-        match xOpt with Some x -> x | None -> null)
+            match xOpt with Some x -> x | None -> null)
 
     /// Generates a value that is not null.
     let noNull (g : Gen<'a>) : Gen<'a> =
--- a/src/Hedgehog/Gen.fs
+++ b/src/Hedgehog/Gen.fs
@@ -447,6 +447,267 @@ module Gen =
         }
 
     //
+    // Combinators - Convenience
+    //
+
+    /// Fisher-Yates shuffle / Knuth shuffle from
+    /// https://www.rosettacode.org/wiki/Knuth_shuffle#F.23
+    let private shuffle lst =
+        let arr = lst |> List.toArray
+        let swap i j =
+            let item = arr.[i]
+            arr.[i] <- arr.[j]
+            arr.[j] <- item
+        let rnd = new System.Random()
+        let ln = arr.Length
+        [0..(ln - 2)] |> Seq.iter (fun i -> swap i (rnd.Next(i, ln)))
+        arr |> Array.toList
+
+    /// Generates a permutation the specified list (shuffles its elements).
+    let permutationOf (list : 'a list) : Gen<'a list> =
+        gen { return list |> shuffle }
+
+    let private randomizeCase (s:string) =
+        let r = System.Random()
+
+        let randomizeCharCase c =
+            let f = if r.Next() % 2 = 0
+                    then System.Char.ToUpper
+                    else System.Char.ToLower
+            f c
+
+        s |> Seq.map randomizeCharCase |> System.String.Concat
+
+    /// Randomizes the case of the characters in the string.
+    let casePermutationOf (str : string) : Gen<string> =
+        gen { return str |> randomizeCase }
+
+    /// Generates a character that is not whitespace.
+    let notWhiteSpace : (Gen<char> -> Gen<char>) =
+        filter (not << System.Char.IsWhiteSpace)
+
+    /// Generates a character that is not a control character.
+    let notControl : (Gen<char> -> Gen<char>) =
+        filter (not << System.Char.IsControl)
+
+    /// Generates a character that is not punctuation.
+    let notPunctuation : (Gen<char> -> Gen<char>) =
+        filter (not << System.Char.IsPunctuation)
+
+    /// Shortcut for Gen.list (Range.exponential lower upper).
+    let eList (lower : int) (upper : int) : (Gen<'a> -> Gen<List<'a>>) =
+        list (Range.exponential lower upper)
+
+    /// Shortcut for Gen.list (Range.linear lower upper).
+    let lList (lower : int) (upper : int) : (Gen<'a> -> Gen<List<'a>>) =
+        list (Range.linear lower upper)
+
+    /// Shortcut for Gen.list (Range.constant lower upper).
+    let cList (lower : int) (upper : int) : (Gen<'a> -> Gen<List<'a>>) =
+        list (Range.constant lower upper)
+
+    /// Shortcut for Gen.array (Range.exponential lower upper).
+    let eArray (lower : int) (upper : int) : (Gen<'a> -> Gen<array<'a>>) =
+        array (Range.exponential lower upper)
+
+    /// Shortcut for Gen.array (Range.linear lower upper).
+    let lArray (lower : int) (upper : int) : (Gen<'a> -> Gen<array<'a>>) =
+        array (Range.linear lower upper)
+
+    /// Shortcut for Gen.array (Range.constant lower upper).
+    let cArray (lower : int) (upper : int) : (Gen<'a> -> Gen<array<'a>>) =
+        array (Range.constant lower upper)
+
+    /// Shortcut for Gen.seq (Range.exponential lower upper).
+    let eSeq (lower : int) (upper : int) : (Gen<'a> -> Gen<seq<'a>>) =
+        seq (Range.exponential lower upper)
+
+    /// Shortcut for Gen.seq (Range.linear lower upper).
+    let lSeq (lower : int) (upper : int) : (Gen<'a> -> Gen<seq<'a>>) =
+        seq (Range.linear lower upper)
+
+    /// Shortcut for Gen.seq (Range.constant lower upper).
+    let cSeq (lower : int) (upper : int) : (Gen<'a> -> Gen<seq<'a>>) =
+        seq (Range.constant lower upper)
+
+    /// Shortcut for Gen.string (Range.exponential lower upper).
+    let eString (lower : int) (upper : int) : (Gen<char> -> Gen<string>) =
+        string (Range.exponential lower upper)
+
+    /// Shortcut for Gen.string (Range.linear lower upper).
+    let lString (lower : int) (upper : int) : (Gen<char> -> Gen<string>) =
+        string (Range.linear lower upper)
+
+    /// Shortcut for Gen.string (Range.constant lower upper).
+    let cString (lower : int) (upper : int) : (Gen<char> -> Gen<string>) =
+        string (Range.constant lower upper)
+
+    /// Generates null part of the time.
+    let withNull (g : Gen<'a>) : Gen<'a> =
+        g |> option |> map (fun xOpt ->
+        match xOpt with Some x -> x | None -> null)
+
+    /// Generates a value that is not null.
+    let noNull (g : Gen<'a>) : Gen<'a> =
+        g |> filter (not << isNull)
+
+    /// Generates a value that is not equal to another value.
+    let notEqualTo (other : 'a) : (Gen<'a> -> Gen<'a>) =
+        filter ((<>) other)
+
+    /// Generates a value that is not equal to another option-wrapped value.
+    let notEqualToOpt (other : 'a option) : (Gen<'a> -> Gen<'a>) =
+        filter (fun x -> match other with Some o -> x <> o | None -> true)
+
+    /// Generates a value that is not contained in the specified list.
+    let notIn (list: 'a list) (g : Gen<'a>) : Gen<'a> =
+        g |> filter (fun x -> not <| List.contains x list)
+
+    /// Generates a list that does not contain the specified element.
+    /// Shortcut for Gen.filter (not << List.contains x)
+    let notContains (x: 'a) : (Gen<'a list> -> Gen<'a list>) =
+      filter (not << List.contains x)
+
+    /// Generates a string that is not equal to another string using
+    /// StringComparison.OrdinalIgnoreCase.
+    let iNotEqualTo (str : string) : (Gen<string> -> Gen<string>) =
+        filter <| fun s ->
+            not <| str.Equals(s, System.StringComparison.OrdinalIgnoreCase)
+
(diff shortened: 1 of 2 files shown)


moodmosaic:
The additional ones you’ve added are interesting, but it’d allow us to move faster if you could propose them on a separate PR.

It will also be easier for @jystic to follow up on this PR if we keep it focused on those combinations we discussed in #139.

Thank you for all the your hard work, @cmeeren!

cmeeren:
> The additional ones you’ve added are interesting, but it’d allow us to move faster if you could propose them on a separate PR.

Sure, I can do that, but personally I'm in no hurry. Do you think it's necessary, or could we just make the decision on each combinator in this PR?

Also, the new improved implementation of `withDistinctMapTo` makes use of the new `notIn` combinator, so either that'd need to be included, or it would need to be manually coded into the `withDistinctMapTo` implementation. And if `notIn` is to be included, then `notContains` is a natural companion. And then we're very close to `notSubstringOf` which is kind of the same only for strings, and so on and so forth.

Still, if you want me to remove some/all of the ombinators not mentioned in #139 and include them in a separate PR after this is merged, I'll make it happen. Let me know how you want to proceed.

src/Hedgehog/Gen.fs
```diff
@@ -447,6 +447,267 @@ module Gen =
...
+      filter <| fun s ->
+          str.IndexOf(s, System.StringComparison.OrdinalIgnoreCase) = -1
+
+    /// Generates a string that does not start with another string.
+    let notStartsWith (str : string) : (Gen<string> -> Gen<string>) =
+      filter <| fun s -> not <| s.StartsWith str
+
+    /// Generates a string that does not start with another string using
+    /// StringComparison.OrdinalIgnoreCase.
+    let iNotStartsWith (str : string) : (Gen<string> -> Gen<string>) =
```
moodmosaic: Perhaps `iNotStartsWith` a bit cryptic 😕 Could we come up with a more descriptive name? 😕

src/Hedgehog/Gen.fs
```diff
@@ -447,6 +447,267 @@ module Gen =
...
+        filter <| fun s ->
+            not <| str.Equals(s, System.StringComparison.OrdinalIgnoreCase)
+
+    /// Generates a string that is not a substring of another string.
+    let notSubstringOf (str : string) : (Gen<string> -> Gen<string>) =
+      filter <| fun s -> not <| str.Contains s
+
+    /// Generates a string that is not a substring of another string using
+    /// StringComparison.OrdinalIgnoreCase.
+    let iNotSubstringOf (str : string) : (Gen<string> -> Gen<string>) =
```
moodmosaic: Perhaps `iNotSubstringOf` is a bit cryptic 😕 Could we come up with a more descriptive name? 😕

src/Hedgehog/Gen.fs
```diff
@@ -447,6 +447,267 @@ module Gen =
...
+        g |> filter (fun x -> not <| List.contains x list)
+
+    /// Generates a list that does not contain the specified element.
+    /// Shortcut for Gen.filter (not << List.contains x)
+    let notContains (x: 'a) : (Gen<'a list> -> Gen<'a list>) =
+      filter (not << List.contains x)
+
+    /// Generates a string that is not equal to another string using
+    /// StringComparison.OrdinalIgnoreCase.
+    let iNotEqualTo (str : string) : (Gen<string> -> Gen<string>) =
```
moodmosaic: Perhaps `iNotEqualTo` is a bit cryptic 😕 Could we come up with a more descriptive name? 😕

src/Hedgehog/Gen.fs
```diff
@@ -447,6 +447,267 @@ module Gen =
...
+        let randomizeCharCase c =
+            let f = if r.Next() % 2 = 0
+                    then System.Char.ToUpper
+                    else System.Char.ToLower
+            f c
+
+        s |> Seq.map randomizeCharCase |> System.String.Concat
+
+    /// Randomizes the case of the characters in the string.
+    let casePermutationOf (str : string) : Gen<string> =
```
moodmosaic: This one uses `randomizeCase` which in turn uses `System.Random ()` under the cover, which we should consider using (or not using) once we're done with #135.

For this reason, I'd like to propose to discuss about it separately, on next PR(s), and not include it on this one.

src/Hedgehog/Gen.fs
```diff
@@ -447,6 +447,267 @@ module Gen =
...
+            arr.[i] <- arr.[j]
+            arr.[j] <- item
+        let rnd = new System.Random()
+        let ln = arr.Length
+        [0..(ln - 2)] |> Seq.iter (fun i -> swap i (rnd.Next(i, ln)))
+        arr |> Array.toList
+
+    /// Generates a permutation the specified list (shuffles its elements).
+    let permutationOf (list : 'a list) : Gen<'a list> =
+        gen { return list |> shuffle }
```
moodmosaic: This one uses `shuffe` which in turn uses `System.Random ()` under the cover, which we should consider using (or not using) once we're done with #135.

For this reason, I'd like to propose to discuss about it separately, on next PR(s), and not include it on this one.

cmeeren:
> I suggest we take on separately: ... 

Sure, makes sense. I'll remove them from this PR.

> And perhaps we can find better names for i...

We could call it `notSubstringOfIgnoringCase` etc., but that's awfully long. I can't really think of anything shorter. I'm used case-insensitive functions being prefixed with `i`, so for me `iNotSubstringOf` isn't a problem. Also, I'd guess users wanting this functionality are probably writing `Gen.substring` to see which functions turn up in Intellisense, and the function documentation makes it immediately clear what it does.

Let me know if you have a better name; I can't think of any that does not use `i` and is not awfully long-winded.

moodmosaic:
>Let me know if you have a better name; I can't think of any that does not use `i` and is not awfully long-winded.

I can't currently come up with a better name...

In fact, the other option, `notSubstringOfIgnoringCase`, hints that we're doing 2 things in 1 call, and perhaps we should refactor 1 of those 2 things as a parameter...

Still, we can discuss as much as we can on separate PR(s), so instead of jumping the gun by 'forcing' a name we're not entirely sure about, we can move all those `iNotXyz` ones on a separate PR...

The more short and focused a PR is, the easier it comes for reviewers to find a continuous time-slot to review it, try it, and perhaps even merge it. (Some interesting tips [here](http://blog.ploeh.dk/2015/01/15/10-tips-for-better-pull-requests/).)

cmeeren:
Thank you for the link, Mark makes good sense as always. I have now removed both the `i...` and its corresponding case-sensitive brethren.

cmeeren:
One thing I'd like to do is add unit tests for some of the combinators. However, I can't see how this can be done sensibly without using property-based testing and auto-generated values. How do you feel about using basic Hedgehog generators/combinators to test the more advanced ones?

moodmosaic:
If you want to add automated tests, there are a few ways to do this:

* Add example(s) in the Src/Script.fsx
* Add example-based (traditional) tests using xUnit.net in the Hedgehog.Tests project
* Add property-based tests using Hedgehog in the Hedgehog.Tests project which already references Hedgehog anyway.

moodmosaic:
Up to now, we usually add examples in the Src/Script.fsx...

cmeeren:
> Up to now, we usually add examples in the Src/Script.fsx...

As far as I can see, that serves only to exemplify, not test for correct behavior.

I've added some tests now to `GenTests.fs`.

Also, I'm wondering if there really is much utility for `cArray`, `cSeq` , etc. (since one can always use `cList` etc. together with `Array.fromList` etc., and in my experience, I mostly only use lists in tests). What do you think, should I remove the six `xArray`/`xSeq` functions?

moodmosaic:
The more you remove the better; because once you add them it's *breaking* change to remove something.

Yes, we're on version < 1, and according to Semantic Versioning 2.0, breaking changes are allowed... Still, we should try to conservative when possible.

cmeeren:
True that. All right, I've removed some more now. It's very close to #139; the only additions are `eString`, `lString`, `cString`. `notIn`, and `notContains`. How do you feel about that? :)

cmeeren:
PS, don't feel you need to cater to my whims. If you want those too removed from this PR, I'll do that.

moodmosaic:
Those look fine to me 👍

moodmosaic:
@cmeeren, should we close this one in in favor of https://github.com/hedgehogqa/fsharp-hedgehog-experimental/pull/4?

cmeeren:
Certainly.

Issue #146, opened 2017-11-05
"greedy traversal" test performance


Porges:
This is the test in question:

https://github.com/hedgehogqa/fsharp-hedgehog/blob/1dfb754f0a8f3fa1e8806af11054ec21e545fb2e/tests/Hedgehog.Tests/MinimalTests.fs#L65-L89

I ran this under the profiler and the problem is entirely due to the `sprintf` in `forAll`:

https://github.com/hedgehogqa/fsharp-hedgehog/blob/1dfb754f0a8f3fa1e8806af11054ec21e545fb2e/src/Hedgehog/Property.fs#L276-L281

Replacing this with `x.ToString()` reduces the runtime from to 40 s to 0.55 s (!)

However, `ToString` isn't as useful as `sprintf "%A"` for 'native' F# types. A better fix would be to make this function lazily-invoked, so that it's only called if needed.

moodmosaic:
Very interesting. See also #135.

Porges:
@moodmosaic yes, I suspect there's a bunch more easy fixes like this 🙂

moodmosaic:
Fixed in https://github.com/hedgehogqa/fsharp-hedgehog/pull/147 and released as Hedgehog 0.6.

cmeeren:
Just want to let you guys know that this reduced my test suite for a small-ish project (156 tests) from 14 seconds to 4 seconds. Great job, keep it up! 👍

PR #147, opened 2017-11-05
Make Journal store delayed strings, so they are computed on-demand

Porges:
This fixes the performance issues in the "greedy traversal" test (#146).

I'm not entirely sure about the change to `PropertyBuilder.CounterExample`, but it's not really user-visible (they would be using the `counterexample` 'action' instead of calling this directly).

There's another `string` call which is _not_ delayed (line 275) as it didn't seem to be a performance issue.

### DIFF
--- a/src/Hedgehog/Property.fs
+++ b/src/Hedgehog/Property.fs
@@ -4,7 +4,7 @@ open FSharpx.Collections
 open System
 
 type Journal =
-    | Journal of List<string>
+    | Journal of LazyList<string>
 
 type Result<'a> =
     | Failure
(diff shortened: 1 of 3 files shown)


Porges:
Before:

```
=== TEST EXECUTION SUMMARY ===
   Hedgehog.Tests  Total: 159, Errors: 0, Failed: 0, Skipped: 0, Time: 48.931s
```

After:

```
=== TEST EXECUTION SUMMARY ===
   Hedgehog.Tests  Total: 159, Errors: 0, Failed: 0, Skipped: 0, Time: 2.287s
```

Porges:
Just FYI: `sprintf "%A"` was spending a lot of its time in reflection(!), it's not a fast call. It must have to check the object type for the various F#-specific attributes that control formatting.

Porges:
... actually, that's already meant to exist: https://github.com/fsharp/fslang-suggestions/issues/429

jystic:
👍 This is great

Porges:
Not sure why that happened twice...

moodmosaic:
>Not sure why that happened twice...

No worries, it's now manually merged in https://github.com/hedgehogqa/fsharp-hedgehog/commit/813e1f9c8d95fa4327b315d4329abfca76615158 via `git merge --no-ff`  🎉 

image

PR #153, opened 2018-01-24
Convert to .NET Standard

Porges:
- `Hedgehog` targets .NET Standard 1.6 (the lowest version that there exists a `FSharp.Core` for). This means the new lowest supported .NET Framework is 4.6.1.
- Removed FAKE build for now as there are still issues with .NET Core, and the build itself is very simple anyway (`dotnet build` does everything).
- Updated CI to use `dotnet` directly. Removed the Paket part of the build for now as getting it working is annoying (requires invoking it via `mono`). Paket should have a proper .NET Core-targetted .exe soon, so we can add this back in when it is released.
- Added build instructions.
- Split Paket dependencies into two groups, Build & Test (this is to prevent Test dependencies affecting what we want the nuget package to depend upon).
- Add fixes for some weird F# compiler/tooling bugs. 😢
  - Explicit `main` in test suite to fix initialization problem.
  - Forward slashes instead of backslashes in solution file.
  - Added version to Paket template as it is not properly generated in output assembly.


cmeeren:
Can't we target .NET Standard 2.0? The lowest framework version will still be 4.6.1 (see [version chart](https://github.com/dotnet/standard/blob/master/docs/versions.md)).

AFAIK 1.5 and 1.6 were essentially dead ends and are not recommended for targeting.

Porges:
Is there any point in doing so if all the APIs we need are in 1.6? It does bump the Mono/Xamarin minimums up if we go to 2.0.

cmeeren:
Fair point. In any case, should we need both it's trivial to multitarget (just change `<TargetFramework>` to `<TargetFrameworks>` and use a semicolon-delimited list of targets).

moodmosaic:
Yes, I think we can multitarget if/when needed.

paket.dependencies
```diff
@@ -1,24 +1,36 @@
...
 //
-// If we just specify "FSharp.Core 4.0.0.1" then the nuget package
+// If we just specify "FSharp.Core 4.1.17" then the nuget package
 // has an dependency on that *exact* version.
 //
-// So, we specify 4.0.0.1 as a minimum bound, and set lowest_matching
+// So, we specify 4.1.17 as a minimum bound, and set lowest_matching
 // so it won't be updated automatically to a newer one by 'paket update'.
-nuget FSharp.Core >= 4.0.0.1 lowest_matching: true
+nuget FSharp.Core >= 4.1.17 lowest_matching: true
```
moodmosaic: This means we're going to target F# 4.1?

paket.dependencies
```diff
@@ -1,24 +1,36 @@
 redirects: on
+storage: none
```
moodmosaic: Do we need `storage: none`?

src/Hedgehog/Hedgehog.fsproj
```diff
@@ -1,45 +1,10 @@
...
-    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
-    <ProjectGuid>{B338B8C4-1C98-4F42-9C45-56BD7EF2AD0A}</ProjectGuid>
-    <OutputType>Library</OutputType>
-    <RootNamespace>Hedgehog</RootNamespace>
-    <AssemblyName>Hedgehog</AssemblyName>
-    <TargetFrameworkVersion>v4.5.2</TargetFrameworkVersion>
-    <TargetFSharpCoreVersion>4.4.0.0</TargetFSharpCoreVersion>
+    <TargetFramework>netstandard1.6</TargetFramework>
+    <AssemblyVersion>0.6.0.0</AssemblyVersion>
+    <FileVersion>0.6.0.0</FileVersion>
```
moodmosaic: I think here we need to also take into account the generated NuGet package which right now gets generated from the [`.paket.template`](https://github.com/hedgehogqa/fsharp-hedgehog/blob/3f7f34c5e6d77e5d333ec7b0a0f8aaaf78aca573/src/Hedgehog/paket.template).

We probably need to get rid of that file and move all its content here, as attributes.

paket.dependencies
Porges: Paket recommended this for new-style projects in the upgrade guide

src/Hedgehog/Hedgehog.fsproj
Porges: I tried that but the package that is generated doesn't have the right versions for the dependencies (since only paket knows about the dependencies in `paket.dependencies`).

Porges: e.g `FSharp.Core` is 4.3.

paket.dependencies
Porges: This is the first `FSharp.Core` version that supports .NET Standard.

paket.dependencies
Porges: Note to self: Sounds like I can override this with `storage: packages` on the FAKE dependency to get it working again.

paket.dependencies
moodmosaic: Oh, I see.

paket.dependencies
moodmosaic: Is this the only thing that's blocking us at the moment?

Porges:
Failing test seems to be due to https://github.com/Microsoft/visualfsharp/issues/1371, which I find _highly_ surprising!

moodmosaic:
Yeah, that one is surprising, indeed!

Porges:
BTW, current status is: FAKE still can't build it. I'll need to try updating FAKE to a newer version that might have better support for new-style `.*proj`s.

moodmosaic:
That's weird. I wonder how other projects done it, which I'm 100% sure they use FAKE too. I can try updating to the newest version, hmm...

Porges:
- Added bootstrap for Fake via `dotnet restore` (based on https://github.com/matthid/fake-bootstrap, but I'm assuming `dotnet` will present on the build machine).
- XUnit Fake target fails since it assumes that the test runner will be present in `/tools`. It would be easiest if there was a `dotnet test` target for Fake, since that works fine.

Porges:
It's kind of funny because the build is so simple it could be replaced by `dotnet build && dotnet test && paket pack` 😆

moodmosaic:
>It's kind of funny because the build is so simple it could be replaced by `dotnet build && dotnet test && paket pack` 😆 

Nothing actually stops us from doing that, right? :wink:

Porges:
Switched the Travis build to just use `dotnet` directly for now.

Porges:
Old package:

```xml
<?xml version="1.0" encoding="utf-8" standalone="yes"?>
<package xmlns="http://schemas.microsoft.com/packaging/2011/10/nuspec.xsd">
  <metadata>
    <id>Hedgehog</id>
    <version>0.6.0.0</version>
    <title>Hedgehog</title>
    <authors>Jacob Stanley, Nikos Baxevanis</authors>
    <licenseUrl>https://raw.githubusercontent.com/hedgehogqa/fsharp-hedgehog/master/LICENSE</licenseUrl>
    <projectUrl>https://github.com/hedgehogqa/fsharp-hedgehog</projectUrl>
    <iconUrl>https://github.com/hedgehogqa/fsharp-hedgehog/raw/master/img/SQUARE_hedgehog_300x300.png</iconUrl>
    <description>Hedgehog will eat all your bugs.</description>
    <summary>An alternative property-based testing system for F#, in the spirit of QuickCheck. The key improvement is that shrinking comes for free.</summary>
    <copyright>Copyright 2017</copyright>
    <tags>fsharp f# testing</tags>
    <dependencies />
  </metadata>
</package>
```

New package:

```xml
<?xml version="1.0" encoding="utf-8" standalone="yes"?>
<package xmlns="http://schemas.microsoft.com/packaging/2011/10/nuspec.xsd">
  <metadata>
    <id>Hedgehog</id>
    <version>0.0.0.0</version>
    <authors>Jacob Stanley, Nikos Baxevanis</authors>
    <licenseUrl>https://raw.githubusercontent.com/hedgehogqa/fsharp-hedgehog/master/LICENSE</licenseUrl>
    <projectUrl>https://github.com/hedgehogqa/fsharp-hedgehog</projectUrl>
    <iconUrl>https://github.com/hedgehogqa/fsharp-hedgehog/raw/master/img/SQUARE_hedgehog_300x300.png</iconUrl>
    <description>Hedgehog will eat all your bugs.</description>
    <summary>An alternative property-based testing system for F#, in the spirit of QuickCheck. The key improvement is that shrinking comes for free.</summary>
    <copyright>Copyright 2017</copyright>
    <tags>fsharp f# testing</tags>
    <dependencies>
      <dependency id="FSharp.Core" version="4.1.17" />
      <dependency id="System.Collections.Specialized" version="4.0.1" />
    </dependencies>
  </metadata>
</package>
```

Looks good except:

- version is wrong
- doc xml is missing from package

Porges:
All fixed.

src/Hedgehog/paket.template
```diff
@@ -2,6 +2,8 @@ type
     project
 id
     Hedgehog
+version
+	0.7.0.0
```
moodmosaic: Do we even need `paket.template` now? :sweat_smile:

build.fsx
```diff
@@ -1,22 +0,0 @@
-#r @"packages/FAKE/tools/FakeLib.dll"
```
moodmosaic: Now that `build.fsx`, `build.sh` and `build.cmd` are removed, how should we tell to the users to build the source locally from the command-line? (So we can put it on the README.)

moodmosaic: I guess they would have to do similar with `travis.yml`?

```
./validate.sh
dotnet build -c Release
dotnet test -c Release --no-build tests/Hedgehog.Tests/Hedgehog.Tests.fsproj
dotnet test -c Release --no-build tests/Hedgehog.CSharp.Tests/Hedgehog.CSharp.Tests.csproj
```

And then, how can we build the NuGet package in order to push it on NuGet Gallery?

moodmosaic:
@Porges, this looks awesome and much cleaner than what we have for the full .NET Framework. :100: :rocket: 

I just left a few comments regarding the way we can build locally from the command-line and the creation of the NuGet package. Once those are clear, we should merge this. :horse_racing: 

/cc @jystic

build.fsx
Porges: Will update the README.

src/Hedgehog/paket.template
Porges: At the moment, yes. Building from the `.fsproj` with `dotnet` tooling generates a package with different version bounds...

moodmosaic:
About to try this on macOS, Ubuntu, and Windows. If all goes well, we should be good to go.

paket.dependencies
```diff
@@ -1,24 +1,32 @@
...
+nuget System.Collections.Specialized >= 4.0.1 lowest_matching: true
+
+// These files are imported directly into Hedgehog:
 
 github fsprojects/FSharpx.Collections:e2722ce571f4926586ae907413b4a5511f0ef74e src/FSharpx.Collections/Collections.fs
 github fsprojects/FSharpx.Collections:e2722ce571f4926586ae907413b4a5511f0ef74e src/FSharpx.Collections/LazyList.fsi
 github fsprojects/FSharpx.Collections:e2722ce571f4926586ae907413b4a5511f0ef74e src/FSharpx.Collections/LazyList.fs
 
-nuget Unquote 3.1.2
+// For test projects:
```
moodmosaic: Instead of comments, I think we can use Paket's (dependency groups)[https://fsprojects.github.io/Paket/groups.html] feature. This can be done later though (so we don't hold the PR just for this).

src/Hedgehog/Hedgehog.fsproj
```diff
@@ -1,45 +1,11 @@
...
   <PropertyGroup>
-    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
-    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
-    <ProjectGuid>{B338B8C4-1C98-4F42-9C45-56BD7EF2AD0A}</ProjectGuid>
-    <OutputType>Library</OutputType>
-    <RootNamespace>Hedgehog</RootNamespace>
-    <AssemblyName>Hedgehog</AssemblyName>
-    <TargetFrameworkVersion>v4.5.2</TargetFrameworkVersion>
-    <TargetFSharpCoreVersion>4.4.0.0</TargetFSharpCoreVersion>
+    <TargetFramework>netstandard1.6</TargetFramework>
```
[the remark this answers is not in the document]
moodmosaic: So, I'd propose

```xml
<TargetFrameworks>netstandard1.6;net461</TargetFrameworks>
```

no?

moodmosaic:
Getting extremely close to having this merged :rocket: I left a [comment](https://github.com/hedgehogqa/fsharp-hedgehog/pull/153/files#r167123207) about targeting .NET 4.6.1 as well.

src/Hedgehog/Hedgehog.fsproj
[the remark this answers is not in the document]
Porges: Not needed; .NET 4.6.1 can use .NET Standard 1.6 assemblies (as long as you have the DotNet SDK). See [the compatibility chart here](https://docs.microsoft.com/en-us/dotnet/standard/net-standard).

paket.dependencies
Porges: I did think about this but it seemed like more hassle than needed... e.g. in the `paket.references` for the projects you have to use the same groups.

Porges:
There is one thing I want to check, which is to make sure everything works nicely with the new `FSharp.Core` setup but old-style projects.

moodmosaic:
Sure, just let me know when you're done so we can merge this.

---

From my side, I've run `dotnet build` on a fresh Ubuntu 16.04.3 and on a fresh Windows 10 and works fine.

On Windows I had to install the .NET Core SDK, and on Ubuntu I had to also update Mono:

```
# .NET Core SDK
curl https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor > microsoft.gpg
sudo mv microsoft.gpg /etc/apt/trusted.gpg.d/microsoft.gpg

sudo sh -c 'echo "deb [arch=amd64] https://packages.microsoft.com/repos/microsoft-ubuntu-xenial-prod xenial main" > /etc/apt/sources.list.d/dotnetdev.list'
sudo apt-get update

sudo apt-get install dotnet-sdk-2.1.3

# Mono
sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF
echo "deb http://download.mono-project.com/repo/ubuntu xenial main" | sudo tee /etc/apt/sources.list.d/mono-official.list
sudo apt-get update
sudo apt-get install mono-devel
```

Perhaps we can update the README at some point to mention those.

paket.dependencies
moodmosaic: Sounds reasonable. Let's leave it as-is for now.

src/Hedgehog/Hedgehog.fsproj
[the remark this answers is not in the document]
moodmosaic: OK, cool.

Porges:
Readme already mentions .NET Core SDK, but I didn't put that you need Mono on Linux (that's due to the Paket executable, right?)

moodmosaic:
Right

Porges:
I actually did find a reason to use Paket groups, so I've done that. Unquote depends on a higher version of FSharp.Core so this was actually forcing up the version for the whole project.

moodmosaic:
Cool. So when you get a chance to try [this](https://github.com/hedgehogqa/fsharp-hedgehog/pull/153#issuecomment-364324707)
>make sure everything works nicely with the new FSharp.Core setup but old-style projects

we should merge and release a NuGet package.

Porges:
Okay, so everything works nicely in new-style .fsproj for both .NET Core and .NET Framework.

However, if you're using old-style .fsproj then in this case the nuget package DLL for FSharp.Core is preferred to the one your project depends on.

I'll have a think on what we should do here. It looks like FsCheck puts the dependency in anyway, so if you install it your version can get downgraded, but you can upgrade it to the version you want after that.

moodmosaic:
So it looks like it will work from both C# and F# projects now (just that in F# it'll use the specific version of FSharp.Core from NuGet).

Porges:
Yes. If you are using old-style .fsproj and you aren't already consuming FSharp.Core via nuget then you'll get whatever version is installed when you install Hedgehog (probably 4.1.17).

moodmosaic:
This PR also resolves #108, by the way.

Porges:
Okay, I [asked for update on guidance](https://github.com/fsharp/fsharp.github.io/issues/85) and Don has [updated the FSharp.Core](https://github.com/fsharp/fsharp.github.io/pull/86/files) info. The guidance now is "always reference FSharp.Core via the nuget package".

moodmosaic:
So we're good to go and merge this, right?

Porges:
🚀

PR #160, opened 2018-05-28
Sync F# version of Seed with Haskell

moodmosaic:
* Use 64-bit unsigned integers as in https://github.com/hedgehogqa/haskell-hedgehog/pull/185
* Avoid weak gamma-γ values as in https://github.com/hedgehogqa/haskell-hedgehog/pull/198

/cc @jystic


jystic:
:100: from me, merge away

PR #161, opened 2018-05-30
Verify that Seed.from 'fixes' the γ-value

moodmosaic:
As described in https://github.com/hedgehogqa/haskell-hedgehog/commit/39b15b9b4d147f6001984c4b7edab00878269da7


moodmosaic:
/cc @jystic

We can actually have a similar group of tests also for the Haskell version, just to make sure we'll never break things accidentally :sweat_smile:

jystic:
:+1:

PR #167, opened 2018-06-30
fix paths of source files in the script.

jwChung:


jwChung:
> This must have been a breaking change of Paket(?)

I guess that the code in the script file was written before declaring the `Build` group in the paket.dependencies file.  Declaring the group tirggered to create the new `build` directory in the packages directory(`/packages/build/...`).

moodmosaic:
I'll merge when I'm at my laptop. Thank you for discovering and fixing this.

PR #168, opened 2018-06-30
Improve paket dependencies

### LINKED ISSUE #166: Use the global nuget packages
Since the new project format is used, the global nuget packages are referenced as default settings. As   the `storage: none` line is set on the top in the `paket.dependencies` file, the line is not applied to each group - Build, Test.

I don't know it is intentional.  Can the line be set for each group?

jwChung:
This closes #166.


paket.dependencies
```diff
@@ -7,6 +7,8 @@ storage: none
 
 group Build
 
+    framework: netstandard1.6
```
jwChung: This prevent to generate dependencies for other .NET Framework versions than the ones we projects use. You can see the chages of the paket.lock file below.

paket.dependencies
```diff
@@ -7,6 +7,8 @@ storage: none
...
+    storage: none
```
jwChung: It is to use the global nuget directory. The packages in the root directory is going to be unnecessary.

moodmosaic: We aren't checking in the packages in the root directory so I think this change makes sense. After all, I think the new *.fsproj file format (and .NET Core) uses the global/cache NuGet directory by default(?)

paket.dependencies
```diff
@@ -33,6 +35,8 @@ group Build
 
 group Test
 
+    framework: netcore2.0
```
moodmosaic: We need `netcore2.0` here to satisfy the dependencies like xUnit.net and Unquote, is that right? 😃

moodmosaic: What happens if we make this `netstandard2.0`, why not even `1.6`? (Just thinking out loud.)

moodmosaic:
💯 👍 This seems to be moving to the right direction, as the `paket.lock` file seems much cleaner. I left a few comments as I'm trying to fully understand each one of those changes.

paket.dependencies
jwChung: > I think the new *.fsproj file format (and .NET Core) uses the global/cache NuGet directory by default(?)

Yes, the `PackageReference` tag in the  new *.fsproj file is declared to use the global/cache NuGet directory by default.

moodmosaic: Yes, it's all written in [here](https://docs.microsoft.com/en-us/nuget/consume-packages/managing-the-global-packages-and-cache-folders), thank you 😃

>The *global-packages* folder is where NuGet installs any downloaded package. [..] Projects using the PackageReference format always use packages directly from this folder.

paket.dependencies
jwChung: The framework restriction is corresponding to frameworks of projects in a solution. If the line is changed to netstandard2.0 or 1.6, all the nuget packages in the Test group cannot be referenced so compile errors will be occurred.

That line is like saying that only  projects under .net core 2.0 are supported. If we need other frameworks side by side, we can declare simply like,

```
framework: netcore2.0, netstandard2.0, netstandard1.6
```

moodmosaic: So perhaps `netcore1.6` is what we need? (Instead of `netcore2.0`.)

Porges: This is correct for the test projects, they should target a concrete framework, not standard 👍

Porges:
It's strange that `storage: none` at the root doesn't affect all the groups. LGTM! 👍

paket.dependencies
moodmosaic: @Porges, sounds good. I'm now looking again at https://github.com/Microsoft/visualfsharp/issues/3680#issuecomment-363698804 😃

moodmosaic:
>It's strange that `storage: none` at the root doesn't affect all the groups. LGTM! 👍 

Could be a Paket issue?

moodmosaic:
Manually rebased and `git merge --no-ff`'ed from the command line and into `master` via 435349d74a27e3d0b206f08f46cfd72621520e95.

moodmosaic:
Thank you for your contribution, @jwChung 💯 🥇 🚀 🎊

PR #173, opened 2018-07-27
fix for generating out-of-range values,

jwChung:
This commit addresses that generating uint32 and uint64 returns
out-of-range values. This problem occurs in numeric conversions.

### DIFF
--- a/tests/Hedgehog.Tests/GenTests.fs
+++ b/tests/Hedgehog.Tests/GenTests.fs
@@ -1,5 +1,6 @@
 ﻿module Hedgehog.Tests.GenTests
 
+open System
 open Hedgehog
 open Swensen.Unquote
 open Xunit
@@ -63,3 +64,15 @@ let ``dateTime shrinks to correct mid-value`` () =
         |> Array.item 1
         |> System.DateTime.Parse
     System.DateTime (2000, 1, 1) =! result
+
+[<Fact>]
+let ``uint64 doesn't return any out-of-range value`` () =
+    let gen = Gen.uint64 <| Range.constant 1UL UInt64.MaxValue
+    let actual = Gen.sample 0 100 gen
+    test <@ actual |> List.contains 0UL |> not @>
+
+[<Fact>]
+let ``uint32 doesn't return any out-of-range value`` () =
+    let gen = Gen.uint32 <| Range.constant 1ul UInt32.MaxValue
+    let actual = Gen.sample 0 100 gen
+    test <@ actual |> List.contains 0ul |> not @>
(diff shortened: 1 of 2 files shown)


moodmosaic:
@jwChung Good catch! See also this [recent version of the original Numeric.fs]. They do the same there, and keep the old method with the cast for backwards compatibility with .NET 3.x.

💯

tests/Hedgehog.Tests/GenTests.fs
```diff
@@ -62,4 +63,16 @@ let ``dateTime shrinks to correct mid-value`` () =
...
         |> System.DateTime.Parse
-    System.DateTime (2000, 1, 1) =! result
\ No newline at end of file
+    System.DateTime (2000, 1, 1) =! result
+
+[<Fact>]
+let ``uint64 doesn't return any out-of-range value`` () =
+    let gen = Gen.uint64 <| Range.constant 1UL UInt64.MaxValue
+    let actual = Gen.sample 0 100 gen
+    Assertions.test <@ actual |> List.contains 0UL |> not @>
```
moodmosaic: Do we need to qualify the module `Assertions`? `test <@ actual …` should work fine.

tests/Hedgehog.Tests/GenTests.fs
```diff
@@ -62,4 +63,16 @@ let ``dateTime shrinks to correct mid-value`` () =
...
+
+[<Fact>]
+let ``uint32 doesn't return any out-of-range value`` () =
+    let gen = Gen.uint32 <| Range.constant 1ul UInt32.MaxValue
+    let actual = Gen.sample 0 100 gen
+    Assertions.test <@ actual |> List.contains 0ul |> not @>
```
moodmosaic: Do we need to qualify the module `Assertions`? `test <@ actual …` should work fine.

moodmosaic:
I left two μ-nitpicks, but should be OK to be merged even if you don't want to address those 👍

tests/Hedgehog.Tests/GenTests.fs
jwChung: Aha, I didn't know that it was automatically opened by `AutoOpenAttribute`.

jwChung:
According to the original Numeric.fs, it seems that there are the same problems with other types - `sbyte`, `int16`, `byte` and  `uint16` especially `float32` and `decimal`. If I encounter the problem, I'll open an issue or PRs.

moodmosaic:
Perhaps we can also update Numeric.fs with the latest one.

moodmosaic:
Thank you, @jwChung 💯

Issue #172, opened 2018-07-16
Many functions that take FSharpFunc arguments are very difficult to use from C#


Porges:
As in title, this needs to be changed to use `Func` instead.

moodmosaic:
I think there are two options here, either we add a `System.Func<T, TResult>` annotation, or add a façade to expose all the 'incompatible' functions to the rest of .NET in a friendlier way.

Porges:
Also, `Gen.Delay`

moodmosaic:
Or put them on a different module to avoid adding the `Func` suffix? I'm sure I've seen this elsewhere but can't really remember where 😯

Porges:
That would mean you need to use `Gen` in some places and `Gen2` in others?

moodmosaic:
No, that would be messy. Perhaps what you suggest in https://github.com/hedgehogqa/fsharp-hedgehog/issues/172#issuecomment-405812551 is the best we can do.

PR #183, opened 2019-03-09
introduce the Gen static class

jwChung:
This PR addresses #172. To expose C# friendly members, the Gen static class is introduced. This causes renaming the gen module from 'Gen' to 'GenModule'. We can use the new Gen static class in C# without F# type disturbance. One big drawback is that we should re-declare all the gen members.

In a short, frequent and repetitive way, I has re-declared only some members. 

/cc @Porges

### DIFF
--- a/src/Hedgehog/Gen.fs
+++ b/src/Hedgehog/Gen.fs
@@ -32,7 +32,6 @@ module Gen =
     let create (shrink : 'a -> LazyList<'a>) (random : Random<'a>) : Gen<'a> =
         Random.map (Tree.unfold id shrink) random |> ofRandom
 
-    [<CompiledName("Constant")>]
     let constant (x : 'a) : Gen<'a> =
         Tree.singleton x |> Random.constant |> ofRandom
 
@@ -64,7 +63,6 @@ module Gen =
     let mapTree (f : Tree<'a> -> Tree<'b>) (g : Gen<'a>) : Gen<'b> =
         mapRandom (Random.map f) g
 
-    [<CompiledName("Map")>]
     let map (f : 'a -> 'b) (g : Gen<'a>) : Gen<'b> =
         mapTree (Tree.map f) g
 
@@ -550,7 +548,12 @@ module GenOperators =
     let (<!>) = Gen.map
     let (<*>) = Gen.apply
 
+open System
+
 /// In C# friendly manner, a generator for values and shrink trees of type 'a
 type Gen =
     static member Bool = Gen.bool
     static member Int32(range : Range<int>) : Gen<int> = Gen.int range
+    static member Constant(x : 'a) : Gen<'a> = Gen.constant x
+    static member Map(f : Func<'a, 'b>) (g : Gen<'a>) : Gen<'b> =
+        Gen.map f.Invoke g
--- a/src/Hedgehog/Gen.fs
+++ b/src/Hedgehog/Gen.fs
@@ -398,7 +398,7 @@ module Gen =
     /// ('\000'..'\65535' excluding '\55296'..'\57343').
     [<CompiledName("Unicode")>]
     let unicode : Gen<char> =
-        let isNoncharacter x = 
+        let isNoncharacter x =
                x = Operators.char 65534
             || x = Operators.char 65535
         unicodeAll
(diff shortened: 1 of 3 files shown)


src/Hedgehog/Gen.fs
```diff
@@ -551,3 +547,13 @@ module GenBuilder =
...
     let (<*>) = Gen.apply
+
+open System
+
+/// In C# friendly manner, a generator for values and shrink trees of type 'a
+type Gen =
+    static member Bool = Gen.bool
+    static member Int32(range : Range<int>) : Gen<int> = Gen.int range
+    static member Constant(x : 'a) : Gen<'a> = Gen.constant x
+    static member Map(f : Func<'a, 'b>) (g : Gen<'a>) : Gen<'b> =
```
moodmosaic: Don't we have a [`Select`](https://github.com/hedgehogqa/fsharp-hedgehog/blob/ed75041fe59082153a0a659128c41cfabee2db10/src/Hedgehog/LinqSupport.fs#L19) already? C# users are more familiar with that instead of `Map`.

src/Hedgehog/Gen.fs
```diff
@@ -551,3 +547,13 @@ module GenBuilder =
 module GenOperators =
     let (<!>) = Gen.map
     let (<*>) = Gen.apply
+
+open System
+
+/// In C# friendly manner, a generator for values and shrink trees of type 'a
+type Gen =
+    static member Bool = Gen.bool
```
moodmosaic: The issue here is only for those generators taking `FSharpFunc` for example `Gen.delay`. I don't think we need to carry on each and every generator.

src/Hedgehog/Gen.fs
```diff
@@ -551,3 +547,13 @@ module GenBuilder =
 module GenOperators =
     let (<!>) = Gen.map
     let (<*>) = Gen.apply
+
+open System
+
+/// In C# friendly manner, a generator for values and shrink trees of type 'a
+type Gen =
```
moodmosaic: Unsure if we really need a new `Gen` type since there's the existing [`GenLinqSupport`](https://github.com/hedgehogqa/fsharp-hedgehog/blob/ed75041fe59082153a0a659128c41cfabee2db10/src/Hedgehog/LinqSupport.fs) by @Porges.

src/Hedgehog/Gen.fs
jwChung: I moved it from the previous code and it is better to remove it.

jwChung:
@moodmosaic Thank you for your comments.

> Being more C#-friendly means even broader reach. 

I agree. I should remove the term 'C#-friendly'.

> I'm a little bit concerned about adding another Gen type since we do actually have one already for supporting LINQ. — So, the simplest solution I can think of is to

From C# user perspective, C# users probably do not want to use the two Gen types. - the Gen module and the one `CSharpSupport`(?) you suggest. This is the reason why I redeclare the Gen members - `Bool` and `Int32`.  I think this is no big deal.

moodmosaic:
>I agree. I should remove the term 'C#-friendly'.

No, we *do* want C# folks here :heart: Sorry if I wasn't clear.

moodmosaic:
>From C# user perspective, C# users probably do not want to use the two Gen types

Hmm... Not sure what you mean. You might want to look first in the LINQ examples [here](https://github.com/hedgehogqa/fsharp-hedgehog/blob/master/tests/Hedgehog.CSharp.Tests/LinqTests.cs).

jwChung:
> Hmm... Not sure what you mean. You might want to look first in the LINQ examples here.

Sorry for being unclear. If the members of `Gen` in C# are used, I would like to show a list of the available members by just typing '.' after `Gen`, not using both `Gen` and `CSharpSupport`.

moodmosaic:
@jwChung, I think we should close this :confused: I think we're going off on a tangent if we have to declare each generator twice as in https://github.com/hedgehogqa/fsharp-hedgehog/pull/183#discussion_r263988161.

Then there's the existing `GenLinqSupport` as in https://github.com/hedgehogqa/fsharp-hedgehog/pull/183#discussion_r263988189. Perhaps we should rename and use that type for any additional generators that we need to make C#-friendly.

Perhaps we should start with `Gen.delay`, and see where it's the best place to add it. Unsure if it's worth keeping this pull request open though, so I'm just going to close it, feel free to reopen if necessary.

As always, thank you for all the great work you've been putting in. :+1:

jwChung:
@moodmosaic  I was not persuasive for this pull request. I agree to close.

Issue #185, opened 2019-04-02
Gen.int64 (Range.ExponentialBounded()) causes OverflowException


cmeeren:
```f#
[<Fact>]
let ``can create exponentially bounded int64`` () =
    Property.check <| property {
        let! _ = Gen.int64 (Range.exponentialBounded ())
        return true
    }
```

Causes the following exception:

`Message: System.OverflowException : Value was either too large or too small for an Int64.`

cmeeren:
Any chance of fixing this?

moodmosaic:
I may be able to look into it next week (traveling at the moment). Thanks for the heads up☝

cmeeren:
It seems this also happens with `uint64`: https://github.com/cmeeren/fsharp-hedgehog-experimental/issues/32#issuecomment-505968571

moodmosaic:
Fixed in https://github.com/hedgehogqa/fsharp-hedgehog/pull/188, released as Hedgehog 0.8.3 :tada:

PR #186, opened 2019-06-27
Proposed fix for issue #185

marklam:
https://github.com/hedgehogqa/fsharp-hedgehog/issues/185

The problem looks to be that the clamp on exponential ranges is performed after the conversion to the target type, so if the value falls just outside the range then an exception is thrown before the clamp occurs.

This moves the clamp to before the conversion, but maybe a better fix would be to sort out the calculation that's going out of bounds in the first place.

### DIFF
--- a/src/Hedgehog/Range.fs
+++ b/src/Hedgehog/Range.fs
@@ -135,9 +135,8 @@ module Range =
             let diff =
                  (((float (abs (n - z) + 1I)) ** (float sz / 99.0)) - 1.0) * float (sign (n - z))
 
-            bigint (round (float z + diff))
-            |> clamp (toBigInt lo) (toBigInt hi)
-            |> fromBigInt
+            // https://github.com/hedgehogqa/fsharp-hedgehog/issues/185
+            fromBigInt (clamp (toBigInt lo) (toBigInt hi) (bigint (round (float z + diff))))
 
     /// Construct a range which scales the bounds relative to the size
     /// parameter.
@@ -175,10 +174,13 @@ module Range =
     [<CompiledName("ExponentialFrom")>]
     let inline exponentialFrom (z : 'a) (x : 'a) (y : 'a) : Range<'a> =
         Range (z, fun sz ->
+            let scale =
+                // https://github.com/hedgehogqa/fsharp-hedgehog/issues/185
+                scaleExponential x y sz z
             let x_sized =
-                scaleExponential x y sz z x
+                scale x
             let y_sized =
-                scaleExponential x y sz z y
+                scale y
             x_sized, y_sized)
 
     /// Construct a range which scales the second bound exponentially relative
--- a/src/Hedgehog/Range.fs
+++ b/src/Hedgehog/Range.fs
@@ -122,7 +122,7 @@ module Range =
             fromBigInt (z + diff)
 
         /// Scale an integral exponentially with the size parameter.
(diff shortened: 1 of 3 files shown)


src/Hedgehog/Range.fs
```diff
@@ -174,9 +176,9 @@ module Range =
     let inline exponentialFrom (z : 'a) (x : 'a) (y : 'a) : Range<'a> =
         Range (z, fun sz ->
             let x_sized =
-                clamp x y (scaleExponential sz z x)
+                scaleExponential x y sz z x
             let y_sized =
-                clamp x y (scaleExponential sz z y)
+                scaleExponential x y sz z y
```
moodmosaic: If we can avoid passing `x` twice, we should be able to take this as-is for now.

src/Hedgehog/Range.fs
```diff
@@ -135,7 +135,9 @@ module Range =
             let diff =
                  (((float (abs (n - z) + 1I)) ** (float sz / 99.0)) - 1.0) * float (sign (n - z))
 
-            fromBigInt (bigint (round (float z + diff)))
+            bigint (round (float z + diff))
```
moodmosaic: `diff` and `round` might also be problematic

tests/Hedgehog.Tests/GenTests.fs
```diff
@@ -76,3 +76,17 @@ let ``uint32 doesn't return any out-of-range value`` () =
...
     let actual = Gen.sample 0 100 gen
     test <@ actual |> List.contains 0ul |> not @>
+
+[<Fact>]
+let ``can create exponentially bounded int64`` () =
+    Property.check <| property {
+        let! _ = Gen.int64 (Range.exponentialBounded ())
+        return true
+    }
+
```
moodmosaic: Those are fine :100:

src/Hedgehog/Range.fs
marklam: Since one call has only x repeated as the last parameter, and the other has only y repeated, I'm not sure how that could be done (other than partially applying the function)

src/Hedgehog/Range.fs
```diff
@@ -122,7 +122,7 @@ module Range =
             fromBigInt (z + diff)
 
         /// Scale an integral exponentially with the size parameter.
-        let inline scaleExponential (sz0 : Size) (z0 : 'a) (n0 : 'a) : 'a =
+        let inline scaleExponential (lo : 'a) (hi : 'a) (sz0 : Size) (z0 : 'a) (n0 : 'a) : 'a =
```
moodmosaic: @jacobstanley, I don't like this change, but other alternatives I've tried make the F# compiler infer `'a` as `bigint` and breaks everything. Since `scaleExponential` is `internal` I'll go ahead and merge this.

Right now exponential 64-bit un/signed integers can cause arithmetic overflows, and break upstream packages, so I'll go ahead and proceed with this fix.

moodmosaic:
Thank you for all the help @marklam and @cmeeren. A new release will be up on NuGet Gallery soon.

moodmosaic:
@marklam, do you have a Twitter account? Usually, when doing a new release, we announce it on Twitter and give credit to all contributors that helped get the release out.

marklam:
Thanks @moodmosaic, I'm @_marklam_ on twitter.

Issue #182, opened 2019-02-23
Merge stuff from Hedgehog.Experimental?


cmeeren:
Hedgehog.Experimental has existed in relatively stable form for quite a while now. Could you consider merging some functionality into the core Hedgehog library? For example, the auto-generation of arbitrary types is really nice to have.

Note that when I say "stable" I just mean "hasn't changed in a while". I have no idea how many people actually use Hedgehog.Experimental.

moodmosaic:
>[Hedgehog.Experimental](https://github.com/cmeeren/fsharp-hedgehog-experimental/) has existed in relatively stable form for quite a while now.

:heart: :+1: 

>Could you consider merging some functionality into the core Hedgehog library?

What would be the reason of having Hedgehog.Experimental then?

cmeeren:
"Experimental" doesn't sound safe to use. Having some core, important features such as auto-generation part of the main Hedgehog package could make it more attractive.

Also, as I understood [your comment here](https://github.com/hedgehogqa/fsharp-hedgehog/issues/139#issuecomment-339833927), this was the intent from the beginning: For Hedgehog.Experimental to be a proving ground for new features with the intent to promote them to the main Hedgehog library after a while.

At the very least, I think Hedgehog.Experimental should be clearly linked to from this repo's readme and/or tutorial. There is absolutely no mention of Hedgehog.Experimental or "auto" (e.g. auto-generation) anywhere in this repo, so potential Hedgehog users have no way of knowing that Hedgehog actually supports e.g. auto-generation except for searching the issues or stumbling across Hedgehog.Experimental themselves.

moodmosaic:
>I think Hedgehog.Experimental should be clearly linked to from this repo's readme and/or tutorial

Absolutely. :+1: 

There's https://github.com/hedgehogqa/fsharp-hedgehog/issues/138 for this. Feel free to make a pull request, otherwise I may do so (currently I'm travelling so it might take a while before I get to it).

cmeeren:
Sounds good, I'm in no immediate hurry. 👍

moodmosaic:
>Having some core, important features such as auto-generation part of the main Hedgehog package could make it more attractive to users.

I'm worried that if we add auto-generation inside the core library, people coming from QC might start using just that and start to forget about the Range type and combinators—[an important quality of Hedgehog](https://youtu.be/AIv_9T0xKEo?t=721).

---

[...]

PR #187, opened 2019-07-01
Mention fsharp-hedgehog-experimental in README

moodmosaic:
ref.: #182 
c/c: @cmeeren, @jacobstanley

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,8 @@ You can then load the module in F# Interactive, and run it:
 
 More examples can be found in the [tutorial](doc/tutorial.md).
 
+👉 For auto-generators (à la AutoFixture) and other convenience generators, check out [fsharp-hedgehog-experimental](https://github.com/cmeeren/fsharp-hedgehog-experimental/).
+
 ## Building from source
 
 To build Hedgehog from source, you will need either the


README.md
```diff
@@ -44,6 +44,8 @@ You can then load the module in F# Interactive, and run it:
 
 More examples can be found in the [tutorial](doc/tutorial.md).
 
+👉 For auto-generators (à la AutoFixture) check out [fsharp-hedgehog-experimental](https://github.com/cmeeren/fsharp-hedgehog-experimental/).
```
moodmosaic: AutoFixture is so popular that I think there's no reason to add a link to it. :sweat_smile:

cmeeren: Since the package contains much more than just auto-generators, how about "for more generators, including auto-generators, ..." or something like that?

moodmosaic: I'd like to emphasize auto-generators, so we'd rather do it vice-versa "For auto-generators (à la AutoFixture), and other, extra, generators check out ...". Sounds good?

cmeeren: Sounds good. If you want, you can say "and other convenience generators".

moodmosaic: :+1:

moodmosaic:
Preview:

image

@jacobstanley, I'll merge this tomorrow.

PR #195, opened 2020-04-20
exclude obj directories from nuget package

ThisFunctionalTom:
Hi,

as you already noticed we have some generated files from obj folder also packed in the nuget package into fable folder.

I do not think fable compiler needs this.

I just added to ignore all files from obj folder when generating nuget package.

Hope this is OK.

### DIFF
--- a/src/Hedgehog/paket.template
+++ b/src/Hedgehog/paket.template
@@ -23,3 +23,4 @@ tags
 files
     *.fsproj ==> fable/
     **/*.fs ==> fable/
+    !obj/**/*.*


moodmosaic:
Looks better now, @ThisFunctionalTom :tada: Here's what goes in the NuGet Package (not released yet):

```
$ du -a .
20      ./fable/Gen.fs
4       ./fable/Hedgehog.fsproj
4       ./fable/LinqSupport.fs
12      ./fable/Numeric.fs
16      ./fable/Property.fs
4       ./fable/Random.fs
8       ./fable/Range.fs
8       ./fable/Seed.fs
8       ./fable/Shrink.fs
4       ./fable/Tree.fs
92      ./fable
4       ./Hedgehog.nuspec
296     ./lib/netstandard1.6/Hedgehog.dll
24      ./lib/netstandard1.6/Hedgehog.XML
320     ./lib/netstandard1.6
320     ./lib
4       ./package/services/metadata/core-properties/coreProp.psmdcp
4       ./package/services/metadata/core-properties
4       ./package/services/metadata
4       ./package/services
4       ./package
4       ./[Content_Types].xml
1       ./_rels/.rels
1       ./_rels
429     .
```

I guess, what's next is to see whether Fable is happy with what's in the newly created NuGet Package and if so, we can do a release.

PR #196, opened 2020-04-25
Fable tests

ThisFunctionalTom:
Hi,

I just converted all of the Hedgehog tests from xunit to Fable.Mocha to test if everything is working OK. 

There a 3 tests failing with Range.exponentialBounded. I put them in "pending" mode.
Otherwise it looks good.

Two tests are taking a lot of time "unicode doesn't return any surrogate" and "unicode doesn't return any noncharacter" because they generate 100000 values but they are green.
If you would like to speed the tests you can put them in pending state by changing the fact to pfact and theory to ptheory.

To run tests with nodjs use `npm test`
To run live tests in browser use `npm start` and then open http://localhost:8085

I also added expecto as the runner for Hedgehog.Fable.Tests. To run it you have to compile the Hedghog.Fable.Tests project with msbuild (dotnet build is not working for the reason you already know). After the build you can run Hedgehog.Fable.Tests like a normal console application and expecto will run the tests. There is also a npm run script `npm run dotnet-expecto` but you need to have msbuild in your path to use it.

I would also like to create some sample with Fable tests but I guess I will put it in another pull requests.

kind regards,
Tom


moodmosaic:
Awesome :heart_eyes: I'll look into it during the next few days

ThisFunctionalTom:
I also converted Hedgehog.Fable.Tests.fsproj to use paket and updated paket.exe to the newest version. Hope this is OK.

inosik:
> dotnet build is not working for the reason you already know

Do you run the latest VS 2019? The issue (failing do read `FSharp.Core.dll`) should be fixed in 16.5.3 and later.

inosik:
[SDK 3.1.201 was reported to be good again](https://github.com/dotnet/fsharp/issues/8739#issuecomment-610835284), but I can't build with it either.

ThisFunctionalTom:
> 
> 
> [SDK 3.1.201 was reported to be good again](https://github.com/dotnet/fsharp/issues/8739#issuecomment-610835284), but I can't build with it either.

Yeah, I had the same problems.

Following is working for me:

In the root directory after fresh clone or after `git clean -xdf`

For running expecto tests:
```
dotnet restore
msbuild
dotnet run -p .\tests\Hedgehog.Fable.Tests\Hedgehog.Fable.Tests.fsproj
```
I am not sure why .paket\paket.exe restore is not working here but `dotnet restore` is working for me.


For running fable/mocha tests in nodejs (You have to wait a little bit, there are some long running tests)
```
cd .\tests\Hedgehog.Fable.Tests\
npm install
npm test
```

For running fable/mocha tests in browser (Here I have to click in browser to wait for the application because of the long running tests).
After `npm start` open `http://localhost:8085` and wait.
```
cd .\tests\Hedgehog.Fable.Tests\
npm install
npm start
```

ThisFunctionalTom:
> 
> 
> This is so worth it 😲 🚀 Thank you, thank you, _thank you_, for all the great work.

No problem. This is fun.

 — If we'd use Expecto from start, now we would avoid all the duplication, right?

Yes. If we write everything in expecto style we can use it for Fable.Tests. The thing that does not work in Fable.Mocha is it does not support property based tests like Expecto but we are going to change it :-)

> 
> If _that's_ the case, shouldn't we consider replacing the existing (xUnit.net) test-suite with just a single Expecto one?
> 
> As I see it, there's a huge benefit here: one test-suite for both native (F#) and compiled (JS headless and browser) targets.

Yes. This is the idea to write one test for both worlds. That is excatly why Zaid Ajaj made Fable.Mocha library to have the same interface as Expecto.

moodmosaic:
>The thing that does not work in Fable.Mocha is it does not support property based tests like Expecto

This shouldn't be a problem as most (if not all) of the test-suite is xUnit.net-based.

moodmosaic:
:ping_pong:

ThisFunctionalTom:
> 
> 
> 🏓

Sorry, is it expected for me to do something here? I am not sure how this works :-)

moodmosaic:
It's just a 'ping' (mostly for me) to get back to this anytime soon. Sorry for any trouble caused.

ThisFunctionalTom:
> 
> 
> It's just a 'ping' (mostly for me) to get back to this anytime soon. Sorry for any trouble caused.

No problem. I just didn't know if I had to do something so I asked.

moodmosaic:
Marking this pull request as draft now. The plan is to try to remove the duplication in the test code.

adam-becker:
I would love to get this merged. For some reason ionide gets stuck in a tight loop in the presence of double back-ticks, and makes my computer sound like it's going to achieve flight.

ThisFunctionalTom:
> @moodmosaic I would love to get this merged. For some reason ionide gets stuck in a tight loop in the presence of double back-ticks, and makes my computer sound like it's going to achieve flight. Given that all of the xunit tests have double back-ticks everywhere this is a problem for me.
> 
> @ThisFunctionalTom are you able to rebase these changes?

Did it.

Also, we could try to upgrade fable compiler to fable 3. Fable 3 is a dotnet tool and we do not need npm anymore 🥳

moodmosaic:
@ThisFunctionalTom, thanks for rebasing. :+1:

I haven't followed Fable (3) but if that means we can run everything without having to install Node.js (and NPM) then perhaps it'll make things simpler.

@adam-becker, off the top of my head, we'd have to check the following before even consider merging:

1. The tests in this PR have all the newly added ones (for example those added in #228)
2. GitHub Actions will pick up all the tests during build(s)

ThisFunctionalTom:
> @ThisFunctionalTom, thanks for rebasing. 👍
> 
> I haven't followed Fable (3) but if that means we can run everything without having to install Node.js (and NPM) then perhaps it'll make things simpler.

I don't think we can escape Node.js 😢  because we have to run the tests with mocha test runner and it comes as npm package. 

> 
> @adam-becker, off the top of my head, we'd have to check the following before even consider merging:
> 
> 1. The tests in this PR have all the newly added ones (for example those added in #228)
> 2. GitHub Actions will pick up all the tests during build(s)

I have to check this. I will try to do it this week.

adam-becker:
> I don't think we can escape Node.js :cry: because we have to run the tests with mocha test runner and it comes as npm package.

Isn't the point of these tests that it's flexible? According to the documentation, we should be able to run with [expecto][fable-expecto].

[fable-expecto]: https://github.com/Zaid-Ajaj/Fable.Mocha#running-the-tests-on-dotnet-with-Expecto

ThisFunctionalTom:
@adam-becker The problem is that the underlying

> > I don't think we can escape Node.js 😢 because we have to run the tests with mocha test runner and it comes as npm package.
> 
> @ThisFunctionalTom Isn't the point of these tests that it's flexible? According to the documentation, we should be able to run with [expecto](https://github.com/Zaid-Ajaj/Fable.Mocha#running-the-tests-on-dotnet-with-Expecto). At least in a CI setting, whatever a developer chooses to do with their local environment is up to them.

Running the tests just with expecto has nothing to do with fable compiler. It is just normal F# compiler + .NET runtime.
The Fable compiler generates javascript and we should test if the generated javascript code is correct. To do it we need nodejs and mocha (unit test runner).
The new fable compiler is a .NET tool but it doesnt help us because we need to test the generated code. At least I think so.

adam-becker:
> Running the tests just with expecto has nothing to do with fable compiler. It is just normal F# compiler + .NET runtime.
> The Fable compiler generates javascript and we should test if the generated javascript code is correct. To do it we need nodejs and mocha (unit test runner).
> The new fable compiler is a .NET tool but it doesnt help us because we need to test the generated code. At least I think so.

@ThisFunctionalTom I just meant from a CI perspective, but you're right that we would still want to exercise the Fable generated code in case any bugs crop up there. As long as the project doesn't ***require*** nodejs for development I think it's still fine.

@moodmosaic I'm unsure how much you're allowed customize the build environment with github actions, but if it's anything like GitLab CI with docker then this should be trivial to support.

ThisFunctionalTom:
Hi, sorry for the long delay.

I made following changes:

- rebased again on master
- copied changes made in Hedgehog.Tests to Hedgehog.Fable.Tests
- copied and converted new TreeTests to Hedgehog.Fable.Tests
- upgraded fable compiler to fable3 (added fable3 compiler dotnet tool)

npm is still needed for running tests in node/browser runtimes.

If you would like to merge this we should probably think/talk about exchanging Hedgehog.Tests with Hedghog.Fable.Tests because they are duplicates now. If we run Hedghog.Fable.Tests with `dotnet run` they are running in .NET runtime with expecto and should actually be the same tests as Hedgehog.Tests run with `dotnet test` (xunit runner). It would be really cool if we could do this because we would not have to maintain two separate projects with completely same tests.

To try the same tests in javascript you can:

- `cd` into  Hedgehog.Fable.Tests directory
- `npm install` -> download/install javascript dependencies
- `npm run build` -> build all tests with fable 3
- `npm run test` -> run the tests in node environment
- `npm run start` and then open http://localhost:8085/ and wait -> run the tests in browser environment

src/Hedgehog/Tuple.fs
```diff
@@ -9,7 +9,7 @@ module private Tuple =
         x, f y
 
 
-module private GenTuple =
+module GenTuple =
```
adam-becker, line 12: Did this change so tests can be written?

adam-becker (changes requested):
Great work though @ThisFunctionalTom! I'm in favor of these changes, just a few things I think we should address.

- `Hedgehog.Tests` should be supplanted by these changes. In other words, delete that project, and rename `Hedgehog.Fable.Tests` to just `Hedgehog.Tests`.
- Since `Xunit` is no longer a dependency, the functions in `TestHelpers.fs` should be updated to remove `Xunit` terminology. However, I would prefer those helpers being inlined and removed.
- The `theory` tests use `<|` to pass a lambda as the last argument, we removed usages of this operator in #276, so we should consider not reintroducing it here. I won't block on this point however.
- These tests need to run as part of CI, so changing `master.yml` and `pull_request.yml` to include these tests would be my only real requirement.

src/Hedgehog/Tuple.fs
moodmosaic: Yep, see https://github.com/hedgehogqa/fsharp-hedgehog/pull/196#issuecomment-765893234.

adam-becker: I'd be curious if `private` could be changed to `internal` to still avoid exposing this module for now.

TysonMN: Or `internal` and combined with the [`InternalsVisibleTo`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.internalsvisibletoattribute) attribute.

ThisFunctionalTom:
OK. I think it is somewhat better now.

I made following changes:

- replaced Hedgehog.Tests with Hedghog.Fable.Tests
- renamed xunit terminology and refactored the TestHelper a little bit. I didn't removed TestHelpers because then we need to have `#if FABLE_COMPILER` in each file. If this is what you want I will do it.
- I still find <| for testCases more readable because you do not have to count the parenthesis, just use indentation. In a function itself I don't like `<|` operator but on the top level for `testCase` and `testCases` definition I find it OK. What do you think?
- Fable tests are now running in github actions
- Added nuget.config to the root of the project because I had issue that `dotnet restore` did not restore packages and github actions failed. [See here for more information](https://github.com/actions/setup-dotnet/issues/155#issuecomment-761195782)

tests/Hedgehog.Tests/MinimalTests.fs
```diff
@@ -65,11 +63,10 @@ let rec genExp : Gen<Exp> =
         ]
 
         Gen.choiceRec recs nonrecs
-        |> Gen.shrink shrinkExp // comment this out to see the property fail
+        |> Gen.shrink shrinkExp
     )
 
-[<Fact>]
-let ``greedy traversal with a predicate yields the perfect minimal shrink``() =
+let ``greedy traversal with a predicate yields the perfect minimal shrink`` () =
```
adam-becker, line 69: Since we're exposing this for `Hedgehog.Benchmarks`, I'd prefer a simple name without double backticks.

tests/Hedgehog.Tests/Program.fs
```diff
@@ -0,0 +1,26 @@
+module Hedgehog.Tests.Main
+
+open Hedgehog
+
+#if FABLE_COMPILER
+open Fable.Mocha
+
+let allTests = testList "All tests" [
```
adam-becker, line 8: I'd be okay with single-sourcing this (ie, not making it `FABLE_COMPILER` specific), then it'll require us to keep it in sync. Then a helper can be exposed in `TestHelpers` to keep this compiler switching stuff in one place.

tests/Hedgehog.Tests/tests-js/index.html
```diff
@@ -0,0 +1,9 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <title>Hedgehog.Fable tests</title>
```
adam-becker, line 4: This name should be updated. Also this appears to be a duplicate of `tests/Hedgehog.Tests/index.html`, are both versions required?

tests/Hedgehog.Tests/Main.fs
```diff
@@ -0,0 +1,6 @@
+module test_expecto
+open Expecto
+
+[<EntryPoint>]
+let main argv =
+    Tests.runTestsInAssembly defaultConfig argv
```
adam-becker, line 6: Do we want this file or `tests/Hedgehog.Tests/Program.fs`?

adam-becker (changes requested):
@ThisFunctionalTom This is looking much better! Thanks again for doing all of this. I had a few more comments, mostly to clarify what's being presented in this PR. The rest looks good, and I'd be comfortable with approving once my remaining comments are addressed.

tests/Hedgehog.Tests/MinimalTests.fs
ThisFunctionalTom: I renamed it to perfectMinimalSrink.

tests/Hedgehog.Tests/Program.fs
ThisFunctionalTom: Done.

tests/Hedgehog.Tests/tests-js/index.html
ThisFunctionalTom: Renamed to Hedgehog.Tests.
Oh, tests-js should not be commited to git 😅. Removed and gitignored it.

tests/Hedgehog.Tests/Main.fs
ThisFunctionalTom: Ooops. `Main.fs` shouldn't be commited. It is a leftover from trying `dotnet test` to work.
Removed it.

ThisFunctionalTom:
There are still two things I think we need to discuss about.

1. There is now a difference between running `Hedgehog.Tests` with `npm test`, `dotnet run` and `dotnet test`. The first two use the list `allTests` defined in `Program.fs` and the `dotnet test` uses the `TestAttribute` so if the one or the other is not added we will run different tests with different commands.
2. To be able to run `dotnet test` I had to add `YoloDev.Expecto.TestSdk` package to `Hedgehog.Tests.fsproj` and today, when I tried to add net461 back to TargetFrameworks (I forgot it before 😅), I found out that this package is not available for net461.

There is single solution to solve both problems, I think. We remove `YoloDev.Expecto.TestSdk` and `Microsoft.NET.Test.Sdk` packages from `Hedgehog.Tests`, add net461 back to TargetFrameworks and run the tests in the .NET world with `dotnet run`.

What do you think?

adam-becker:
>     1. There is now a difference between running `Hedgehog.Tests` with `npm test`, `dotnet run` and `dotnet test`. The first two use the list `allTests` defined in `Program.fs` and the `dotnet test` uses the `TestAttribute` so if the one or the other is not added we will run different tests with different commands.

I'm not sure why this is the case, can you explain a bit more?

>     2. To be able to run `dotnet test` I had to add `YoloDev.Expecto.TestSdk` package to `Hedgehog.Tests.fsproj` and today, when I tried to add net461 back to TargetFrameworks (I forgot it before sweat_smile), I found out that this package is not available for net461.

I'm not against removing `net461` from the test project, it doesn't provide any benefit for me. Other members of the project can weigh in on this one.

ThisFunctionalTom:
> > ```
> > 1. There is now a difference between running `Hedgehog.Tests` with `npm test`, `dotnet run` and `dotnet test`. The first two use the list `allTests` defined in `Program.fs` and the `dotnet test` uses the `TestAttribute` so if the one or the other is not added we will run different tests with different commands.
> > ```
> 
> I'm not sure why this is the case, can you explain a bit more?

OK, as I started writing the answer to your question I found out the solution for this (😁). But to be clear I will write what the problem was.
When run with `dotnet run` and with `npm test` the `allTests` list defines what tests will be run. OTOH when run with `dotnet test` all tests which have `TestsAttribute` will be run. Before I have added `TestsAttribute` to each test list in every file, now I put it on the `allTests` list and with this change we should always run the same tests in all three cases.
I hope I could bring some clarity and not confuse you even more 😅 .

> 
> > ```
> > 2. To be able to run `dotnet test` I had to add `YoloDev.Expecto.TestSdk` package to `Hedgehog.Tests.fsproj` and today, when I tried to add net461 back to TargetFrameworks (I forgot it before sweat_smile), I found out that this package is not available for net461.
> > ```
> 
> I'm not against removing `net461` from the test project, it doesn't provide any benefit for me. Other members of the project can weigh in on this one.

moodmosaic:
Is this pull request in a state where we could merge it? 🔖 Given the nature of this pull request, if we don't merge it it'll become stale soon (e.g. there's a new test added in #288).

adam-becker:
@moodmosaic I believe we can merge this.

tests/Hedgehog.Tests/GenTests.fs
```diff
@@ -1,94 +1,76 @@
...
-[<InlineData(128)>]
-[<InlineData(256)>]
-[<InlineData(512)>]
-let ``dateTime creates System.DateTime instances`` count =
-    let actual = Gen.dateTime (Range.constant System.DateTime.MinValue System.DateTime.MaxValue) |> Gen.sample 0 count
-    actual
-    |> List.distinct
-    |> List.length
-    =! actual.Length
+let dtRange = Range.constantFrom (System.DateTime (2000, 1, 1)) System.DateTime.MinValue System.DateTime.MaxValue
```
moodmosaic, line 7: Can we inline this (as it currently is in `master`) instead of extracting it to `dtRange`? It's nice to have autonomous tests, where each ([setup](http://xunitpatterns.com/fixture%20setup.html)) phase is just right there.

Hedgehog.sln
```diff
@@ -7,13 +7,13 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution items", "Solution
 		README.md = README.md
 	EndProjectSection
 EndProject
-Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "Hedgehog.Benchmarks", "tests/Hedgehog.Benchmarks/Hedgehog.Benchmarks.fsproj", "{7E2DE9D9-3831-46D7-B80F-B643E9D53FD3}"
+Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "Hedgehog.Benchmarks", "tests\Hedgehog.Benchmarks\Hedgehog.Benchmarks.fsproj", "{7E2DE9D9-3831-46D7-B80F-B643E9D53FD3}"
```
moodmosaic, line 10: Are those changes in `Hedgehog.sln` necessary?

moodmosaic (commented):
I left some minor comments, which I'm happy to address myself unless you want to address them @ThisFunctionalTom.

Once again, thank you for your hard work, @ThisFunctionalTom, and thank you for the thorough review @adam-becker and @TysonMN.

moodmosaic:
Also, now that #288 is merged, we must add that test also here.

ThisFunctionalTom:
> Also, now that #288 is merged, we must add that test also here. (It's the last one (Line 442) in tests/Hedgehog.Tests/ShrinkTests.fs)
> 
> ```fsharp
> let ``Property.reportWith respects shrinkLimit`` shrinkLimit =
> ...
> ```

Added the missing test from ShrinkTests.fs and also removed the old fable 2 compiler from package.json.
This time I had some troubles with rebase so I just merged master into fable-tests branch. I hope this is OK.

tests/Hedgehog.Tests/TestHelpers.fs
```diff
@@ -0,0 +1,43 @@
+module Hedgehog.Tests.TestHelpers
```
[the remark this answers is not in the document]
ThisFunctionalTom: Done

tests/Hedgehog.Tests/GenTests.fs
ThisFunctionalTom: I looked at all differences in tests from master and made the same changes on fable-test branch. The only difference in test bodies now is not using Unquote because it is not compatible with fable so instead of `test <@ some_expression @>` we now have `Expect.isTrue ( some_expression )`. Everything else should be completely the same. If you find some differences please tell me and I will correct it.

Hedgehog.sln
ThisFunctionalTom: No, this changes are not necessary. I guess it happened because i used `dotnet sln add` and `dotnet sln remove` for Hedgehog.Fable.Tests. I reverted the changes.

adam-becker:
> Added the missing test from ShrinkTests.fs and also removed the old fable 2 compiler from package.json.

:tada:

> This time I had some troubles with rebase so I just merged master into fable-tests branch. I hope this is OK.

As long as we can rebase (and/or squash) this branch before merging, I have no problems with you doing that to get unblocked.

tests/Hedgehog.Tests/GenTests.fs
moodmosaic: One of the reasons we're using Unquote is because the test failure messages can be given as step-by-step F# expression evaluations. Dropping Unquote is a trade-off here, but I think it's worth it since we'll have one test-suite for both .NET and Fable.

Do you also agree @adam-becker and @TysonMN?

adam-becker: @moodmosaic I mostly agree. I think having `Unquote` would be worthwhile, but I don't think it should prevent this PR being merged. I can't look at/modify the current tests because of an issue with ionide/fsac, so having these would be excellent for me.

adam-becker (approved):
Renaming `tests/Hedgehog.Tests/TestHelpers.fs` to `tests/Hedgehog.Tests/TestDsl.fs` is the only other change I can think of. I've approved in case you don't want to make that change yourself.

Great work, @ThisFunctionalTom! I think this is ready to be merged.

moodmosaic:
> Renaming `tests/Hedgehog.Tests/TestHelpers.fs` to `tests/Hedgehog.Tests/TestDsl.fs` is the only other change I can think of. I've approved in case you don't want to make that change yourself.
> 
> Great work, @ThisFunctionalTom! I think this is ready to be merged.

@adam-becker, perhaps you can merge and we can make that change afterwards. The question is... should we squash and merge and create a merge commit...

tests/Hedgehog.Tests/GenTests.fs
TysonMN: I also want just one test suit.  I am not convinced that the tradeoff must be so big.  What I say next doesn't have to be achieved in this PR.

> The only difference in test bodies now is not using Unquote because it is not compatible with fable so instead of `test <@ some_expression @>` we now have `Expect.isTrue ( some_expression )`.

It seems like it should be possible to me to write this as
```fsharp
let ourTest (expr: Expr<bool>) =
#if FABLE_COMPILER
    expr |> Expr.eval |> Expect.isTrue
#else
    expr |> test // this test is from Swensen.Unquote
#endif
```

where `Expr.eval` has type `Expr<'a> -> 'a`.  One possible implementation that works for my limited use case at work is
```fsharp
open Microsoft.FSharp.Quotations
open FSharp.Linq.RuntimeHelpers.LeafExpressionConverter
let eval (expr: Expr<'a>) =
  expr |> EvaluateQuotation :?> 'a
```

My understanding is that that implementation cannot handle all possible expressions.  In my notes, I wrote down how one might reach two different implementations.

> This evaluator can evaluate "simple" expressions.
 > It is contained in FSharp.Core.dll,
 > so no additional dependency is needed to use it.
 > 
 > FSharp.Quotations.Evaluator can evaluate more, but not all, expressions.
 > http://fsprojects.github.io/FSharp.Quotations.Evaluator/
 > 
 > Swensen.Unquote contains another evaluator that it uses internally.
 > https://github.com/SwensenSoftware/Unquote

Doesn't seem like the evaluator from Swensen.Unqoute is an option since lack for Fable by Swensen.Unquote is the whole issue.

adam-becker: @TysonMN Completely agree that we shouldn't lose `Unquote`, but I think we should restore that in a follow up PR. I can make an issue for `0.10.0` to make sure we do that before releasing.

ThisFunctionalTom:
> > Renaming `tests/Hedgehog.Tests/TestHelpers.fs` to `tests/Hedgehog.Tests/TestDsl.fs` is the only other change I can think of. I've approved in case you don't want to make that change yourself.
> > Great work, @ThisFunctionalTom! I think this is ready to be merged.
> 
> @adam-becker, perhaps you can merge and we can make that change afterwards. The question is... should we squash and merge, or create a merge commit...

Sorry, I just forgot this one. I changed the module name but not the file name 🤦‍♂️ . Now I did it.

adam-becker (approved):

adam-becker:
@ThisFunctionalTom If you're able to get this rebased I will merge it straight away.

ThisFunctionalTom:
> @ThisFunctionalTom If you're able to get this rebased I will merge it straight away.

Does it need to be a rebase? I merged the master into fable-test branch. Hope this is ok. I am not sure how to rebase now because I already merged master into fable-tests branch at commit d666faa2419300ad62d32823839d6837cd64681f

moodmosaic:
@adam-becker @TysonMN perhaps we want to squash and merge at this point(?) 🤔

TysonMN:
Yes, squashing sounds good to me.  If git it unable to do it, what we want is equivalent to copying the state of the repo on this branch, creating a new branch on `master`, pasting that repo state into the repo, and committing.

adam-becker:
I was able to rebase the branch, I'll give it another look and merge.

adam-becker:
@ThisFunctionalTom Thanks again! Great work.

PR #200, opened 2020-07-01
Make Hedgehog.Benchmarks the default startup project

bender2k14:
This PR makes the `Hedgehog.Benchmarks` project the default startup project by listing it first in the `sln` file.

This change improves the development experience. When first cloning the repo or after deleting the `.vs` directory and then opening the `sln` file with Visual Studio, the default startup project will now be the `Hedgehog.Benchmarks` project. This is an improvement because the `Hedgehog` project was the previous default startup project, but it cannot be executed.  (The two test projects can be started/executed, but I do not believe that it is intended behavior to do so.)

### DIFF
--- a/Hedgehog.sln
+++ b/Hedgehog.sln
@@ -8,13 +8,13 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution items", "Solution 
 		README.md = README.md
 	EndProjectSection
 EndProject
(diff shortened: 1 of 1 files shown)


moodmosaic:
Thank you :+1: The reason it's like this is because the benchmarks were added later on, so Visual Studio just appended the project at the end of the list. I don't see any reason holding us back from having it as the startup project since it's the only executable one.

Issue #201, opened 2020-07-02
perfect minimal shrink test failing in master


bender2k14:
Is it the case that the test named `greedy traversal with a predicate yields the perfect minimal shrink` fails in the `master` branch?
https://github.com/hedgehogqa/fsharp-hedgehog/blob/6d1c5ff8581d9d1bda052e18d256e6469b134ddf/tests/Hedgehog.Tests/MinimalTests.fs#L65-L85

That test in the `master` branch fails for me.  One small caveat is that I am not so experienced with paket, so I first replaced paket with the normal NuGet package references that I am used to, and then experienced that test failing.

I found this failing test while working on a fix for #199, which is motivated by tests that started failing after being built with a newer version of MSBuild (from a newer version of the .NET Core SDK).  Maybe that could also explain why I am seeing a failing test in the `master` branch.

I debugged this test for quite a while and could share what I learned, but maybe everything I learned is already considered "known" and so not worth repeating.  I would find it very helpful if someone could give a quick reply like
- "That test passes for me, and it should pass for you too" or
- "That test fails for me too but I didn't know that until you pointed it out" or
- "That test fails for everyone and we already know why".

moodmosaic:
Can you please re-run the test-suite from the command-line? Set the MSBuild verbosity level (`-v`) to normal (`n`) as in

```
dotnet test -v n tests/Hedgehog.Tests/Hedgehog.Tests.fsproj
```

---

On my machine ™

```
√ Hedgehog.Tests.MinimalTests.greedy traversal with a predicate yields the perfect minimal shrink [274ms]
```

bender2k14:
Certainly.  Here is the [full output](https://github.com/hedgehogqa/fsharp-hedgehog/files/4863983/output.txt).  The relevant output is
```
...
  X Hedgehog.Tests.MinimalTests.greedy traversal with a predicate yields the perfect minimal shrink [27ms]
  Error Message:
   System.NullReferenceException : Object reference not set to an instance of an object.
  Stack Trace:
     at Microsoft.FSharp.Core.FSharpFunc`2.InvokeFast[V](FSharpFunc`2 func, T arg1, TResult arg2)
   at Hedgehog.Random.unsafeRun[a](Seed seed, Int32 size, Random`1 _arg1) in C:\Code\ThirdParty\fsharp-hedgehog\src\Hedgehog\Random.fs:line 12
[...]

PR #202, opened 2020-07-02
Add test to ensure that genExp is not its default value

bender2k14:
This PR is related to issue #201 about the possibility of a test failing in the `master` branch.

For me, that test fails, and it fails very "late" in its execution.  I eventually determined that it fails because the reference it obtains to `genExp` is its `default` value.  (After realizing that, I found it easier to debug the failing test after removing `[<Struct>]` from the definitions of `Gen<>` and `Random<>`.  Then the test failed significantly sooner in its execution.)

This PR adds a test to ensure that `genExp` is not its default value (when `Gen<>` and `Random<>` are still defined with `[<Struct>]`... this test will fail if `[<Struct>]` is removed from the definition of either of those types).  It also adds some tests to ensure that the `isDefaultGen` test is implemented correctly.

It think it makes sense to merge this now before coming to a conclusion about issue #201 ~. Either something about my development environment is broken, which is causing that test in the `master` branch to fail, and so these new tests also pass for everyone else.  Or my development environment has identified a legitimate issue with that test, and so these new tests help to focus debugging efforts.~ since [the build triggered on travis ci by this PR passed](https://travis-ci.org/github/hedgehogqa/fsharp-hedgehog/builds/704290516).

### DIFF
--- a/tests/Hedgehog.Tests/MinimalTests.fs
+++ b/tests/Hedgehog.Tests/MinimalTests.fs
@@ -62,6 +62,27 @@ let rec genExp : Gen<Exp> =
         App <!> Gen.zip genExp genExp
     ]
 
+let isDefaultGen g =
+  let f =
+    match g with
+    | Gen (Random f) -> f
+  obj.ReferenceEquals(f, null)
+
+// https://github.com/hedgehogqa/fsharp-hedgehog/pull/202/
+[<Fact>]
+let ``isDefaultGen returns true for default value of Gen<Exp>``() =
+    test <@ Unchecked.defaultof<Gen<Exp>> |> isDefaultGen @>
+
+// https://github.com/hedgehogqa/fsharp-hedgehog/pull/202/
+[<Fact>]
+let ``isDefaultGen returns false for constant Gen<Exp> Lit 0``() =
+    test <@ 0 |> Lit |> Gen.constant |> isDefaultGen |> not @>
+
+// https://github.com/hedgehogqa/fsharp-hedgehog/pull/202/
+[<Fact>]
+let ``genExp is not its default value``() =
+    test <@ genExp |> isDefaultGen |> not @>
+
 [<Fact>]
 let ``greedy traversal with a predicate yields the perfect minimal shrink``() =
     Property.check <| property {


tests/Hedgehog.Tests/MinimalTests.fs
```diff
@@ -62,6 +62,24 @@ let rec genExp : Gen<Exp> =
         App <!> Gen.zip genExp genExp
     ]
 
+let isDefaultGen g =
+  let f =
+    match g with
+    | Gen (Random f) -> f
+  obj.ReferenceEquals(f, null)
+
```
moodmosaic, line 70:
```suggestion

// https://github.com/hedgehogqa/fsharp-hedgehog/pull/202/
```

tests/Hedgehog.Tests/MinimalTests.fs
```diff
@@ -62,6 +62,24 @@ let rec genExp : Gen<Exp> =
...
+[<Fact>]
+let ``isDefaultGen returns true for default value of Gen<Exp>``() =
+    test <@ Unchecked.defaultof<Gen<Exp>> |> isDefaultGen @>
+
```
moodmosaic, line 74:
```suggestion

// https://github.com/hedgehogqa/fsharp-hedgehog/pull/202/
```

tests/Hedgehog.Tests/MinimalTests.fs
```diff
@@ -62,6 +62,24 @@ let rec genExp : Gen<Exp> =
...
+[<Fact>]
+let ``isDefaultGen returns false for constant Gen<Exp> Lit 0``() =
+    test <@ 0 |> Lit |> Gen.constant |> isDefaultGen |> not @>
+
```
moodmosaic, line 78:
```suggestion

// https://github.com/hedgehogqa/fsharp-hedgehog/pull/202/
```

bender2k14:
Sure, those comments are good additions.  I accepted all of them, squashed locally, and force pushed.

moodmosaic:
Thanks! Here’s an idea... I can’t reproduce #201, and it looks like our Travis CI setup can’t either... But if merging this PR allows you to investigate further, I guess we can do it. After all, those added tests aren’t failing the build. Sounds good? 👀

bender2k14:
Yep, sounds great!  Thanks :D

PR #209, opened 2020-07-10
Target .NET Standard 2.0

### LINKED ISSUE #204: Add target .NET Standard 2.0
This is a spinoff of issue #199 

As pointed out in https://github.com/hedgehogqa/fsharp-hedgehog/issues/199#issuecomment-653670228, one advantage of targeting .NET Standard 2.0 even when also targeting .NET Standard 1.6 is because there can be fewer dependencies for users that are able to consume the version that targets .NET Standard 2.0.

bender2k14:
Fixes #204

I tested creation of the NuGet package by executing the command...
```
dotnet pack src/Hedgehog/Hedgehog.fsproj
```

...and it produced a NuGet package that listed these dependencies.
2020-07-10_15-16-20_159_devenv

I think this the correct behavior.

### DIFF
--- a/src/Hedgehog/Hedgehog.fsproj
+++ b/src/Hedgehog/Hedgehog.fsproj
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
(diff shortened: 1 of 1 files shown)


bender2k14:
I am ready to merge

Issue #156, opened 2018-02-16
Using linear and exponential ranges from C#


Porges:
When using `Range.Linear` I got this:

```
System.NotSupportedException: 'Specified method is not supported.'
   at Hedgehog.Range.Linear@155.Invoke(Int32 sz)
   at Hedgehog.Gen.List@338-1.Invoke(Seed seed, Int32 size)
   at Hedgehog.Random.bind@45.Invoke(Seed seed, Int32 size)
   at Hedgehog.Random.map@39-2.Invoke(Seed seed, Int32 size)
   at Hedgehog.Gen.bindRandom@40.Invoke(Seed seed0, Int32 size)
   at Hedgehog.Random.map@39-2.Invoke(Seed seed, Int32 size)
   at Hedgehog.Random.map@39-2.Invoke(Seed seed, Int32 size)
   at Hedgehog.Gen.bindRandom@40.Invoke(Seed seed0, Int32 size)
   at Hedgehog.Property.loop@332-21(Int32 n, Random`1 random, Seed seed, Int32 size, Int32 tests, Int32 discards)
   at Hedgehog.Property.Check(Property`1 g)
```

`Range.Constant` works fine, so I'm sticking with that at the moment. I can't see any obvious `NotSupportedException` uses in the code around line 155...

moodmosaic:
That's weird... We're using `Range.linear` in many places, like [here](https://github.com/hedgehogqa/fsharp-hedgehog/blob/763cd0000047ede067669816e6e783d783e1931e/tests/Hedgehog.Tests/RangeTests.fs#L151). 😕

moodmosaic:
Is this happening when you use Hedgehog from C#?

Porges:
Yes, it was.

moodmosaic:
The issue seems to be happening in here

https://github.com/hedgehogqa/fsharp-hedgehog/blob/9a95b50ae37f61539489647682b4ad0a0e2b3985/src/Hedgehog/Numeric.fs#L277

(Called from here)

https://github.com/hedgehogqa/fsharp-hedgehog/blob/9a95b50ae37f61539489647682b4ad0a0e2b3985/src/Hedgehog/Range.fs#L109-L114

---

On my local, I get a more specific exception actually:

```
System.NotSupportedException : Dynamic invocation of ToBigInt is not supported

Stack Trace:

\fsharp-hedgehog\src\Hedgehog\Range.fs(155,0):
[...]

PR #219, opened 2020-10-02
Mangle compiled name for functions that cannot be called from langs o…

mausch:
…ther than F#

prevents https://github.com/hedgehogqa/fsharp-hedgehog/issues/156

### DIFF
--- a/src/Hedgehog/Range.fs
+++ b/src/Hedgehog/Range.fs
@@ -77,7 +77,7 @@ module Range =
 
     /// Construct a range which is unaffected by the size parameter using the
     /// full range of a data type.
-    [<CompiledName("ConstantBounded")>]
+    [<CompiledName("-ConstantBounded")>]
     let inline constantBounded () : Range<'a> =
         let lo = minValue ()
         let hi = maxValue ()
@@ -140,7 +140,7 @@ module Range =
 
     /// Construct a range which scales the bounds relative to the size
     /// parameter.
-    [<CompiledName("LinearFrom")>]
+    [<CompiledName("-LinearFrom")>]
     let inline linearFrom (z : 'a) (x : 'a) (y : 'a) : Range<'a> =
         Range (z, fun sz ->
             let x_sized =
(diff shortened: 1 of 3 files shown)


moodmosaic (commented):
Thank you for the pull request, @mausch. See https://github.com/hedgehogqa/fsharp-hedgehog/issues/156#issuecomment-703052451 first.

src/Hedgehog/Range.fs
```diff
@@ -77,7 +77,7 @@ module Range =
 
     /// Construct a range which is unaffected by the size parameter using the
     /// full range of a data type.
-    [<CompiledName("ConstantBounded")>]
+    [<CompiledName("--ConstantBounded")>]
```
moodmosaic, line 80: Is `--` some special trick that'll hide the member in other .NET languages? (I tried also `-` and it worked the same, I think.)

mausch: One `-` would be the same yeah. It's just a way to set a name that's invalid for C# / VB.NET so you can't call it.

moodmosaic: Should we use just one `-` then? (The simplest possible thing that gets the job done.)

mausch: Sure why not. Changed in dbc14b4a4bf18a790e084532796e07e56faece07

moodmosaic: :+1: :100:

moodmosaic:
Alrighty, let's see if build passes.

moodmosaic:
Build fails for some non-obvious reason. Let's investigate. :thinking:

moodmosaic:
I'm not sure I can tell why build fails. Everything runs fine locally.

moodmosaic:
Unsure why it fails on CI

```
Total tests: 182. Passed: 182. Failed: 0. Skipped: 0.
Test Run Successful.
Test execution time: 9.2574 Seconds
     1>Done Building Project "D:\a\fsharp-hedgehog\fsharp-hedgehog\Hedgehog.sln" (VSTest target(s)) -- FAILED.

Build FAILED.
    0 Warning(s)
    0 Error(s)

Time Elapsed 00:00:21.27
Error: Process completed with exit code 1.
```

moodmosaic (approved):
:+1: :beers:

PR #222, opened 2020-10-04
Explicitly instantiate static inline functions for C# callers

mausch:
https://github.com/hedgehogqa/fsharp-hedgehog/issues/156#issuecomment-702977269


moodmosaic:
:confetti_ball:

PR #225, opened 2020-10-30
added missing output of Property.print in tutorial examples

TysonMN:
In both of these examples, `Property.print` doesn't just print the value of `a`; it also prints the value of `b`.


moodmosaic (approved):
:+1:

PR #228, opened 2020-10-30
Add Tree.toString

TysonMN:
Resolves #227

This gets the job done.  I am open to feedback to adhere to the standards of the project or the preferences of the maintainers.

Issue #226, opened 2020-10-30
toString versions of print functions


TysonMN:
I see three functions that print:
1. [Gen.printSample](https://github.com/hedgehogqa/fsharp-hedgehog/blob/043b4648b6f8aaa590a35a13ee852e8a240e06d5/src/Hedgehog/Gen.fs#L547-L555),
2. [Property.print](https://github.com/hedgehogqa/fsharp-hedgehog/blob/043b4648b6f8aaa590a35a13ee852e8a240e06d5/src/Hedgehog/Property.fs#L411-L414), and
3. [Property.print'](https://github.com/hedgehogqa/fsharp-hedgehog/blob/043b4648b6f8aaa590a35a13ee852e8a240e06d5/src/Hedgehog/Property.fs#L405-L408).

As I explore the codebase and try to understand how things work, I would find it more convenient if a variant for each of these functions existed that returns a string (instead of directly printing that string).  For example, while debugging, I would prefer to compute that string, assign it to a local variable, and view its value in the debugger.

Would it be ok for each of these printing functions to add an additional function that returns the string currently being printed?

moodmosaic:
That would be good idea :+1: 

FTR, some of the stuff in F# Hedgehog existed before the Haskell version comes out (in this case that's why there's no such function for rendering a string) but the ultimate goal is to keep in sync at least `Gen`, `Tree`, and `Shrink` with Haskell.

If you're willing to work on it, it'd be nice to follow the same style and naming as in the Haskell codebase, where for example `printTree` ends up calling [`renderTree`](https://github.com/hedgehogqa/haskell-hedgehog/blob/7030059c089b45429c9ed23f37971e1b5334dc2e/hedgehog/src/Hedgehog/Internal/Gen.hs#L1764) internally (which is the function that returns the string you want).

TysonMN:
I like this goal and am happy to follow the same style and naming as in the Haskell codebase.

As I gain more experience with this codebase, I will just directly create a PR when I want to add a feature that already exists in the Haskell codebase.

moodmosaic:
Great, @TysonMN :100:

PR #229, opened 2020-11-02
Add Tree to string function

### LINKED ISSUE #227: printTree function
Haskell Hedgehog has a function [`printTree`](https://github.com/hedgehogqa/haskell-hedgehog/search?q=printTree) that can print trees of generators [like this](https://github.com/hedgehogqa/haskell-hedgehog/pull/406#issuecomment-709275941).  I think such a function would help me in debugging issue #224.  However, just like issue #226, I would prefer if this was a function that returned this tree as a string instead of directly printing it.

Can a `Gen.printTree` or `Gen.treeToString` function be added to F# Hedgehog?

TysonMN:
Resolves #227

(This PR was previously PR #228 with a different branch name)

This gets the job done. I am open to feedback to adhere to the standards of the project or the preferences of the maintainers.

Based on comment https://github.com/hedgehogqa/fsharp-hedgehog/issues/226#issuecomment-719817482 by @moodmosaic, I named these functions `render` (instead of `toString` like in PR #228).

After looking more closely at the Haskell code, the function that returns `string list` I named `renderLines`.  I also restricted these functions to only work for trees containing data of type `string`.

### DIFF
--- a/src/Hedgehog/Tree.fs
+++ b/src/Hedgehog/Tree.fs
@@ -89,23 +89,28 @@ module Tree =
         Seq.filter (f << outcome) xs
         |> Seq.map (filter f)
 
-    let rec renderLines (Node (x, xs) : Tree<string>) : string list =
-        let mapFirstDifferently f g = function
-            | [] -> []
-            | x :: xs -> (f x) :: (xs |> List.map g)
-        let mapLastDifferently f g = List.rev >> mapFirstDifferently g f >> List.rev
-        let tail =
-            xs
-            |> Seq.map renderLines
-            |> Seq.toList
-            |> mapLastDifferently
-                (mapFirstDifferently ((+) "├-")
-                                     ((+) "| "))
-                (mapFirstDifferently ((+) "└-")
-                                     ((+) "  "))
+    let rec renderLines (Node (x, xs00) : Tree<string>) : List<string> =
+        // zipWith (++) (hd : repeat other)
+        let shift f g xs0 =
+            match xs0 with
+            | [] ->
+                []
+            | x :: xs ->
+                (f x) :: (List.map g xs)
+
+        let xs =
+            xs00
+            |> List.ofSeq
+            |> List.map renderLines
+            |> shift
+                   (shift ((+) "├-") ((+) "| "))
+                   (shift ((+) "└-") ((+) "  "))
             |> List.concat
             |> List.map ((+) " ")
-        x :: tail
+
+        x :: xs
 
     let render (t : Tree<string>) : string =
-        renderLines t |> Seq.reduce (fun a b -> a + System.Environment.NewLine + b)
+        renderLines t
+        |> Seq.reduce (fun a b ->
+            a + System.Environment.NewLine + b)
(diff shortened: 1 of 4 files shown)


moodmosaic (commented):
Thank you, Tyler :+1:  I should be able to review this during the week, though from what I see it looks reasonable as-is.

tests/Hedgehog.Tests/TreeTests.fs
```diff
@@ -0,0 +1,58 @@
+﻿module Hedgehog.Tests.TreeTests
+
+open Hedgehog
+open Swensen.Unquote
+open Xunit
+
+[<Fact>]
+let ``render of singleton tree`` () =
+    Property.check <| property {
+        let! v = Gen.uint64 (Range.exponentialBounded ())
```
moodmosaic, line 10: Any strong reason around `uint64` instead of something smaller? e.g. `uint16`

moodmosaic (commented):
I did some housekeeping which will push over the next few days. It looks good either ways! I may be adding a few more comments too.

tests/Hedgehog.Tests/TreeTests.fs
TysonMN: I just copied it from some other test.  I think any generator will do here.

src/Hedgehog/Tree.fs
```diff
@@ -88,3 +88,29 @@ module Tree =
     and filterForest (f : 'a -> bool) (xs : seq<Tree<'a>>) : seq<Tree<'a>> =
         Seq.filter (f << outcome) xs
         |> Seq.map (filter f)
+
+    let rec renderList (Node (x, xs00) : Tree<string>) : List<string> =
```
moodmosaic, line 92: @TysonMN, is it OK if we name this `renderList`? It returns a list, at the end...

TysonMN: Yep.  I don't have a preference on the name.  I only picked `renderLines` to be consistent with the names of [these Haskell Hedgehog functions](https://github.com/hedgehogqa/haskell-hedgehog/blob/7030059c089b45429c9ed23f37971e1b5334dc2e/hedgehog/src/Hedgehog/Internal/Tree.hs#L672-L719).

moodmosaic: Yep, those are named like that because of Monad `m` while here we're wokring with List.

I was going back and forth, but I think from a consumer's point of view `renderList` is easier to understand.

We can surely change this until we do a release.

tests/Hedgehog.Tests/TreeTests.fs
```diff
@@ -0,0 +1,79 @@
+﻿module Hedgehog.Tests.TreeTests
+
+open Hedgehog
+open Swensen.Unquote
+open Xunit
+
+[<Fact>]
+let ``render tree with depth 0`` () =
+    Property.check <| property {
```
moodmosaic, line 9: We can have those run just once instead of 100 times. I'll leave them as-is for now as I'm planning to do some separate houskeeping to the whole test-suite.

moodmosaic:
Thank you, @TysonMN 💯

src/Hedgehog/Tree.fs
TysonMN: I do think `renderList` is better

PR #230, opened 2020-11-04
added shrink direction to the frequency generator documentation

TysonMN:
I added the same sentence that exists in the [Haskell documentation for the frequency generator](https://hackage.haskell.org/package/hedgehog-1.0.3/docs/Hedgehog-Gen.html#v:frequency).


moodmosaic (approved):
Thank you, Tyson :+1:

PR #231, opened 2020-11-06
Fix the 'atLeast' function

### LINKED ISSUE #217: Gen.list fails when lower bound is > 0
I fear this is a bug:

https://stackoverflow.com/questions/64109366/how-can-i-define-a-generator-in-hedgehog-that-generates-lists-of-certain-length

     Gen.list (Range.linear 10 20) (Gen.int <| Range.constant 0 5) |> Gen.printSample

fails when generating shrinked values. It only works for lists with lower bound 0. 
Also generator documentation could be enhanced,

---
Maybe in Gen.fs it should be like

    let private atLeast (n : int) (xs : List<'a>) : bool =
        let nn = min (n - 1) xs.Length
        n = 0 || not (List.isEmpty (List.skip nn xs))

adam-becker:
Fixes #217

### DIFF
--- a/src/Hedgehog/Gen.fs
+++ b/src/Hedgehog/Gen.fs
@@ -71,7 +71,7 @@ module Gen =
 
 #if !FABLE_COMPILER
     [<CompiledName("Map")>]
-#endif    
+#endif
     let map2 (f : 'a -> 'b -> 'c) (gx : Gen<'a>) (gy : Gen<'b>) : Gen<'c> =
         bind gx <| fun x ->
         bind gy <| fun y ->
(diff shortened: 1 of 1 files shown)


adam-becker:
Sorry about the extra whitespace changes, my editor is configured to remove extraneous whitespace. I can revert those changes if necessary.

moodmosaic:
@adam-becker, thank you for the pull request 💯 👍 

>Sorry about the extra whitespace changes, my editor is configured to remove extraneous whitespace.

Not a problem at all, in fact removing extraneous whitespace is good, so thanks for the cleanup.

moodmosaic (approved):
👍

PR #233, opened 2020-11-08
Adding recheck method and seed reporting

### LINKED ISSUE #179: Print seed to reproduce failure (as FsCheck allows)
In FsChesk, when a test fails, the seed from the random number is printed. The seed can then be feed into the test to reproduce the same failure. 

Can this be done in hedgehog? I could not see examples or any seed output.

adam-becker:
Fixes #179

### DIFF
--- a/src/Hedgehog/Property.fs
+++ b/src/Hedgehog/Property.fs
@@ -156,7 +156,7 @@ module private Pretty =
         List.iter (append sb) (Journal.toList journal)
 
         renderf sb "This failure can be reproduced by running:"
-        renderf sb "> Property.recheck (%d : Size) ({ Value = %A; Gamma = %A }) <property>" size seed.Value seed.Gamma
+        renderf sb "> Property.recheck (%d : Size) (%0A) <property>" size seed
 
         sb.ToString (0, sb.Length - 1) // Exclude extra newline.
 
@@ -357,7 +357,7 @@ module Property =
                 | Failure ->
                     { Tests = tests + 1<tests>
                       Discards = discards
-                      Status = takeSmallest size seed1 result 0<shrinks> }
+                      Status = takeSmallest size seed result 0<shrinks> }
                 | Success () ->
                     loop seed2 (nextSize size) (tests + 1<tests>) discards
                 | Discard ->
--- a/src/Hedgehog/Property.fs
+++ b/src/Hedgehog/Property.fs
@@ -128,6 +128,9 @@ module private Pretty =
     let private append (sb : StringBuilder) (msg : string) : unit =
         sb.AppendLine msg |> ignore
 
+    let private appendf (sb : StringBuilder) fmt =
+        Printf.kprintf (ignore << sb.AppendLine) fmt
+
     let renderOK (tests : int<tests>) : string =
         sprintf "+++ OK, passed %s." (renderTests tests)
 
@@ -144,16 +147,15 @@ module private Pretty =
             (journal : Journal) : string =
         let sb = StringBuilder ()
 
-        sprintf "*** Failed! Falsifiable (after %s%s%s):"
+        appendf sb "*** Failed! Falsifiable (after %s%s%s):"
             (renderTests tests)
             (renderAndShrinks shrinks)
             (renderAndDiscards discards)
-            |> append sb
-
-        sprintf "--- Seed to recheck this case: %d" seed.Value
-            |> append sb
 
         List.iter (append sb) (Journal.toList journal)
+
+        appendf sb "This failure can be reproduced by running:"
+        appendf sb "> Property.recheck (Size 0) ({ Value = %d; Gamma = %d }) prop_failure" seed.Value seed.Gamma
 
         sb.ToString(0, sb.Length - 1) // exclude extra newline
 
--- a/src/Hedgehog/Property.fs
+++ b/src/Hedgehog/Property.fs
@@ -128,8 +128,8 @@ module private Pretty =
     let private append (sb : StringBuilder) (msg : string) : unit =
         sb.AppendLine msg |> ignore
 
-    let private appendf (sb : StringBuilder) fmt =
-        Printf.kprintf (ignore << sb.AppendLine) fmt
+    let private appendf (sb : StringBuilder) (fmt : Printf.StringFormat<'a, unit>) : 'a =
+        Printf.ksprintf (ignore << sb.AppendLine) fmt
 
     let renderOK (tests : int<tests>) : string =
         sprintf "+++ OK, passed %s." (renderTests tests)
--- a/src/Hedgehog/Property.fs
+++ b/src/Hedgehog/Property.fs
@@ -20,7 +20,7 @@ type Property<'a> =
 [<Measure>] type shrinks
 
 type Status =
-    | Failed of int<shrinks> * Journal
+    | Failed of Seed * int<shrinks> * Journal
     | GaveUp
     | OK
 
@@ -137,6 +137,7 @@ module private Pretty =
             (renderTests tests)
 
     let renderFailed
+            (seed: Seed)
             (tests : int<tests>)
             (discards : int<discards>)
             (shrinks : int<shrinks>)
@@ -147,6 +148,9 @@ module private Pretty =
             (renderTests tests)
             (renderAndShrinks shrinks)
             (renderAndDiscards discards)
+            |> append sb
+
+        sprintf "--- Seed to recheck this case: %d" seed.Value
             |> append sb
 
         List.iter (append sb) (Journal.toList journal)
@@ -163,8 +167,8 @@ type GaveUpException (tests : int<tests>, discards : int<discards>) =
     member __.Tests =
         tests
 
-type FailedException (tests : int<tests>, discards : int<discards>, shrinks : int<shrinks>, journal : Journal) =
-    inherit HedgehogException (renderFailed tests discards shrinks journal)
+type FailedException (seed: Seed, tests : int<tests>, discards : int<discards>, shrinks : int<shrinks>, journal : Journal) =
+    inherit HedgehogException (renderFailed seed tests discards shrinks journal)
 
     member __.Tests =
         tests
@@ -187,8 +191,8 @@ module Report =
             renderOK report.Tests
         | GaveUp ->
             renderGaveUp report.Tests report.Discards
-        | Failed (shrinks, journal) ->
-            renderFailed report.Tests report.Discards shrinks journal
+        | Failed (seed, shrinks, journal) ->
+            renderFailed seed report.Tests report.Discards shrinks journal
 
     let tryRaise (report : Report) : unit =
         match report.Status with
@@ -196,8 +200,8 @@ module Report =
             ()
         | GaveUp ->
             raise <| GaveUpException (report.Tests, report.Discards)
-        | Failed (shrinks, journal)  ->
-            raise <| FailedException (report.Tests, report.Discards, shrinks, journal)
+        | Failed (seed, shrinks, journal)  ->
+            raise <| FailedException (seed, report.Tests, report.Discards, shrinks, journal)
 
 module Property =
 
@@ -299,7 +303,7 @@ module Property =
 
 #if !FABLE_COMPILER
     [<CompiledName("ForAll")>]
-#endif    
+#endif
     let forAll' (gen : Gen<'a>) : Property<'a> =
         forAll gen success
 
@@ -308,24 +312,22 @@ module Property =
     //
 
     let rec private takeSmallest
+            (seed: Seed)
             (Node ((journal, x), xs) : Tree<Journal * Result<'a>>)
             (nshrinks : int<shrinks>) : Status =
         match x with
         | Failure ->
             match Seq.tryFind (Result.isFailure << snd << Tree.outcome) xs with
             | None ->
-                Failed (nshrinks, journal)
+                Failed (seed, nshrinks, journal)
             | Some tree ->
-                takeSmallest tree (nshrinks + 1<shrinks>)
+                takeSmallest seed tree (nshrinks + 1<shrinks>)
         | Discard ->
             GaveUp
         | Success _ ->
             OK
 
-#if !FABLE_COMPILER
-    [<CompiledName("Report")>]
-#endif    
-    let report' (n : int<tests>) (p : Property<unit>) : Report =
+    let private reportWithSeed' (seed : Seed) (n : int<tests>) (p : Property<unit>) : Report =
         let random = toGen p |> Gen.toRandom
 
         let nextSize size =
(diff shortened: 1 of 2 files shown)


src/Hedgehog/Property.fs
```diff
@@ -149,6 +150,9 @@ module private Pretty =
             (renderAndDiscards discards)
             |> append sb
 
+        sprintf "--- Seed to recheck this case: %d" seed.Value
```
moodmosaic, line 153: It'd be nice to have this *after* the shrinked value (at the very end) so that we're consistent with what we do in the Haskell version.

Also, instead of

```
--- Seed to recheck this case: 11194825392996116235
```

it'd be nice to give to the user the full command to run, e.g.

```
> Property.recheck (Size 0) (Seed 6047214807766778671 8250247924578451779) prop_failure
```

moodmosaic (commented):
This would be nice to have 👍 I haven't looked into the implementation yet, but I checked out the branch and run a test and left you some feedback.

src/Hedgehog/Property.fs
adam-becker: I added nearly the string requested, `Seed` is a record, and F# doesn't support the syntax used in the above message. When you used `prop_failure` was that meant to be taken literally? Not sure how else to identify the property if that wasn't your intention.

moodmosaic: No, it's just an example from what we do in Haskell. I should have mentioned it. `prop_failure` is the name of the failing property.

moodmosaic: We can replace it with `<property>` for example, or just omit it for now.

adam-becker: Do properties have names in the F# version?

src/Hedgehog/Property.fs
```diff
@@ -128,6 +128,9 @@ module private Pretty =
     let private append (sb : StringBuilder) (msg : string) : unit =
         sb.AppendLine msg |> ignore
 
+    let private appendf (sb : StringBuilder) fmt =
```
moodmosaic, line 131:
```suggestion
    let private appendf (sb : StringBuilder) (fmt : Printf.StringFormat<'a, unit>) : 'a =
```

src/Hedgehog/Property.fs
```diff
@@ -128,6 +128,9 @@ module private Pretty =
...
+        Printf.kprintf (ignore << sb.AppendLine) fmt
```
moodmosaic, line 132:
```suggestion
        Printf.ksprintf (ignore << sb.AppendLine) fmt
```

You meant [`ksprintf`](https://fsharp.github.io/fsharp-core-docs/reference/fsharp-core-printfmodule.html#ksprintf), right? :arrow_up:

moodmosaic (approved):
We could have this into this release :+1: I just left two small suggestions/nitpicks. Thank you for your work on this, @adam-becker :100:

src/Hedgehog/Property.fs
adam-becker: Yeah, sorry. Good catch!

src/Hedgehog/Property.fs
```diff
@@ -137,20 +140,23 @@ module private Pretty =
...
+        appendf sb "*** Failed! Falsifiable (after %s%s%s):"
             (renderTests tests)
             (renderAndShrinks shrinks)
             (renderAndDiscards discards)
-            |> append sb
 
         List.iter (append sb) (Journal.toList journal)
 
+        appendf sb "This failure can be reproduced by running:"
+        appendf sb "> Property.recheck (Size 0) ({ Value = %d; Gamma = %d }) <property>" seed.Value seed.Gamma
```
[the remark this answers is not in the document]
moodmosaic: It turns out we want the `%A` format specifier to be able to print the seed correctly

e.g.:

```
Property.recheck ({ Value = 17415578665524963940UL; Gamma = 4364604868884628433UL }) <property>
``` 

with `%d` and `%u` you'd have to append *UL* manually:

```
Property.recheck ({ Value = 17415578665524963940; Gamma = 4364604868884628433 }) <property>
```

moodmosaic (changes requested):
I was adding some tests before merging and came across an issue with `recheck` where the message we show differs from the way the `recheck` function is defined.

src/Hedgehog/Property.fs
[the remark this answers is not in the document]
moodmosaic: Added via https://github.com/hedgehogqa/fsharp-hedgehog/pull/233/commits/63005fab340f81db41da6a28a393fb925becffb3.

moodmosaic (commented):
@adam-becker, I've passed the size to the message (which is basically tests + discards). One last thing, I'd like to see if we can have 

```
> Property.recheck (Size 1) (Seed 2244935751425070623UL 5562567424033617363UL) <property>
```

instead of

```

> Property.recheck (Size 1) ({ Value = 2244935751425070623UL; Gamma = 5562567424033617363UL }) <property>
```

moodmosaic:
FWIW, I've converted this into a draft; we're rendering the size in the output message and what's left is to also pass the size in. For example, `reportWithSeed` must take also `size`, and must be renamed.

moodmosaic (approved):

moodmosaic:
@willsam100, @adam-becker, would you care trying this branch on your local?

TysonMN:
I will also try this branch out later today

TysonMN:
I don't think it is working.  Maybe I am doing it wrong though.  A second issue is that when calling `Property.recheck` on a failing test still outputs how to recheck the test.  Below, I explain how I tested and then give some closing thoughts.

----------------

Here is the test I wrote to see if `Property.recheck` is actually rechecking the same previously failed test case.

```fsharp
let mutable failFlag = true

[<Fact>]
let ``only fail first check`` () =
    Property.check <| property {
        let! n = Range.constant System.Int64.MinValue System.Int64.MaxValue |> Gen.int64
        if failFlag then
            failFlag <- false
            failwith "failFlag was true"
    }
```

Here is the output from one of my executions.

```
Hedgehog.FailedException : *** Failed! Falsifiable (after 1 test):
65061791783614996L
System.Exception: failFlag was true
   at PropertyTests.only fail first check@13-1.Invoke(Int64 _arg1) in C:\Code\ThirdParty\fsharp-hedgehog\tests\Hedgehog.Tests\PropertyTests.fs:line 15
   at Hedgehog.Property.prepend@304-2.Invoke(Unit _arg1) in C:\Code\ThirdParty\fsharp-hedgehog\src\Hedgehog\Property.fs:line 304
> Property.recheck (1 : Size) ({ Value = 1343219580531722066UL; Gamma = 6719782976908311835UL }) <property>
```

Now I expect executing the next test will cause `65061791783614996L` to be generated again.

```fsharp
let mutable failFlag = true

[<Fact>]
let ``only fail first check`` () =
    Property.recheck (1 : Size) ({ Value = 1343219580531722066UL; Gamma = 6719782976908311835UL }) <| property {
        let! n = Range.constant System.Int64.MinValue System.Int64.MaxValue |> Gen.int64
        if failFlag then
            failFlag <- false
            failwith "failFlag was true"
    }
```

But this is not the case.  Instead, `-5078635085846143995L` is generated.

```
Hedgehog.FailedException : *** Failed! Falsifiable (after 1 test):
-5078635085846143995L
System.Exception: failFlag was true
   at PropertyTests.only fail first check@13-1.Invoke(Int64 _arg1) in C:\Code\ThirdParty\fsharp-hedgehog\tests\Hedgehog.Tests\PropertyTests.fs:line 15
   at Hedgehog.Property.prepend@304-2.Invoke(Unit _arg1) in C:\Code\ThirdParty\fsharp-hedgehog\src\Hedgehog\Property.fs:line 304
> Property.recheck (1 : Size) ({ Value = 5500455877167666293UL; Gamma = 6719782976908311835UL }) <property>
```

Now, I don't think `Property.recheck` should display instructions for how to repeat the test.  However, given that this data is being displayed, we can take advantage of it for testing.  Notice that the `seed` I gave is not the `seed` that message says should be used to reproduce the test.

On the other hand, at least that test is deterministic now: running that test will always generate `-5078635085846143995L` and will always incorrectly claim that the `seed` to reproduce is `{ Value = 5500455877167666293UL; Gamma = 6719782976908311835UL }`.

-----------------

First, I recommend automating this manual testing that I have done.  I would do this by having an internal function that returns the `size` and `seed`.  Then the current `Property.check` function and a test can both call this function (because internals can be given access to the test project).  The test would record the value of `n` for which property fails, use the returned `size` and `seed` to execute the generator again, and see if the same value of `n` is returned.  (Since `renderf` is impure, it should be called very near the entry point, which in this case is `Property.check`.)

Second, I recommend getting that test to pass.

Third, I recommend excluding the `This failure can be reproduced by running` output when `Property.recheck` is called.

adam-becker:
@TysonMN I've fixed the issue you pointed out, though `recheck` still reports the seed when the case fails.
@moodmosaic I've changed the format specifier to be `%0A` when printing the seed, so everything is formatted properly.

The function is a little clunky to use, however. Since the size parameter in the message is always going to be `(1 : Size)`, can't it be omitted from both the message and the function's parameter list?

TysonMN:
> @TysonMN I've fixed the issue you pointed out

Great!  Did you push that change?  I don't see it.

> Since the size parameter in the message is always going to be (1 : Size), can't it be omitted from both the message and the function's parameter list?

It is not always `1`.  It will always be `1` in the test I gave above.  Here is a test where it is always `2`.

```fsharp
let mutable testsToPass = 1

[<Fact>]
let ``fail after first test`` () =
    Property.check <| property {
        let! n = Range.constant System.Int64.MinValue System.Int64.MaxValue |> Gen.int64
        if testsToPass = 0 then
            failwith "now always failing"
        else
            testsToPass <- testsToPass - 1
    }
```

Sample output:
```
Hedgehog.FailedException : *** Failed! Falsifiable (after 2 tests and 1 shrink):
-9223372036854775808L
System.Exception: now always failing
   at PropertyTests.fail after first test@13-1.Invoke(Int64 _arg1) in C:\Code\ThirdParty\fsharp-hedgehog\tests\Hedgehog.Tests\PropertyTests.fs:line 14
   at Hedgehog.Property.prepend@304-2.Invoke(Unit _arg1) in C:\Code\ThirdParty\fsharp-hedgehog\src\Hedgehog\Property.fs:line 304
> Property.recheck (2 : Size) ({ Value = 6052627415793634358UL; Gamma = 17135851100886004659UL }) <property>
```

At the "top level", `size` is the test number.  By default, each property is tested `100` times.  The first value of the `size` parameter starts as `1` and its last value is `100`.

moodmosaic:
Size shouldn't be hardcoded to 1 because size = tests + discards

moodmosaic:
I'm on my phone right now (hence the brevity) but here's a property worth testing on this branch

```f#
property {
    let! x = Gen.int <| Range.constantBounded ()
    let! y = Gen.int <| Range.constantBounded ()
    where (x > 0 && y > 0)
    counterexample (sprintf "x * y = %d" <| x * y)
    return x * y > 0
}
```

adam-becker:
> Great! Did you push that change? I don't see it.
Sorry, I thought I had.

> Size shouldn't be hardcoded to 1 because size = tests + discards
Not sure how I was under that impression, thanks for clarifying.

TysonMN:
I tested that commit.  It works correctly now.  Great work :)

src/Hedgehog/Property.fs
```diff
@@ -137,21 +140,25 @@ module private Pretty =
...
             (renderTests tests)
             (renderAndShrinks shrinks)
             (renderAndDiscards discards)
-            |> append sb
 
         List.iter (append sb) (Journal.toList journal)
 
-        sb.ToString(0, sb.Length - 1) // exclude extra newline
+        renderf sb "This failure can be reproduced by running:"
+        renderf sb "> Property.recheck (%d : Size) (%0A) <property>" size seed
```
TysonMN, line 159: (Repeating my suggestion from https://github.com/hedgehogqa/fsharp-hedgehog/pull/233#issuecomment-735229874)

I think it would be better if this were only printed when `Property.check` is called and not also printing it when `Property.recheck` is called.

TysonMN (changes requested):

src/Hedgehog/Property.fs
moodmosaic: I don't think it's a huge blocker if everything else works fine. If it's something easy to do, let's do it... If not, we can wrap up, merge and do a release. (And get back to this in a separate PR.)

moodmosaic:
Great progress so far 👀 @adam-becker🏅@TysonMN 💯

src/Hedgehog/Property.fs
TysonMN: I agree with that.  I am willing to improve this at a later time.

With that bug fix in https://github.com/hedgehogqa/fsharp-hedgehog/pull/233/commits/a697cc135f6092fa2c8f2107767dedbb715fba62, the branch in this PR is now a great improvement over `master`.

adam-becker: I did manage to make this change. In doing so, I reduced the arity of some of the functions in the call stack of `recheck`.

src/Hedgehog/Property.fs
```diff
@@ -128,76 +136,80 @@ module private Pretty =
...
+            (renderTests report.Tests)
+            (renderAndShrinks failure.Shrinks)
+            (renderAndDiscards report.Discards)
+
+        List.iter (append sb) (Journal.toList failure.Journal)
 
-        List.iter (append sb) (Journal.toList journal)
+        if failure.RenderRecheck then
+            renderf sb "This failure can be reproduced by running:"
+            renderf sb "> Property.recheck (%d : Size) (%0A) <property>"
```
moodmosaic, line 162: I think this won't work in F# interactive, it'll render the record values without semicolon `;` e.g.

```f#
Property.recheck (2 : Size) ({ Value = 6795855140008116271UL Gamma = 17914027628548019555UL }) <property>
```

but what we'd want is

```f#
Property.recheck (2 : Size) ({ Value = 6795855140008116271UL; Gamma = 17914027628548019555UL }) <property>
```

```suggestion
            renderf sb "> Property.recheck (%d : Size) ({ Value = %A; Gamma = %A }) <property>"
```

src/Hedgehog/Property.fs
```diff
@@ -128,76 +136,80 @@ module private Pretty =
...
+                failure.Size
+                failure.Seed
```
moodmosaic, line 164:
```suggestion
                failure.Seed.Value
                failure.Seed.Gamma
```

moodmosaic (commented):
Awesome! Getting really close. Left a comment which I think makes sense specially for F# Interactive users.

adam-becker:
> Awesome! Getting really close. Left a comment which I think makes sense specially for F# Interactive users.

Done.

src/Hedgehog/Property.fs
TysonMN: Indeed you did.  Excellent work! :D

moodmosaic:
:confetti_ball:

PR #234, opened 2020-11-09
Reduce noise from GitHub's unchanged files with check annotations feature

TysonMN:
This PR upgrades the CI build to use version 3.1.403 of the .NET SDK.  This is the most recently released version.  The CI build is currently using version 2.1.4.  Both the 2.1 series and the 3.1 series are long-term support (LTS) releases.

I have created this PR to resolve the FS0058 warnings, which are about possibly incorrect indentations.  The F# compiler in the newer SDK is satisfied with the current indentation.

Most practically, this PR will remove these warnings from the files view of a pull request in GitHub.  A beta feature of GitHub under the heading [Unchanged files with check annotations](https://github.com/hedgehogqa/fsharp-hedgehog/pull/233/files#diff-9b7fcde4f0a23cc861c6b1da4c00e8eaf668e9db5a0b07ae8ab2527e3a76373fR66:~:text=Unchanged%20files%20with%20check%20annotations) shows the lines with these warnings.

### DIFF
--- a/.github/workflows/dotnet-core.yml
+++ b/.github/workflows/dotnet-core.yml
@@ -20,7 +20,7 @@ jobs:
     - name: Setup .NET Core
       uses: actions/setup-dotnet@v1
       with:
-        dotnet-version: 2.1.4
+        dotnet-version: 3.1.403
     - name: Install dependencies
       run: dotnet restore
     - name: Build
--- a/tests/Hedgehog.Benchmarks/Hedgehog.Benchmarks.fsproj
+++ b/tests/Hedgehog.Benchmarks/Hedgehog.Benchmarks.fsproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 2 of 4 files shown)


moodmosaic:
Very nice 👍 Should we run another CI build? I'm not sure why this one failed.

TysonMN:
Oh, my bad.  I had tested the CI build [here in a PR in my fork](https://github.com/TysonMN/fsharp-hedgehog/pull/4/checks), but I stopped watching the build after the complication step completed without any warnings.

I will figure out the problem and fix this.

moodmosaic:
Okay 👍

TysonMN:
The tests targeting .NET Core 2.0 are failing because of bug https://github.com/actions/setup-dotnet/issues/136.  I tried installing two versions of the .NET Core SDK, but [that didn't work](https://github.com/TysonMN/fsharp-hedgehog/runs/1374601455?check_suite_focus=true#step:7:10).

The workaround I committed is to increase the targets of the test projects.  I think this is a good tradeoff (i.e. test using 3.0 instead of 2.0 in exchange for not having build warnings pollute the file view of every PR in  GitHub).

Another workaround would be to adjust the indentation so that the older version of the #F compiler is satisfied.  I could do that workaround if you prefer.

moodmosaic:
>The workaround I committed is to increase the targets of the test projects. I think this is a good tradeoff (i.e. test using 3.0 instead of 2.0 in exchange for not having build warnings pollute the file view of every PR in GitHub).

That's a good choice as we aren't actually introducing any breaking change to the library itself 👍

moodmosaic (approved):
Thank you for the analysis and for providing a solution to this 👍 💯

PR #235, opened 2020-11-09
improved the Range documentation

TysonMN:


src/Hedgehog/Range.fs
```diff
@@ -8,9 +8,10 @@ open Hedgehog.Numeric
 type Size = int
 
 /// A range describes the bounds of a number to generate, which may or may not
-/// be dependent on a 'Size'.
+/// be dependent on a 'Size'.  The pair returned by the function contains the
+/// lower bound on the left and the upper bound on the right.
 type Range<'a> =
-    | Range of ('a * (Size -> 'a * 'a))
+    | Range of origin:'a * (Size -> 'a * 'a)
```
moodmosaic, line 14:
```suggestion
    | Range of (origin : 'a) * (Size -> 'a * 'a)
```

TysonMN: Sure.  I commited your suggestion (using GitHub's interface).

src/Hedgehog/Range.fs
```diff
@@ -8,9 +8,10 @@ open Hedgehog.Numeric
 type Size = int
 
 /// A range describes the bounds of a number to generate, which may or may not
-/// be dependent on a 'Size'.
+/// be dependent on a 'Size'.  The pair returned by the function contains the
+/// lower bound on the left and the upper bound on the right.
```
moodmosaic, line 12: We can also mention the origin `'a` which is the origin between the lower and upper bound, and that 
`(Size -> 'a * 'a)` is a function from 'Size' to bounds.  As the size goes towards 0, the values go towards the origin.

src/Hedgehog/Range.fs
TysonMN: The build failed.  I don't think that syntax is allowed.

src/Hedgehog/Range.fs
```diff
@@ -8,9 +8,10 @@ open Hedgehog.Numeric
...
 type Range<'a> =
-    | Range of ('a * (Size -> 'a * 'a))
+    | Range of (origin : 'a) * (Size -> 'a * 'a)
```
moodmosaic, line 14:
```suggestion
    | Range of origin : 'a * (Size -> 'a * 'a)
```

src/Hedgehog/Range.fs
TysonMN: The [documentation for the syntax of discriminated unions](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/discriminated-unions#syntax) does not include parentheses.  I think the parentheses in the existing code are the outermost characters in a "type".

moodmosaic: Oh, it was my parentheses.

src/Hedgehog/Range.fs
TysonMN: I added a sentence in a commit that I just force pushed.  Let me know if you would prefer to say more.

moodmosaic: It's good enough. In another PR you may also (freely) copy some of the docs from Range.hs (Haskell) since this is a port from there.

PR #236, opened 2020-11-09
Fix Tree.render bug

TysonMN:
@moodmosaic, I just realized that you introduced a bug when you changed the implementation of `Tree.renderList` in PR #229.

This PR contains two commits.  The first commit increases the cases covered by the tests by changing the trees in the tests from full binary trees to full ternary trees.  The second commit reverts the implementation of `Tree.renderList` to how [my commit](https://github.com/hedgehogqa/fsharp-hedgehog/pull/229/commits/20d328df67e3def0a0e48b3875085e33237b0aca) computed the output.

In the second commit, I tried to keep all of your copy edits.  I am just trying to get back to an implementation that passes the tests.  Feel free to change it however you want again (but this time, try to not add a bug ;) :P ).

### DIFF
--- a/src/Hedgehog/Tree.fs
+++ b/src/Hedgehog/Tree.fs
@@ -89,22 +89,20 @@ module Tree =
         Seq.filter (f << outcome) xs
         |> Seq.map (filter f)
 
-    let rec renderList (Node (x, xs00) : Tree<string>) : List<string> =
-        // zipWith (++) (hd : repeat other)
-        let shift f g xs0 =
-            match xs0 with
-            | [] ->
-                []
-            | x :: xs ->
-                (f x) :: (List.map g xs)
-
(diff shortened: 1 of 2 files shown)


moodmosaic:
Ouch! I apologize... That doesn't happen often. I should be more careful when editing.

moodmosaic:
Since you're the author of the original implementation let's merge this and we can cleanup afterwards. I assume you'd have to reopen https://github.com/hedgehogqa/fsharp-hedgehog/issues/227.

TysonMN:
No need to apologize.  It happens to all of us :)

I don't see an option to reopen #227.  Do you want it reopened?  I think keeping it closed it fine.

moodmosaic:
Yep, since we quickly merged this one we can keep it closed. Thanks Tyson!

Issue #177, opened 2019-01-08
Large for-loops in property expression causes StackOverflowException


marklam:
This is probably due to something not being tail-recursive.
A large for loop doesn't overflow the `seq` expression, but it does overflow `property`

```f#
open Hedgehog

// Learn more about F# at http://fsharp.org
// See the 'F# Tutorial' project for more help.

[<EntryPoint>]
let main argv = 
    let values1 = seq {
            for a = 0 to System.Int32.MaxValue do
                ()
            
            yield true
        }

    let values2 = property {
            for a = 0 to System.Int32.MaxValue do
                ()
            
            return true
        }

    printfn "Checking Sequence"
    printfn "Sequence %A" (values1 |> Seq.last)
    printfn "Checking Hedgehog"
    printfn "Hedgehog %A" (values2 |> Property.check)
    0 // return an integer exit code
```

The call stack contains *many* repeated sequences of
```
 	Hedgehog.dll!Hedgehog.Gen.bindRandom@40<System.Tuple<Hedgehog.Journal, Hedgehog.Result<Microsoft.FSharp.Core.Unit>>, System.Tuple<Hedgehog.Journal, Hedgehog.Result<Microsoft.FSharp.Core.Unit>>>.Invoke(Hedgehog.Seed seed0, int size = 1) Line 41	F#
 	Hedgehog.dll!Hedgehog.Random.map@39-2<Hedgehog.Tree<System.Tuple<Hedgehog.Journal, Hedgehog.Result<Microsoft.FSharp.Core.Unit>>>, Hedgehog.Tree<System.Tuple<Hedgehog.Journal, Hedgehog.Result<Microsoft.FSharp.Core.Unit>>>>.Invoke(Hedgehog.Seed seed, int size) Line 40	F#
[...]

PR #238, opened 2020-11-10
Moved functions from Random to Gen

adam-becker:
Breaking change, but alluded to in #177

I wasn't able to get the exact type mentioned in that issue, and settled for the type found in this request.

```fsharp
type Gen<'a> = Gen of (Seed -> Size -> Tree<'a>)
// Needs to be this..?
type Gen<'a> = Gen of (Seed -> Size -> Tree<Option<'a>>)
```

Feel free to close if this is too big of a change.


adam-becker:
There is a test failure because of the implementation of `Gen.sampleTree`. Any tips on what the implementation should look like are welcome.

TysonMN:
There are many tests failing.  I analyze just one of them here.

From [this code](https://github.com/hedgehogqa/fsharp-hedgehog/pull/238/files#diff-430ebaae2a51ab8bbd3f157e0645c12aaf06ce9ff4d2610f96405161befb08e5L530-L532) to [this code](https://github.com/hedgehogqa/fsharp-hedgehog/pull/238/files#diff-430ebaae2a51ab8bbd3f157e0645c12aaf06ce9ff4d2610f96405161befb08e5R542-R543) is not behavior preserving.  [`Random.replicate`](https://github.com/hedgehogqa/fsharp-hedgehog/blob/333d331283b1eee9b08d1fde27cc65391d69b917/src/Hedgehog/Random.fs#L53-L62) splits the seed so that each of the `count` values is generated using a different seed.  Your code generates each of the `count` values using the same seed.  As such, the same value is generated every time.  That is why [this test](https://github.com/hedgehogqa/fsharp-hedgehog/blob/333d331283b1eee9b08d1fde27cc65391d69b917/tests/Hedgehog.Tests/GenTests.fs#L7-L20) (for all its parameterizations) is failing.

adam-becker:
@TysonMN good catch, I used a local `mutable` (I know, ew) variable to split the seed for each iteration. I also had to add `CompiledName` attributes, but it all appears to be in order now, pending verification.

moodmosaic:
Thank you, @adam-becker. Thank you, also, @TysonMN, for reviewing this in https://github.com/hedgehogqa/fsharp-hedgehog/pull/238#issuecomment-725688603 as I haven't looked into the changes yet.

---

Question: Shall we get to this after merging the [recheck](https://github.com/hedgehogqa/fsharp-hedgehog/pulls?q=is%3Apr+recheck+author%3Aadam-becker) feature in? Then, the plan is to make a new release, and then review this.

adam-becker:
I also have a question. For this code base are we supposed to use `Gen.unsafeRun` when creating functions inside the `Gen` module?

adam-becker:
@moodmosaic @TysonMN The branch for this PR has been rebased, it should be ready for review again.

TysonMN:
Now that 0.9.0 is released, it is time to resume the review of this PR?

moodmosaic:
> Now that 0.9.0 is released, it is time to resume the review of this PR?

Yes. And perhaps we can pack this and #247 into the same (major, v10.x) release.

adam-becker:
Rebased and ready for review again.

moodmosaic (commented):
:+1:

moodmosaic:
I just did a first round of review (I've only scanned the changes) and left a few nitpicks. Another round of review (looking at the actual impl of the moved functions) should be warranted. We should be good to go then.

I've also run Script.fsx, all the stuff in there worked as usual.

src/Hedgehog/Gen.fs
```diff
@@ -1,57 +1,73 @@
...
 [<Struct>]
 type Gen<'a> =
-    | Gen of Random<Tree<'a>>
+    | Gen of (Seed -> Size -> Tree<'a>)
 
 module Gen =
 
-    let ofRandom (r : Random<Tree<'a>>) : Gen<'a> =
-        Gen r
+    let private unsafeRun (seed : Seed) (size : Size) (Gen(g) : Gen<'a>) : Tree<'a> =
```
moodmosaic, line 13:
```suggestion
    let private unsafeRun (seed : Seed) (size : Size) (Gen g : Gen<'a>) : Tree<'a> =
```

src/Hedgehog/Gen.fs
```diff
@@ -1,57 +1,73 @@
...
+        g seed size
 
-    let toRandom (Gen r : Gen<'a>) : Random<Tree<'a>> =
-        r
+    let run (seed : Seed) (size : Size) (g : Gen<'a>) : Tree<'a> =
+        unsafeRun seed (max 1 size) g
+
+    let extract (seed : Seed) (size : Size) (g : Gen<'a>) : 'a =
```
moodmosaic, line 19: Do we really want to expose this? Can't we just inline it below in that test?

src/Hedgehog/Gen.fs
```diff
@@ -1,57 +1,73 @@
...
+        run seed size g
+        |> Tree.outcome
 
     let delay (f : unit -> Gen<'a>) : Gen<'a> =
-        Random.delay (toRandom << f) |> ofRandom
+        Gen(fun seed size ->
```
moodmosaic, line 24:
```suggestion
        Gen (fun seed size ->
```

src/Hedgehog/Gen.fs
```diff
@@ -1,57 +1,73 @@
...
+            unsafeRun seed size (f ()))
 
     let tryFinally (m : Gen<'a>) (after : unit -> unit) : Gen<'a> =
-        Random.tryFinally (toRandom m) after |> ofRandom
+        Gen(fun seed size ->
```
moodmosaic, line 28:
```suggestion
        Gen (fun seed size ->
```

src/Hedgehog/Gen.fs
```diff
@@ -1,57 +1,73 @@
...
+            try
+                unsafeRun seed size m
+            finally
+                after ())
 
     let tryWith (m : Gen<'a>) (k : exn -> Gen<'a>) : Gen<'a> =
-        Random.tryWith (toRandom m) (toRandom << k) |> ofRandom
+        Gen(fun seed size ->
```
moodmosaic, line 35:
```suggestion
        Gen (fun seed size ->
```

src/Hedgehog/Gen.fs
```diff
@@ -1,57 +1,73 @@
...
+            try
+                unsafeRun seed size m
+            with
+                x -> unsafeRun seed size (k x))
 
-    let create (shrink : 'a -> seq<'a>) (random : Random<'a>) : Gen<'a> =
-        Random.map (Tree.unfold id shrink) random |> ofRandom
+    let create (shrink : 'a -> seq<'a>) (r : Seed -> Size -> 'a) : Gen<'a> =
+        Gen(fun seed size ->
```
moodmosaic, line 42:
```suggestion
        Gen (fun seed size ->
```

src/Hedgehog/Gen.fs
```diff
@@ -1,57 +1,73 @@
...
+            Tree.unfold id shrink (r seed size))
 
     let constant (x : 'a) : Gen<'a> =
-        Tree.singleton x |> Random.constant |> ofRandom
-
-    let private bindRandom (m : Random<Tree<'a>>) (k : 'a -> Random<Tree<'b>>) : Random<Tree<'b>> =
-        Hedgehog.Random <| fun seed0 size ->
-          let seed1, seed2 =
-              Seed.split seed0
+        Gen(fun _ _ ->
```
moodmosaic, line 46:
```suggestion
        Gen (fun _ _ ->
```

src/Hedgehog/Gen.fs
```diff
@@ -1,57 +1,73 @@
...
+            Tree.singleton x)
 
-          let run (seed : Seed) (random : Random<'x>) : 'x =
-              Random.run seed size random
+    let bind (m0 : Gen<'a>) (k0 : 'a -> Gen<'b>) : Gen<'b> =
+        Gen(fun seed0 size ->
```
moodmosaic, line 50:
```suggestion
        Gen (fun seed0 size ->
```

src/Hedgehog/Gen.fs
```diff
@@ -1,57 +1,73 @@
...
     let apply (gf : Gen<'a -> 'b>) (gx : Gen<'a>) : Gen<'b> =
         bind gf <| fun f ->
         bind gx <| (f >> constant)
 
-    let mapRandom (f : Random<Tree<'a>> -> Random<Tree<'b>>) (g : Gen<'a>) : Gen<'b> =
-        toRandom g |> f |> ofRandom
-
     let mapTree (f : Tree<'a> -> Tree<'b>) (g : Gen<'a>) : Gen<'b> =
-        mapRandom (Random.map f) g
+        Gen(fun seed size ->
```
moodmosaic, line 69:
```suggestion
        Gen (fun seed size ->
```

src/Hedgehog/Gen.fs
```diff
@@ -144,12 +160,14 @@ module Gen =
 
     /// Used to construct generators that depend on the size parameter.
     let sized (f : Size -> Gen<'a>) : Gen<'a> =
-        Random.sized (toRandom << f) |> ofRandom
+        Gen(fun seed size ->
```
moodmosaic, line 163:
```suggestion
        Gen (fun seed size ->
```

src/Hedgehog/Gen.fs
```diff
@@ -144,12 +160,14 @@ module Gen =
...
+            unsafeRun seed size (f size))
 
     /// Overrides the size parameter. Returns a generator which uses the
     /// given size instead of the runtime-size parameter.
     let resize (n : int) (g : Gen<'a>) : Gen<'a> =
-        mapRandom (Random.resize n) g
+        Gen(fun seed _ ->
```
moodmosaic, line 169:
```suggestion
        Gen (fun seed _ ->
```

src/Hedgehog/Gen.fs
```diff
@@ -234,42 +256,32 @@ module Gen =
...
+                constant None
             | n ->
-                let r = Random.resize (2 * k + n) r0
-                Random.bind r <| fun x ->
-                    if p (Tree.outcome x) then
-                        Tree.filter p x |> Some |> Random.constant
+                let g' = resize (2 * k + n) g
+                bind g' <| fun x ->
+                    if p x then
+                        constant (Some x)
```
moodmosaic, line 268: It looks like the actual filtering part was removed. (?)

src/Hedgehog/Gen.fs
```diff
@@ -234,42 +256,32 @@ module Gen =
...
                         tryN (k + 1) (n - 1)
 
-        Random.sized (tryN 0 << max 1)
+        sized (tryN 0 << max 1)
 
     /// Generates a value that satisfies a predicate.
     let filter (p : 'a -> bool) (g : Gen<'a>) : Gen<'a> =
         let rec loop () =
-            Random.bind (toRandom g |> tryFilterRandom p) <| function
+            bind (tryFilter p g) <| function
```
moodmosaic, line 277: So much cleaner :+1:

src/Hedgehog/Gen.fs
```diff
@@ -291,19 +303,10 @@ module Gen =
...
-        ofRandom
-        <| (Random.sized
-        <| fun size -> random {
-               let! k = Random.integral range
-               let! xs = Random.replicate k (toRandom g)
-               return Shrink.sequenceList xs
-                   |> Tree.filter (atLeast (Range.lowerBound size range))
-           })
+        bind (integral range) <| fun n ->
+            replicate n g
```
moodmosaic, line 309: Will this implementation behave the same as the old/current one?

src/Hedgehog/Gen.fs
```diff
@@ -474,10 +480,14 @@ module Gen =
     //
 
     let sampleTree (size : Size) (count : int) (g : Gen<'a>) : List<Tree<'a>> =
-        let seed = Seed.random ()
-        toRandom g
-        |> Random.replicate count
-        |> Random.run seed size
+        let action seed _ =
+            let (seed1, seed2) = Seed.split seed
+            (run seed1 size g), seed2
```
moodmosaic, line 485: Good catch on choosing `run` instead of `unsafeRun` here :+1: :100:

src/Hedgehog/Gen.fs
```diff
@@ -474,10 +480,14 @@ module Gen =
...
+
+        Seq.init count id
+        |> Seq.mapFold action (Seed.random ())
```
moodmosaic, line 488: (That's a funky way of avoiding to declare `let seed = Seed.random ()` isn't it?)

moodmosaic (commented):
Looks great. I left a comment around filter (the *diff* there isn't very clear, note that I'm reviewing this from the airport right now as I'm about to board on a plane).

src/Hedgehog/Gen.fs
adam-becker: This probably doesn't need to be exposed, I was thinking of the comonad function `extract`, which we might be able to add in earnest later. This signature doesn't follow the usual signature `extract :: m a -> a` and I don't think we have any `extend` concept to make it a proper comonad. What are your thoughts on that @moodmosaic?

src/Hedgehog/Gen.fs
adam-becker: I'm not sure if it behave the same, but I think it will behave as I would expect :sweat_smile: Basically just generate a number from the range, and generate that many elements as a list. As long as replicate has that behavior, and doesn't just duplicate the same element `n` times.

adam-becker: @moodmosaic I will add tests for this just to be sure, but I'm pretty confident as is.

src/Hedgehog/Gen.fs
adam-becker: The filter part is still there, it's this part I believe (267-270):

```fsharp
if p x then
    constant (Some x)
else
    tryN (k + 1) (n - 1)
```

src/Hedgehog/Gen.fs
TysonMN: I love @adam-becker's idea to implement the randomness as a comonad, so let's not expose any additional parts of the current randomness API lest they become breaking changes later.

src/Hedgehog/Gen.fs
TysonMN: I agree with @moodmosaic.  Most of the changes in these lines are from renames or from merging modules.  But where did the call to `Tree.filter` go?

If not calling `Tree.filter` is a bug, then we should make a test for it.

TysonMN (changes requested):

src/Hedgehog/Gen.fs
moodmosaic: >the usual signature `extract :: m a -> a`

`m` should be `w` (upside down *m*!) since it's a comonad 👀 I haven't thought about using comonads in generators and PRNGs.

src/Hedgehog/Gen.fs
moodmosaic: I agree with @TysonMN, it'd be nice with a test.

src/Hedgehog/Gen.fs
TysonMN: I agree with @adam-becker. It is hard to know if the implementations are exactly the same, but the new implementation is "the right" one (because it doesn't do a deep dive into implementation details).

However, as I pointed out in other comments, a stack overflow bug was introduced in `replicate`.

src/Hedgehog/Gen.fs
adam-becker: I see what happened here. There were a few `tryFilter*` functions, `tryFilterRandom` and `tryFilter`. When `Random` was removed, the `tryFilterRandom` function was also removed. The code is close enough that the diff shows them as the same function that changed, when this isn't the case.

src/Hedgehog/Gen.fs
```diff
@@ -1,57 +1,73 @@
...
 
-    let bind (m0 : Gen<'a>) (k0 : 'a -> Gen<'b>) : Gen<'b> =
-        bindRandom (toRandom m0) (toRandom << k0) |> ofRandom
+    let replicate (times : int) (g : Gen<'a>) : Gen<'a list> =
+        let rec loop n xs =
+            if n <= 0 then
+                constant xs
+            else
+                bind g (fun x -> loop (n - 1) (x :: xs))
+        loop times []
```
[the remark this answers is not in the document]
adam-becker: The tricky bit with making this tail recursive is that you're multiple functors deep. `List<Gen<'a>>` where `Gen<'a>` returns a `Tree<'a>` for all of the shrinks. The Haskell version implements this with the `replicateM` function, but I think Haskell's laziness makes this work. Perhaps we just need to wrap the `Gen<'a>` with `delay`? Aside from that, this will be tricky and require some thought.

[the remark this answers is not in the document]
TysonMN: Probably the simplest thing to do is keep the implementation like it was before.  We can keep this refactor in mind and come back to it later.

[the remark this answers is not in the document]
TysonMN: Actually, I am not sure it is possible to completely avoid stack overflows.

Recall that the type `'a -> 'b` is (covariant) functor in `'b`.  The `map` function for this functor is function composition (which is either `>>` or `<<` with only the order of inputs chanding).  Our `Random<'c>` type is a (covariant) functor in `'c` because it is just an wrapper around the type `'a -> 'b -> 'c`, which is also a (covarant) functor in `'c`.  We can simplify things by uncurrying to get back to a function of the form `'a -> 'b`.

 I don't know a way to implement `map` for the (covariant) functor `'a -> 'b` that won't overflow the stack.  It might be impossible.

Specifically, the following test fails.
```fsharp
[<Fact>]
let ``Does function composition overflow the stack? Answer: Yes`` () =
  let n = 100000
  let f =
    id
    |> List.replicate n
    |> List.fold (>>) id
  f ()
```

With this in mind, I think we should be more willing to allow implementations that can overflow the stack like the one this comment thread is about.  Changing the code to put more pressure on the stack will cause it to overflow sooner, but I currently think it is impossible to completely avoid in this case.

In conclusion, I suggest that we no longer hold up this PR because of this recursion that lacks tail calls.  Perhaps we can put a comment in the code to remind us of that lack of tails calls.

moodmosaic:
@TysonMN happy to merge this if you think it's in a state that can be merged. I haven't followed after https://github.com/hedgehogqa/fsharp-hedgehog/pull/238#discussion_r557428046 because I'm still (partially) out and about.

TysonMN:
> @TysonMN happy to merge this if you think it's in a state that can be merged.

I would prefer to merge PR #266 first.  I am completely convinced that removing `Random` is an improvement.

TysonMN:
> I am completely convinced that removing `Random` is an improvement.

Whoa!  I meant to say that I am _NOT_ completely convinced that removing `Random` is an improvement.

My current feeling is that we should not do this.  `Random` is complicated enough, and it is only going to get more complicated as we fix stack overflow bugs like #289 and #177.

I think the separation between `Gen` and `Random` was/is not so good.  I think this lead to the idea that things would be simplified by combining them.  Instead, my guess is that things would get better if the separation were increased.

My current thinking is that `Gen<'a>` should essentially be `Random<Tree<'a>>` and the `Gen` module should contain mostly trivial functions that delegate to the `Random` and `Tree` modules.

I haven't focused directly on the separation between `Gen` and `Random`.  Instead, I have been thinking about this indirectly while working adding features like [`ListGen.traverse`](https://github.com/hedgehogqa/fsharp-hedgehog/pull/260) and fixes for bugs like issue #289.

adam-becker:
> > I am completely convinced that removing `Random` is an improvement.
> 
> Whoa! I meant to say that I am _NOT_ completely convinced that removing `Random` is an improvement.

Can you expand on why? I see the little bits you mentioned later, but these seems more like an instinct rather than a firm judgment.

> My current feeling is that we should not do this. `Random` is complicated enough, and it is only going to get more complicated as we fix stack overflow bugs like #289 and #177.

This is why I think removing this module/type will make things simpler going forward.

> I think the separation between `Gen` and `Random` was/is not so good. I think this lead to the idea that things would be simplified by combining them. Instead, my guess is that things would get better if the separation were increased.

Not sure how much more they could be separated, they are distinct types/modules now.

> My current thinking is that `Gen<'a>` should essentially be `Random<Tree<'a>>` and the `Gen` module should contain mostly trivial functions that delegate to the `Random` and `Tree` modules.

I think `Gen<'a>` should be the central type of this project. It seems like there are two currently, and that leads to confusion/bloat/needless complexity.

TysonMN:
> > I think the separation between `Gen` and `Random` was/is not so good. I think this lead to the idea that things would be simplified by combining them. Instead, my guess is that things would get better if the separation were increased.
> 
> Not sure how much more they could be separated, they are distinct types/modules now.

The best example is probably `Gen.bindRandom`.  It duplicates code from `Random.bind`.

In my study of functional programming, I found found a common way to implement `bind` for nested modands `C<_>` and `D<_>`, which is
```fsharp
module CD =
  let bind (f: 'a -> C<D<'b>>) : C<D<'a>> -> C<D<'b>> =
    f
    |> DC.traverse
    |> C.bind
    >> C.map D.flatten
```

> I think `Gen<'a>` should be the central type of this project. It seems like there are two currently, and that leads to confusion/bloat/needless complexity.

I think that is because of duplicate code like in the above example.  In only `Random` contained code specific to `Random` and  `Gen` implemented everything for the nested monadic type `Gen<_>` without assuming the implementation details of those monadic types, then I don't think it would feel like there are two central types.

> > My current feeling is that we should not do this. `Random` is complicated enough, and it is only going to get more complicated as we fix stack overflow bugs like #289 and #177.
> 
> This is why I think removing this module/type will make things simpler going forward.

It will be easier to fix bugs specific to `Random<_>` when only working with `Random<_>` than when working with `Random<Tree<_>>`.

> > Whoa! I meant to say that I am _NOT_ completely convinced that removing `Random` is an improvement.
> 
> Can you expand on why? I see the little bits you mentioned later, but these seems more like an instinct rather than a firm judgment.

`Random<_>` and `Tree<_>` are separate monads, so they should be implemented separately.

TysonMN:
This PR makes more functions look like [`Gen.bindRandom`](https://github.com/hedgehogqa/fsharp-hedgehog/blob/fffdac4b33279db137fbce72f552feadd2cda0ac/src/Hedgehog/Gen.fs#L33-L41).  I think we should make more functions look [`Random.bind`](https://github.com/hedgehogqa/fsharp-hedgehog/blob/fffdac4b33279db137fbce72f552feadd2cda0ac/src/Hedgehog/Random.fs#L44-L50), [`Tree.bind`](https://github.com/hedgehogqa/fsharp-hedgehog/blob/fffdac4b33279db137fbce72f552feadd2cda0ac/src/Hedgehog/Tree.fs#L28-L32), and the `bind` for nested monadic types that I gave in the previous comment.

TysonMN:
Oh, by `flatten`, I mean the function that some call `join`, such as here.
https://github.com/hedgehogqa/fsharp-hedgehog/blob/00e45b73453fa3dab4bc4025f0fea5e71eab9659/src/Hedgehog/Tree.fs#L34-L35

adam-becker:
I've redone this branch entirely, `Random` is entirely gone. I also went through and streamlined `Gen` functions that were using `Random`. This extra layer of indirection is completely unnecessary, and a ton of extra allocations were removed. This should also help with Hedgehog's efficiency.

TysonMN:
I am still not convinced that this is a good idea.  I will review the current branch and see if I come to a different conclusion.

adam-becker:
Closing this for now. We can re-evaluate using the new contributing process.

PR #240, opened 2020-11-14
removed additional space in tree render #227

TysonMN:
Removes additional space character in the render of a tree as discussed in https://github.com/hedgehogqa/fsharp-hedgehog/issues/227#issuecomment-726357568.


moodmosaic (approved):
👍

Issue #242, opened 2020-11-20
Which .NET version(s) should we support?


moodmosaic:
Things are moving fast now in .NET; which version or versions should we target?

* https://dotnet.microsoft.com/download/dotnet-core
* https://dotnet.microsoft.com/download/dotnet-framework

Do we even need to target .NET Framework? Perhaps .NET (Core) is enough? We're still major version `0` in terms of flexibility.

/cc @TysonMN @cmeeren

cmeeren:
I'm no expert, but my rule of thumb is:

* Generally always target `netstandard2.0`. Then you're compatible with everything from `net461` and up (though they recommend at least `net47`-something because `netstandard2.0` support in `net461` was a bit dubious AFAIK)
* **IF** there are have different/simpler dependencies for a specific target framework (`netcoreappX.X`, `netXXX`), also target those
* **IF** .NET prior to 4.6.1 needs to be targeted (e.g. `net45`), target those - but not a priority unless someone requests it

TysonMN:
In general, the typical strategy is to support the oldest possible targets to increase the set of potential consumers.  At the same time, multitargeting allows those using newer targets to take advantage of new features, improved performance, and smaller footprints.

Pragmatically, I would drop support for an older target when the maintenance cost of keeping support becomes "too high".  I think this typically manifests as an increasing number of (possibly confusing) compiler directives to execute target-specific code or issues with automate builds.  (For example, in PR #210, when I added .NET Framework 4.5 as a target, I had intended to also test that same target, but the automated builds were not working, so tests target version 4.6.1 instead.)

moodmosaic:
Thank you, both @cmeeren and @TysonMN. I'm wondering how many out there haven't switched entirely to .NET (5, or Core Xy). 

What if we drop `netXyz` entirely and just target `netstandard2.0`. More or less, this direction would be on the same track with F# itself; starting with F# 5, FSharp.Core 5.0.0 and higher now [only targets .NET Standard 2.0](https://devblogs.microsoft.com/dotnet/announcing-f-5/#fsharp-core-now-targets-netstandard2-0-only)...

cmeeren:
I just came across [this post](https://visualstudiomagazine.com/articles/2020/09/16/net-standard-future.aspx). So it seems the following is a good way to go (if dropping support for .NET Framework before 4.7.2/4.6.1):

[...]

Issue #215, opened 2020-07-26
Automate release process


TysonMN:
This is a spinoff of https://github.com/hedgehogqa/fsharp-hedgehog/issues/199#issuecomment-663945468

> It's also a nice idea to automate the NuGet Package creation by triggering a GitHub Action by pushing a tag...

I never done this before but am excited to do so.

One thing I need from you is to encrypt your NuGet.org API key as a [GitHub secret](https://docs.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets).

I will test the YAML file before creating a PR by creating my own package on NuGet.org.

moodmosaic:
> One thing I need from you is to encrypt your NuGet.org API key as a [GitHub secret](https://docs.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets).

Sure, I'm planning to look into it during the week.

bender2k14:
No rush.  I mostly wanted to put that link in a comment for future reference.  You don't need to do anything until I have written and tested the YAML.

TysonMN:
My current idea is to have two build actions triggered by a commit into `master` with message matching the pattern `Release x.y.z`.
1. The commit with that message is tagged `vx.y.z`.  (For example, `v0.8.4` like the existing tags.)
2. The NuGet package corresponding to that commit is published as version x.y.z.

TysonMN:
Also, to do this with GitHub Actions

moodmosaic:
Sounds good 👍

adam-becker:
> My current idea is to have two build actions triggered by a commit into `master` with message matching the pattern `Release x.y.z`.
> 
>     1. The commit with that message is tagged `vx.y.z`.  (For example, `v0.8.4` like the existing tags.)
> 
>     2. The NuGet package corresponding to that commit is published as version x.y.z.

I think it would be easier/better to trigger the same actions when a tag is pushed.

Is there a pre-configured [secret][1] for the `NuGet.org` deploy key? The name of that secret will be needed to automate the deployment.

[1]: https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets

TysonMN:
Triggering on a tag push is also very good.

@moodmosaic can get the secret ser up if it is not already the case.

For testing, I have found it useful to use my fork and push packages to the local feed that each GitHub repo has.

moodmosaic:
[...]

Issue #199, opened 2020-06-30
Add some .NET Framework target


bender2k14:
Currently Hedgehog targets only .NET Standard 1.6.
https://github.com/hedgehogqa/fsharp-hedgehog/blob/6d1c5ff8581d9d1bda052e18d256e6469b134ddf/src/Hedgehog/Hedgehog.fsproj#L4

However, only targeting only this version of .NET Standard is problematic.  As Microsoft says in the [documentation for .NET Standard](https://docs.microsoft.com/en-us/dotnet/standard/net-standard),
> The versions listed here represent the rules that NuGet uses to determine whether a given .NET Standard library is applicable. While NuGet considers .NET Framework 4.6.1 as supporting .NET Standard 1.5 through 2.0, there are several issues with consuming .NET Standard libraries that were built for those versions from .NET Framework 4.6.1 projects. For .NET Framework projects that need to use such libraries, we recommend that you upgrade the project to target .NET Framework 4.7.2 or higher.

In other words, if a user of Hedgehog targets .NET Framework with a version in the interval [4.6.1, 4.7.2), then some runtime problem could occur even when the NuGet restore and build succeed.

There are at least two ways to avoid this problem: one by the consumer and one by the creator.

On the consumer side, a workaround (as included in the above quote) is to target .NET Framework 4.7.2 or greater.  However, a problem with this approach is that it increases the difference between the "production code" and automated testing of that code.  For example, if the production code targets .NET Framework 4.6.1, then it would be more useful for the automated tests to also target 4.6.1.

On the creator side, a workaround is to target both .NET Standard and .NET Framework.  For example, your test dependency [Unqoute](https://www.nuget.org/packages/Unquote/) dual targets [.NET Standard 2.0](https://github.com/SwensenSoftware/unquote/blob/c61b9661bee479e0967a449c5e26f3ab21fd6b84/Unquote.NetStandard/Unquote.NetStandard.fsproj#L4) and [.NET Framework 4.5](https://github.com/SwensenSoftware/unquote/blob/c61b9661bee479e0967a449c5e26f3ab21fd6b84/Unquote/Unquote.fsproj#L12).

[...]

PR #194, opened 2020-04-12
Make Hedgehog Fable-friendly

ThisFunctionalTom:
Hi,

I tried compiling hedghog with fable compilert and it didn't work.
Here I made some changes to make it fable friendly.

I tested it by creating a nuget package and using it with fable.mocha.

Please look over the changes if everything is OK for you.

Some notes (optional read :-)):
I had some troubles building fsharp-hedghog on windows and had to start linux docker image to build the nuget package. Is this a known problem? In linux world I had to use paket as local tool to make it work. If this is something you would like to have tell me and I will make another pull request for paket as local tool (this works only with the new .net core sdk 3.1 maybe 3.0).

PR #243, opened 2020-12-04
Normalize target frameworks

adam-becker:
I also took the liberty of removing the `Hedgehog.Fable.Tests` project since it wasn't used anywhere and was half baked. If that's not acceptable I can revert those changes, but it feels like we'd want just supplant the current tests with `Fable.Mocha`.


src/Hedgehog/Hedgehog.fsproj
```diff
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netstandard1.6;netstandard2.0;net45</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
```
TysonMN, line 4: I don't recommend this.  You are reintroducing bug #199 here.

TysonMN: #242 is a recent discuss about which versions to target.

adam-becker: I can add these back in. I think we should try to target the net standard versions unless that's impossible. The impetus for this PR was that I couldn't build the code in a Linux environment at all, which is why there were build failures on previous PRs.

Has this been mostly developed on Windows thus far?

TysonMN: Which SDK did you try?  I think the recently released .NET 5.0.100 SDK or the previous .NET Core 3.1.404 SDK should be able to build this code base on Linux including all the targets.  I could be wrong though.

adam-becker: @TysonMN This is the output I get for the dotnet version. I'm using a docker container to develop in, so I only have core.

```
$ dotnet --version
3.1.403
```

I tried building with various target frameworks but for one reason or another I wasn't able to get a compile/test run to work. Threading the needle was not possible for me.

If I'm understanding correctly, this is a deployment concern, not a development one? As in, developers of Hedgehog are more or less running with .NET Core, but we have to target older frameworks for NuGet packaging?

TysonMN: Huh.  I thought that should work.

This is work is similar to the work needed to complete issue #215.  Want to help with that?

adam-becker: Following the guidance [here][1], I can add `netstandard1.6` back in, anything lower than that fails to build.

[1]: https://docs.microsoft.com/en-us/dotnet/standard/net-standard#which-net-standard-version-to-target

adam-becker: > This is work is similar to the work needed to complete issue #215. Want to help with that?

As part of this PR, you mean?

TysonMN: For simplicity, let's have a different PR for that work

TysonMN: Working on that other issue could help us reach the same understanding about building on Linux because we can have the GitHub Action execute on a Linux machine, possibility see it fail, and look at the logs to see what didn't work.

moodmosaic (commented):
Thanks for the pull request, @adam-becker. 👍 

Hedgehog.Fable.Tests were added in https://github.com/hedgehogqa/fsharp-hedgehog/pull/194/. I'd keep them as-is.

The topic around which version(s) to target is worth discussing. But as @TysonMN wrote, there's an (ongoing) discussion in https://github.com/hedgehogqa/fsharp-hedgehog/issues/242.

---

I'd propose we reopen https://github.com/hedgehogqa/fsharp-hedgehog/issues/242, and perhaps close this pull request for now. Once the path forward is clear we can come back here.

PR #244, opened 2020-12-06
Implement C# Fluent API, along with LINQ

### LINKED ISSUE #172: Many functions that take FSharpFunc arguments are very difficult to use from C#

adam-becker:
Fixes #172 

I also followed the conversation on #183. I believe the best way to support C# is to just make a compatibility layer. I decided to target `Gen` initially to get some feedback before continuing. With that in mind, this PR is marked as a draft.

#### Pros

- Overloads in the F# code looked ugly. ie. Defining a function called `bind2` just to be able to apply the `CompiledNameAttribute` to support overloads. This is better accomplished with a `class` in F# since members allow overloading.
- `Func<_>` incompatibilities. The current approach of adding "C# friendly" methods as needed would make this awkward to use from C#. Which module would a user import or access to get the method they're looking for?
- All static methods. C# users would be more comfortable with a fluent API much like they get with LINQ now.
- CSharp support in same namespace. I think it's easier to get a clean separation by introducing a namespace _solely_ for C# support (`Hedgehog.CSharp` in this PR). Note that this isn't duplicating implementation of functions, just exposing a forwarding function for C# users.
- Order of arguments is backwards from what C# users would expect. Another reason to introduce a compatibility layer is to "fix" the order of parameters so they're more natural to C# users.
- Fable compatibility. Removing the `CompiledNameAttribute` from the core library helps Fable support because you can also remove the `#if !FABLE_COMPILER ... #end` directives. The directive can be thrown on all of the `Hedgehog.CSharp` types instead and not compile any of the compatibility layer when using Fable.
- Type parameter names. A minor win, but we can control the names of type parameters so they are idiomatic for both C# (`T`, `TResult`) and F# (`'a`, `'b`)

#### Cons

- Having to duplicate documentation comments. Because the functions are separate from the methods, the documentation comments also separate, so they'd have to be duplicated.
- Having to duplicate function declarations as methods.

Example usage:

```csharp
using Hedgehog.CSharp;

static class Example
{
    public static void GenExample()
    {
        var date = Gen.DateTime // Gen<(TimeSpan, TimeSpan)[]>
                      .Where(dateTime => dateTime > DateTime.Parse("1900-01-01"))
                      .Select(dateTime => DateTime.Now - dateTime)
                      .Tuple()
                      .Array();
    }
}
```


src/Hedgehog/Gen.fs
```diff
@@ -47,80 +38,58 @@ module Gen =
 
           Tree.bind (run seed1 m) (run seed2 << k)
 
-    [<CompiledName("Bind")>]
     let bind (m0 : Gen<'a>) (k0 : 'a -> Gen<'b>) : Gen<'b> =
         bindRandom (toRandom m0) (toRandom << k0) |> ofRandom
 
-    [<CompiledName("Apply")>]
+    let bindFlipped (k0 : 'a -> Gen<'b>) (m0 : Gen<'a>) : Gen<'b> =
```
moodmosaic, line 44: Where would we need this?

moodmosaic (commented):
The idea looks good! Though I haven't pulled the branch locally yet. It'd be nice to have feedback from @Porges, @jwChung, and @mausch (ordered by date of opening/discussion on related issue(s)) about whether the overall direction matches what they have/had in mind.

src/Hedgehog/Gen.fs
adam-becker: Good catch!

I think that's a left over from an earlier experiment, I'll remove it.

dharmaturtle:
Would it make sense to move the `__` prefixed methods in [this PR](https://github.com/hedgehogqa/fsharp-hedgehog/pull/222/files), as discussed in [this issue](https://github.com/hedgehogqa/fsharp-hedgehog/issues/156), to these new files and namespace?

Retagging @mausch.

moodmosaic:
Perhaps, yes, as that will reduce the 'noise'.

TysonMN:
I would love it if those functions where defined in a separate file.  I don't like looking at them when I am trying to think about and work with the core Hedgehog concepts.

adam-becker:
@dharmaturtle, @moodmosaic, @TysonMN

I went ahead and finished porting the rest of the functions marked with `[<CompiledName("...")>]`. All of the ported functions are marked as extension methods where appropriate. This changed the way you check properties from `Property.Check(prop)` to `prop.Check()`. Not sure if there's a way to make extension methods that can be called both ways? Maybe something magical going on with the `inline` keyword on the extensions?

Either way, the rest of the modules could be ported as well, but they didn't seem to have pre-established C# support so I left them out.

moodmosaic:
Looking forward trying this! Thank you for all the hard work 🍻

adam-becker:
I think this should be evaluated again now that `0.9.0` is released. Having a better C# story would help the library quite a bit in the adoption department.

moodmosaic:
Agreed.

adam-becker:
I've rebased on `master`, everything is building and working. I've tried a few handful of permutations of the new API, all appears to be well. Unless we want all of the modules in the F# side to be exposed, I think this is ready to merge.

src/Hedgehog/Property.fs
```diff
@@ -375,87 +356,55 @@ module Property =
...
-    // Overload for ease-of-use from C#
-#if !FABLE_COMPILER
-    [<CompiledName("Check")>]
-#endif
     let checkBool' (n : int<tests>) (g : Property<bool>) : unit =
         bind g ofBool |> check' n
 
     /// Converts a possibly-throwing function to
     /// a property by treating "no exception" as success.
-    let internal fromThrowing (f : 'a -> unit) (x : 'a) : Property<unit> =
```
TysonMN, line 417: Why the change from `internal` to `public` scope?

tests/Hedgehog.CSharp.Tests/NameTests.cs
```diff
@@ -9,9 +9,9 @@ namespace Hedgehog.CSharp.Tests
...
         private static Type[] _publicApiTypes =
-            { typeof(Gen)
-            , typeof(Range)
-            , typeof(Property)
+            { typeof(Hedgehog.CSharp.Gen)
+            , typeof(Hedgehog.CSharp.Range)
+            , typeof(Hedgehog.CSharp.Property)
             };
 
         public static IEnumerable<object[]> AllPublicMembers()
```
TysonMN, line 17: This method contains
```csharp
// Ignore static inline methods
// by convention we mark them as starting with '`'
if (mi != null && mi.Name.StartsWith("`"))
```

Are there still function names that start with that character?

TysonMN (approved):
Overall, looks great.

What is visible in C# now?  For example, Are `Gen.Digit` and `Gen.digit` both visible?

Maybe in this PR or a future one, the output of `Property.Check` can give the C# code to recheck a failed property.

TysonMN:
> What is visible in C# now? For example, Are `Gen.Digit` and `Gen.digit` both visible?

Oh, maybe you already answered this when you said

> Unless we want all of the modules in the F# side to be exposed...

How is it that the F# modules are not exposed?

adam-becker:
> > What is visible in C# now? For example, Are `Gen.Digit` and `Gen.digit` both visible?
> 
> Oh, maybe you already answered this when you said
> 
> > Unless we want all of the modules in the F# side to be exposed...
> 
> How is it that the F# modules are not exposed?

I mean that they are exposed in the C# API. Both would be visible to C# users, with the advisory that they use the `Hedgehog.CSharp` namespace.

src/Hedgehog/Property.fs
adam-becker: I figured this is useful functionality, it can be changed back if that's not the case.

tests/Hedgehog.CSharp.Tests/NameTests.cs
adam-becker: Removing that statement doesn't cause any existing tests to fail, I think it can be discarded.

TysonMN:
I think it might be better to not expose the F# modules, but I don't think this PR needs to wait for that change.

> ...with the advisory that they use the `Hedgehog.CSharp` namespace.

Assuming we don't expose the F# modules, would it be possible to expose the C# code in the same namespace, namely `Hedgehog`?

adam-becker:
> Assuming we don't expose the F# modules, would it be possible to expose the C# code in the same namespace, namely `Hedgehog`?

I think the only way to do that would be to provide a separate nuget package for C# users, and use some directives to change the functions or modules to be private? I don't think it's possible to just hide F# code from other .NET languages in the same package meant for F# consumption.

src/Hedgehog/Property.fs
TysonMN: I don't care either way.  Just wanted to draw some attention to this change.

tests/Hedgehog.CSharp.Tests/NameTests.cs
TysonMN: I will look into this tomorrow

moodmosaic:
> Maybe in this PR or a future one, the output of `Property.Check` can give the C# code to recheck a failed property.

That would be a great addition. :+1: 

>would be to provide a separate nuget package for C# user

I'm wondering how other projects deal with this :thinking:

tests/Hedgehog.CSharp.Tests/NameTests.cs
TysonMN: I have tested now as well.  I also think it can be removed.

The purpose of this test is to ensure that all the methods exposed for the purpose of C# start with a capital letter as is idiomatic for C#.  That special character was put at the beginning of inlined functions by convention. which are functions that don't have an equivalent for C#.  Now that you have separated the C# API, there are no such functions.

TysonMN:
@adam-becker, there is a conflict in `Gen.fs`.

adam-becker:
@TysonMN I saw that, I'm still at work and haven't had a chance to get to it yet.

moodmosaic (commented):
Massive piece of work :roller_coaster: :100: :rocket: :unicorn: :+1:

This makes the F# part a lot cleaner as there are no attributes related to C# and no preprocessor directives related to Fable.

(I wonder if we could use T4 to generate all the stuff in CSharp/Xyz.fs.)

@adam-becker, if you can rebase this with `master` (there's a conflict around Gen.dateTime) I believe we can consider merging this. :beers:

adam-becker:
@moodmosaic Concerning the `CompiledNameAttribute`, I think we can add them back in for now but we should probably investigate the issues that @TysonMN raised with blocking any F# functions from being used from C#. That work could also address your concerns.

src/Hedgehog/Property.fs
adam-becker: I've renamed it to `ofThrowing`, to match the existing naming.

moodmosaic:
> @moodmosaic Concerning the `CompiledNameAttribute`, I think we can add them back in for now but we should probably investigate the issues that @TysonMN raised with blocking any F# functions from being used from C#. That work could also address your concerns.

We could add back in only those I mentioned in the comments above.

adam-becker:
> We could add back in only those I mentioned in the comments above.

Is it only those? I see a few more functions in `src/Hedgehog/Range.fs` that are `inline` or the return type wouldn't be inferrable in C#. `scaleLinear`, `scaleExponential` and `constantBounded` are the ones I've noticed, basically anything that has a type `^a` I don't think would play nice in C#.

tests/Hedgehog.CSharp.Tests/NameTests.cs
adam-becker: I've left this in, and put the `CompiledName` attributes back in for a selection of the functions identified by @moodmosaic

src/Hedgehog/Range.fs
```diff
@@ -211,7 +136,6 @@ module Range =
 
     /// Construct a range which scales the bounds relative to the size
     /// parameter.
-    [<CompiledName("`LinearFrom")>]
```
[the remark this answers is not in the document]
TysonMN: Are we sure we understand PR #219 and issue #156 that it claims to have fixed?  I can call the inlined function `linearForm` from C#.  Did we think that it would throw a `NotSupportedException`?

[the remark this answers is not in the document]
adam-becker: Some of these are probably usable from C#, it's when you only have polymorphic return types that aren't tied to parameter types? That sounds at least reasonable to me, but I'm fuzzy on how `inline` actually works in compiled code.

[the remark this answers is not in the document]
TysonMN: I suggest we only mangle the names of function for which a test like the following passes.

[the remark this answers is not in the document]
TysonMN: This tests passes.  I am not sure what the difference is.  The original report in issue #156 was for a function with one argument, so the difference is not the existence or nonexistence of arguments.
```fsharp
public void CSharpCannotCall_Range_constantBounded() =>
    Assert.Throws<NotSupportedException>(() => Hedgehog.Range.constantBounded<int>());
```

dharmaturtle:
You can use the [`CompilerMessage` attribute](https://fsharp.github.io/fsharp-core-docs/reference/fsharp-core-compilermessageattribute.html) to hide code from F#. It has a property called `IsHidden`, and the docs say it

> Indicates if the construct should always be hidden in an editing environment.

However, it only hides from _F#_ editing environments. It still displays in C#... at least in Rider and Visual Studio. Dunno how to check what Ionide does.

E.g. this module....

```f#
module Asdf =
    [<CompiledName("HerpDerp"); CompilerMessage("This method is not intended for use from F#.", 10001, IsHidden=true, IsError=false)>]
    let Qwerty = ()
```

looks like this in C# VS:

image

but looks like this in F# VS:

image

To be clear, you can still call it; it just won't show up in Intellisense:

image

C# Rider:

image

F# Rider:

image

src/Hedgehog/Range.fs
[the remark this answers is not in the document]
TysonMN: I suggest we complete this PR as is (i.e. without the mangled compiled names).  Then we can try to fix #156 in a future PR.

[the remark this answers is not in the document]
adam-becker: Looking at #156, I have to ask. Does this PR address the issue there? The functions are now available for use in C# once this PR is merged. The only remaining issue is how to best guide a user towards the correct API for their environment.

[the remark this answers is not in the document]
moodmosaic: > Are we sure we understand PR #219 and issue #156 that it claims to have fixed? I can call the inlined function `linearForm` from C#. Did we think that it would throw a `NotSupportedException`?
> 
> 

The idea of that PR was to make it hard to find those names when using IntelliSense. If you *do* manage to call those members from outside you'd get a NotSupportedException as explained in https://github.com/hedgehogqa/fsharp-hedgehog/issues/156#issuecomment-702977269.

As @adam-becker [writes below](https://github.com/hedgehogqa/fsharp-hedgehog/pull/244#issuecomment-756876241) this can happen to anything that has a type `^a`.

moodmosaic:
Yep, it's tricky to actually hide those... We've tried also this in the past: https://github.com/hedgehogqa/fsharp-hedgehog/issues/156#issuecomment-703212887

src/Hedgehog/CSharp/Property.fs
```diff
@@ -0,0 +1,136 @@
...
+type Property private () =
+
+    static member Failure : Property<unit> =
+        Property.failure
+
+    static member Discard : Property<unit> =
+        Property.discard
+
+    static member Success (value : 'T) : Property<'T> =
+        Property.Success(value)
```
moodmosaic, line 20: Shouldn't this be `success` (with lower `s`)?

moodmosaic, line 20:
```suggestion
        Property.success(value)
```

tests/Hedgehog.CSharp.Tests/LinqTests.cs
```diff
@@ -1,7 +1,8 @@
 ﻿using Xunit;
 
-// Import Check and ForAll:
-using static Hedgehog.Property;
+// Import ForAll:
+using Hedgehog.CSharp;
```
moodmosaic, line 4:
```suggestion

```

I don't think we need to have this using directive anymore. The other below also exposes `ForAll`.

moodmosaic (commented):
@TysonMN's suggestion to tackle #156 separately makes sense, given the size and context of this pull request.

However, since now it's going to be *explicit* whether you're using Hedgehog vs Hedgehog.CSharp, perhaps there's no need to think about #156.

The idea could be that
* *if* you're using Hedgehog.CSharp from C# code, nothing you pick up via IntelliSense should throw a NotSupportException
* *if* you're using stuff from Hedgehog (not Hedgehog.CSharp) from C# code, then anything that uses a statically resolved type parameter *could* throw...

---

Ready to merge at this point, once those 2 small suggestions I just left are resolved (one in CSharp.Property.fs and one in LinqTests.fs).

src/Hedgehog/Range.fs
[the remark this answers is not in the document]
moodmosaic: > Perhaps we can leave that `CompiledName` in, in order to prevent users from calling it from C#? See #219.

No need to put those attributes back, at least in this PR, after all the discussion. Great feedback, @dharmaturtle, @TysonMN, @adam-becker.

src/Hedgehog/CSharp/Property.fs
TysonMN: Yes, definitely needs to be a lowercase `s`.  Otherwise that method is calling itself.

I was thinking of suggesting that all these implementations be fully qualified.  Then this would have been easier to spot since
```fsharp
Hedgehog.Property.Success value
```
doesn't compile.

However, I was surprised to discover that
```fsharp
Hedgehog.Property.success value
```
doesn't compile either.  Even though Intellisense suggests completing `Hedgehog.Property.s` with the above, it doesn't compile because `Hedgehog.Property` resolves to the type `Property` instead of the module `Property`.

adam-becker:
@moodmosaic Ready to merge!

moodmosaic:
Great! Sometimes we can merge via a merge commit, and sometimes via squash and merge. What would you prefer?

adam-becker:
> Great! Sometimes we can merge via a merge commit, and sometimes via squash and merge. What would you prefer?

Either is fine with me

src/Hedgehog/CSharp/Property.fs
```diff
@@ -17,14 +17,17 @@ type Property private () =
...
 
     static member FromBool (value : bool) : Property<unit> =
         Property.ofBool(value)
 
     static member FromGen (gen : Gen<Journal * Result<'T>>) : Property<'T> =
         Property.ofGen gen
 
+    static member FromResult (result : Result<'T>) : Property<'T> =
+        Property.ofResult result
+
```
moodmosaic, line 30: Yeah, if we could generate all src/Hedgehog/CSharp/*.fs files via T4 it'd be awesome (T4 is a pain to work with OTOH). :+1:

moodmosaic:
What a milestone :rocket: :full_moon: Thank you @adam-becker, and also @TysonMN and @dharmaturtle for the feedback. :+1:

PR #247, opened 2020-12-08
Split Property.fs across multiple files

### LINKED ISSUE #181: Hedgehog.ResultModule.map is conflicting with FSharp.Core.ResultModule.map
I have a property-based test written in xUnit:

```fsharp
open Hedgehog

[<Fact>]
let ``My Test`` () =
    Property.check <| property {
        // Does not work because it is compiler is resolving to map: f ('a -> 'b) -> Hedgehog.Result<'a> -> Hedgehog.Result<'b>
        // when I actually want map: f ('a -> 'b) -> FSharp.Core.Result<'a, 'error> -> FSharp.Core.Result<'b, 'error>
        let myList =
            Ok ["foo"; "bar"]
            |> Result.map Set.ofList 

    }
```

Which is quite frustrating as now I have to always fully qualify to `FSharp.Core.Result.map`. Is it possible to rename the `Hedgehog.Result` type?

### LINKED ISSUE #246: Property.fs is getting messy
I think it would be worthwhile to move the types in `Property.fs` into their own files along with their supporting modules.

adam-becker:
Fixes #246.

Also, while I moved `Result<'a>` I opted to rename it to `Outcome<'a>`, therefor this also fixes #181.

### DIFF
--- a/src/Hedgehog/Journal.fs
+++ b/src/Hedgehog/Journal.fs
@@ -5,20 +5,31 @@ type Journal =
     | Journal of seq<unit -> string>
 
 module Journal =
-    let ofList (xs : seq<unit -> string>) : Journal =
-        Journal xs
 
-    let toList (Journal xs : Journal) : List<string> =
-        Seq.toList <| Seq.map (fun f -> f ()) xs
+    /// Creates a journal from a sequence of entries.
+    let ofSeq (entries : seq<unit -> string>) : Journal =
+        Journal entries
 
+    /// Evaluates a single entry, returning it's message.
+    let private evalEntry (f : unit -> string) : string =
+        f()
+
+    /// Evaluates all entries in the journal, returning their messages.
+    let eval (Journal entries : Journal) : seq<string> =
+        Seq.map evalEntry entries
+
+    /// Represents a journal with no entries.
     let empty : Journal =
-        Seq.empty |> ofList
+        ofSeq []
 
-    let singleton (x : string) : Journal =
-        Seq.singleton (fun () -> x) |> ofList
+    /// Creates a single entry journal from a given message.
+    let singletonMessage (message : string) : Journal =
+        ofSeq [ fun () -> message ]
 
-    let delayedSingleton (x : unit -> string) : Journal =
-        Seq.singleton x |> ofList
+    /// Creates a single entry journal from a given entry.
+    let singleton (entry : unit -> string) : Journal =
+        ofSeq [ entry ]
 
+    /// Creates a journal composed of entries from two journals.
     let append (Journal xs) (Journal ys) : Journal =
-        Seq.append xs ys |> ofList
+        ofSeq <| Seq.append xs ys
--- a/src/Hedgehog/Journal.fs
+++ b/src/Hedgehog/Journal.fs
@@ -1,0 +1,24 @@
+namespace Hedgehog
+
+[<Struct>]
+type Journal =
+    | Journal of seq<unit -> string>
+
+module Journal =
+    let ofList (xs : seq<unit -> string>) : Journal =
+        Journal xs
+
+    let toList (Journal xs : Journal) : List<string> =
+        Seq.toList <| Seq.map (fun f -> f ()) xs
+
+    let empty : Journal =
+        Seq.empty |> ofList
+
+    let singleton (x : string) : Journal =
+        Seq.singleton (fun () -> x) |> ofList
+
+    let delayedSingleton (x : unit -> string) : Journal =
+        Seq.singleton x |> ofList
+
+    let append (Journal xs) (Journal ys) : Journal =
+        Seq.append xs ys |> ofList
--- a/src/Hedgehog/Journal.fs
+++ b/src/Hedgehog/Journal.fs
@@ -32,4 +32,5 @@ module Journal =
 
     /// Creates a journal composed of entries from two journals.
     let append (Journal xs) (Journal ys) : Journal =
-        ofSeq <| Seq.append xs ys
+        Seq.append xs ys
+        |> ofSeq
--- a/src/Hedgehog/Outcome.fs
+++ b/src/Hedgehog/Outcome.fs
@@ -19,9 +19,9 @@ module Outcome =
     [<CompiledName("Map")>]
     let map (f : 'a -> 'b) (result : Outcome<'a>) : Outcome<'b> =
         cata result
-            <| always Failure
-            <| always Discard
-            <| (f >> Success)
+            (always Failure)
+            (always Discard)
+            (f >> Success)
 
     [<CompiledName("Filter")>]
     let filter (f : 'a -> bool) (result : Outcome<'a>) : Outcome<'a> =
@@ -32,13 +32,13 @@ module Outcome =
                 Discard
 
         cata result
-            <| always Failure
-            <| always Discard
-            <| successOrDiscard
+            (always Failure)
+            (always Discard)
+            successOrDiscard
 
     [<CompiledName("IsFailure")>]
     let isFailure (result : Outcome<'a>) : bool =
         cata result
-            <| always true
-            <| always false
-            <| always false
+            (always true)
+            (always false)
+            (always false)
--- a/src/Hedgehog/Outcome.fs
+++ b/src/Hedgehog/Outcome.fs
@@ -1,0 +1,40 @@
+namespace Hedgehog
+
+type Outcome<'a> =
+    | Failure
+    | Discard
+    | Success of 'a
+
+module Outcome =
+    [<CompiledName("Map")>]
+    let map (f : 'a -> 'b) (result : Outcome<'a>) : Outcome<'b> =
+        match result with
+        | Failure ->
+            Failure
+        | Discard ->
+            Discard
+        | Success x ->
+            Success (f x)
+
+    [<CompiledName("Filter")>]
+    let filter (f : 'a -> bool) (result : Outcome<'a>) : Outcome<'a> =
+        match result with
+        | Failure ->
+            Failure
+        | Discard ->
+            Discard
+        | Success x ->
+            if f x then
+              Success x
+            else
+              Discard
+
+    [<CompiledName("IsFailure")>]
+    let isFailure (result : Outcome<'a>) : bool =
+        match result with
+        | Failure ->
+            true
+        | Discard ->
+            false
+        | Success _ ->
+            false
--- a/src/Hedgehog/Outcome.fs
+++ b/src/Hedgehog/Outcome.fs
@@ -6,35 +6,39 @@ type Outcome<'a> =
     | Success of 'a
 
 module Outcome =
-    [<CompiledName("Map")>]
-    let map (f : 'a -> 'b) (result : Outcome<'a>) : Outcome<'b> =
+
+    let cata result failure discard success =
         match result with
         | Failure ->
-            Failure
+            failure()
         | Discard ->
-            Discard
-        | Success x ->
-            Success (f x)
+            discard()
+        | Success(x) ->
+            success(x)
+
+    [<CompiledName("Map")>]
+    let map (f : 'a -> 'b) (result : Outcome<'a>) : Outcome<'b> =
+        cata result
+            <| always Failure
+            <| always Discard
+            <| (f >> Success)
 
     [<CompiledName("Filter")>]
     let filter (f : 'a -> bool) (result : Outcome<'a>) : Outcome<'a> =
-        match result with
-        | Failure ->
-            Failure
-        | Discard ->
-            Discard
-        | Success x ->
+        let successOrDiscard x =
             if f x then
-              Success x
+                Success(x)
             else
-              Discard
+                Discard
+
+        cata result
+            <| always Failure
+            <| always Discard
+            <| successOrDiscard
 
     [<CompiledName("IsFailure")>]
     let isFailure (result : Outcome<'a>) : bool =
-        match result with
-        | Failure ->
-            true
-        | Discard ->
-            false
-        | Success _ ->
-            false
+        cata result
+            <| always true
+            <| always false
+            <| always false
--- a/src/Hedgehog/Property.fs
+++ b/src/Hedgehog/Property.fs
@@ -56,7 +56,7 @@ module Property =
             failure
 
     let counterexample (msg : unit -> string) : Property<unit> =
-        Gen.constant (Journal.delayedSingleton msg, Success ()) |> ofGen
+        Gen.constant (Journal.singleton msg, Success ()) |> ofGen
 
     let private mapGen
             (f : Gen<Journal * Outcome<'a>> -> Gen<Journal * Outcome<'b>>)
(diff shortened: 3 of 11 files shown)


TysonMN:
I love your intensity @adam-becker!  Keep on proposing changes like this that you think are an improvement :D

src/Hedgehog/Shrink.fs
```diff
@@ -50,7 +48,7 @@ module Shrink =
     /// Shrink a list by edging towards the empty list.
     /// Note we always try the empty list first, as that is the optimal shrink.
     let list (xs : List<'a>) : seq<List<'a>> =
-        Seq.concat <| Seq.map (fun k -> removes k xs) (halves <| List.length xs)
+        Seq.collect (fun k -> removes k xs) (halves <| List.length xs)
```
moodmosaic, line 51: Good catch! :beers: (...wish there was an emoji with *3* beers! Fixed! :beers: + :beer:, so @TysonMN can join.)

moodmosaic (commented):
Looking great :+1: If @TysonMN also approves, we can merge it. Shall we wait for v0.9.0 first, or merge it and then do v0.9.0?

src/Hedgehog/Shrink.fs
TysonMN: Ah, the ol' `map f >> flaten = bind f`.

src/Hedgehog/Journal.fs
```diff
@@ -0,0 +1,24 @@
+namespace Hedgehog
+
+[<Struct>]
+type Journal =
+    | Journal of seq<unit -> string>
+
+module Journal =
+    let ofList (xs : seq<unit -> string>) : Journal =
```
TysonMN, line 8: How about calling this `ofSeq`?

src/Hedgehog/Journal.fs
```diff
@@ -0,0 +1,24 @@
...
+        Seq.toList <| Seq.map (fun f -> f ()) xs
+
+    let empty : Journal =
+        Seq.empty |> ofList
+
+    let singleton (x : string) : Journal =
+        Seq.singleton (fun () -> x) |> ofList
+
+    let delayedSingleton (x : unit -> string) : Journal =
+        Seq.singleton x |> ofList
```
TysonMN, line 21: Since `Journal` wraps many items of type `unit -> string`, so I think it makes more sense to give the preferred name of `singleton` to the function that accepts `unit -> string` and call the other function `eagerSingleton`.

src/Hedgehog/Outcome.fs
```diff
@@ -0,0 +1,40 @@
+namespace Hedgehog
+
+type Outcome<'a> =
+    | Failure
+    | Discard
+    | Success of 'a
+
+module Outcome =
```
TysonMN, line 8: Very optional suggestion.  Instead of matching three times, the duplicate code could be extracted into a function called `cata` function [like Mark Seemann does](https://blog.ploeh.dk/2019/04/29/catamorphisms/).

src/Hedgehog/Report.fs
```diff
@@ -0,0 +1,106 @@
...
+    let private renderAndShrinks : int<shrinks> -> string = function
+        | 0<shrinks> ->
+            ""
+        | 1<shrinks> ->
+            " and 1 shrink"
+        | n ->
+            sprintf " and %d shrinks" n
+
+    let private append (sb : StringBuilder) (msg : string) : unit =
+        sb.AppendLine msg |> ignore
```
TysonMN, line 59: How about calling this `appendLine` to match the wrapped method?

src/Hedgehog/Property.fs
```diff
@@ -278,25 +72,25 @@ module Property =
...
+            (k : 'a -> Gen<Journal * Outcome<'b>>) : Gen<Journal * Outcome<'b>> =
         Gen.bind m <| fun (journal, result) ->
             match result with
             | Failure ->
                 Gen.constant (journal, Failure)
             | Discard ->
                 Gen.constant (journal, Discard)
             | Success x ->
-                Gen.map (first (Journal.append journal)) (k x)
+                Gen.map (Tuple.mapFirst (Journal.append journal)) (k x)
```
TysonMN, line 93: When I have nested functors like this, I like to define a module for them.
```fsharp
module GenTuple =
  let mapFirst f = f |> Tuple.mapFirst |> Gen.map
```

This helps to reduce the number of matches pars of parentheses, which [I prefer to minimize](https://blog.ploeh.dk/2020/10/19/monomorphic-functors/#bb0bb9d1bf094c1897f9dd42fb49712c).

src/Hedgehog/Journal.fs
```diff
@@ -0,0 +1,24 @@
...
+
+[<Struct>]
+type Journal =
+    | Journal of seq<unit -> string>
+
+module Journal =
+    let ofList (xs : seq<unit -> string>) : Journal =
+        Journal xs
+
+    let toList (Journal xs : Journal) : List<string> =
```
TysonMN, line 11: How about calling this `eval` and returning `seq` instead of `list`?

TysonMN (approved):
This is great.  I highly prefer the new organization.  I think this can be merged as is.

I left optional comments.  I think all of them are independent of this reorganization.

src/Hedgehog/Shrink.fs
adam-becker: This was suggested to me by `ionide`, so I think the kudos should go to the analyzer :wink:

I don't know if this is more efficient, but it's definitely simpler.

adam-becker:
> Looking great +1 If @TysonMN also approves, we can merge it. Shall we wait for v0.9.0 first, or merge it and then do v0.9.0?

With all of the suggestions from @TysonMN it might make sense to apply all of them and release 0.9.0 as a breaking version. Although anything before 1.0.0 is free to break things according to [SemVer][1] guidelines.

That said, this library does have users and we should probably stuff as many breaking changes into a single version bump, instead of breaking constantly.

[1]: https://semver.org/#spec-item-5

src/Hedgehog/Shrink.fs
TysonMN: When I implement my own monads, sometimes I find it easier to directly implement `flatten` and then implement `bind f` as `map f >> flatten` and other times I find it easier to implement `bind` and then implement `flatten` as `bind id`.  Typically though I would expect `bind f` to be more efficient than `map f >> flatten`.

src/Hedgehog/Journal.fs
TysonMN: ...or maybe `literalSingleton` or `constantSingleton`.  Something like that.

src/Hedgehog/Journal.fs
adam-becker: Done

src/Hedgehog/Property.fs
adam-becker: Done

src/Hedgehog/Report.fs
adam-becker: Done, also renamed `renderf` to `appendLinef`

src/Hedgehog/Outcome.fs
adam-becker: Done

src/Hedgehog/Journal.fs
adam-becker: Done

src/Hedgehog/Journal.fs
adam-becker: Done

moodmosaic:
>@moodmosaic Do we want to make a project for the 0.9.0 release and quantify some breaking work we want to do?

This is excellent work! I think we should cut a release first [what's in `master`](https://github.com/hedgehogqa/fsharp-hedgehog/compare/v0.8.3...d54a06726bc64e792cbfff29eba951468d7931c5) 'already' and then release this as 0.10.x. We have quite a lot of stuff to be released already so I think we should let our regular users digest these changes in chunks rather than all at once.

src/Hedgehog/Tuple.fs
```diff
@@ -0,0 +1,19 @@
...
+module private Tuple =
+
+    let mapFirst (f : 'a -> 'c) (x : 'a, y : 'b) : 'c * 'b =
+        f x, y
+
+    let mapSecond (f : 'b -> 'c) (x : 'a, y : 'b) : 'a * 'c =
+        x, f y
+
+
+module private GenTuple =
```
TysonMN, line 12: As I said in https://github.com/hedgehogqa/fsharp-hedgehog/pull/260#discussion_r551046128, I think a better name for this module is `GenPair`.

src/Hedgehog/Tuple.fs
```diff
@@ -0,0 +1,19 @@
+namespace Hedgehog
+
+module private Tuple =
+
+    let mapFirst (f : 'a -> 'c) (x : 'a, y : 'b) : 'c * 'b =
+        f x, y
+
+    let mapSecond (f : 'b -> 'c) (x : 'a, y : 'b) : 'a * 'c =
```
TysonMN, line 8: I prefer calling these two functions `map1` and `map2` because then their names are more symmetric at a syntactic level.  I can understand though if others find these names too terse.

moodmosaic: What's the equivalent naming in Haskell/Hoogle?

TysonMN: The F# pair type `'a * 'b` would be an instance of the Haskell type class [Data.Bifunctor](https://hackage.haskell.org/package/bifunctors-5/docs/Data-Bifunctor.html), so the equivalent naming in Haskell would be `first` and `second`, which is neither `mapFirst` and `mapSecond` nor `map1` and `map2`.

I think `first` and `second` are bad names.  It is generally impossible to [get the value out of your ~monad~ _functor_](https://blog.ploeh.dk/2019/02/04/how-to-get-the-value-out-of-the-monad/), but it is possible with the pair type `'a * 'b`, and in this case, I would expect `first` and `second` to have that behavior.  More specifically, I expect their types to be `first : 'a * 'b -> 'a` and `second : 'a * 'b -> 'b`.  In fact, F# has functions with those exact types and abbreviations of those names as `fst` and `snd`.

For that reason, I do not recommend following the Haskell naming convention here.  I am fine with either `mapFirst` and `mapSecond` or `map1` and `map2` but with a preference for the latter.

What about calling them `mapFst` and `mapSnd`?  That satisfies my preference for the names to be my syntactically similar (i.e. they are the same length) while also being consistent with existing functions in the core F# library.

dharmaturtle: I also like `map1` and `map2`. The F# core library can't call them `1` and `2` because they're illegal identifiers, but in context `map1` and `map2` are very readable (c.f. `mapFst`, `mapSnd`).

moodmosaic: I'm not sure... Perhaps, in this case, `map1` and `map2` clash with function arity(?) ― Also, nothing actually stops us from having those inlined as long as they don't violate the rule of three.

In fact:
*"[..]wait, until you have more facts available to you. You don't have to take the rule literally either. You can wait until you have four, five, or six examples of the duplication, if the rate of change is low."*
-- https://blog.ploeh.dk/2014/08/07/why-dry/

TysonMN: > Perhaps, in this case, `map1` and `map2` clash with function arity(?)

That is possible.

> nothing actually stops us from having those inlined

What do you mean by "inlined"?

adam-becker: `map1`, `map2` etc for this use conflicts directly with the standard library. See `List.map2`, `List.map3`, where the number is a reference to the arity of the function being mapped.

`mapFst` or `mapSnd` fit in better with the standard library, which provides a `fst` and `snd` function, and also a plethora of `map` functions. Though lately I've taken to providing a sub module to group these things and just keeping the simple name. For example `Result.Error.map` vs `Result.mapError`.

TysonMN: Good points.

> `mapFst` or `mapSnd` fit in better with the standard library, which provides a `fst` and `snd` function, and also a plethora of `map` functions. Though lately I've taken to providing a sub module to group these things and just keeping the simple name. For example `Result.Error.map` vs `Result.mapError`.

Oh, yes.  I too have been writing more code like `Result.Error.map`.  I put all the nouns first and then the name of the function gets to be a single verb.

adam-becker: Renamed to `mapFst`, `mapSnd`.

src/Hedgehog/Tuple.fs
adam-becker: I'm going to leave these as is, I don't think they hurt anything.

TysonMN:
I just reviewed again.  Looks good to me.

src/Hedgehog/Tuple.fs
TysonMN: The PR has already been completed, so this is a bit late.  Nonetheless, another point of comparison is with [FSharpPlus, which calls these `mapItem1` and `mapItem2`.](https://github.com/fsprojects/FSharpPlus/blob/a36d35a3470a0a59b176e7527f426f9014333c1d/src/FSharpPlus/Extensions/Tuple.fs)

src/Hedgehog/Tuple.fs
TysonMN: The PR has already been completed, so this is a bit late.  Nonetheless, another point of comparison is with [FSharpPlus, which calls this `Tuple2`.](https://github.com/fsprojects/FSharpPlus/blob/a36d35a3470a0a59b176e7527f426f9014333c1d/src/FSharpPlus/Extensions/Tuple.fs)

moodmosaic: There's still room for changes, if you want. Until we release the changes. (And even after that, theoretically, we can rename those, as per SemVer.)

src/Hedgehog/Util.fs
```diff
@@ -0,0 +1,5 @@
+[<AutoOpen>]
+module private Hedgehog.Util
+
+let inline always (a : 'a) (_ : 'b) : 'a =
```
moodmosaic, line 4: Could we move this to the newly added `AutoOpen` module (#279)?

moodmosaic:
Should be ready to merge after resolving the conflicts. I've also left a small comment in https://github.com/hedgehogqa/fsharp-hedgehog/pull/247#discussion_r560524975.

moodmosaic (commented):
Ready to merge after resolving those conflicts. Massive piece of work, @adam-becker, fantastic feedback, @TysonMN.

adam-becker:
The test failed randomly, due to exhausting its discards. I'll make an issue to fix flaky tests.

moodmosaic:
One last thing, before we merge, if you open Script.fsx you'll notice that it'll error about some missing references. All we have to do is just add them:

```diff
@@ -1,13 +1,18 @@
 #if INTERACTIVE
-#load "Numeric.fs"
+#load "AutoOpen.fs"
+      "Numeric.fs"
       "Seed.fs"
       "Tree.fs"
       "Range.fs"
       "Random.fs"
       "Shrink.fs"
       "Gen.fs"
+      "Journal.fs"
+      "Tuple.fs"
+      "Outcome.fs"
+      "Report.fs"
       "Property.fs"
 #endif
 
 open Hedgehog
 open System
```

I can make that change myself of course, if you don't have time to do it.

adam-becker:
@moodmosaic Ready to go on this one.

moodmosaic:
:tada:

PR #250, opened 2020-12-11
Add Gen.single and Gen.decimal

### LINKED ISSUE #249: Generate single/float32
I noticed the auto-generator in Hegehog.Experimental failing for one of my record types. It was because it couldn't generate a `single` (`float32`) property. When I tried to add support for it to the auto-generator, I discovered that Hedgehog has no `Gen.single`. Can this be added?

In the meantine, is there any downside to doing this in the auto-generator? (Or if you can implement `Gen.single` quickly, I can wait.)

```f#
Gen.float (Range.exponentialFrom 0. (float Single.MinValue) (float Single.MaxValue)) |> Gen.map single
```

cmeeren:
Fixes #249

### DIFF
--- a/src/Hedgehog/Gen.fs
+++ b/src/Hedgehog/Gen.fs
@@ -494,6 +494,16 @@ module Gen =
     let float (range : Range<float>) : Gen<float> =
         (double range) |> map float
 
+    /// Generates a random 32-bit floating point number.
+    [<CompiledName("Single")>]
+    let single (range : Range<single>) : Gen<single> =
+      double (Range.map ExtraTopLevelOperators.double range) |> map single
+
+    /// Generates a random decimal floating-point number.
+    [<CompiledName("Decimal")>]
+    let decimal (range : Range<decimal>) : Gen<decimal> =
+      double (Range.map ExtraTopLevelOperators.double range) |> map decimal
+
     //
     // Combinators - Constructed
     //


src/Hedgehog/Gen.fs
```diff
@@ -3,12 +3,17 @@
...
+
 /// A generator for values and shrink trees of type 'a.
 [<Struct>]
 type Gen<'a> =
     | Gen of Random<Tree<'a>>
 
 module Gen =
+
+    // Needed because 'double' is shadowed later
+    let inline private fsharpDouble x = double x
```
moodmosaic, line 15: Can't we just use the full qualified name instead? If we have to do that in more than three places, perhaps we can declare an inline function as the one above. Right now it's been used only in two places.

src/Hedgehog/Gen.fs
```diff
@@ -3,12 +3,17 @@
 open System
 open Hedgehog.Numeric
 
+
```
moodmosaic, line 6: What's the purpose of that extra line?

moodmosaic (commented):
Looking good. Left a few comments.

src/Hedgehog/Gen.fs
cmeeren: Sure. I didn't find where it was (I looked in `FSharp.Core.Operators`, but it wasn't there), and decided that it wasn't worth the trouble. Do you know the fully qualified name?

src/Hedgehog/Gen.fs
cmeeren: Absolutely none. I'll remove that rogue edit. It was caused by my adding `fsharpDouble` there first before realizing it had to be moved, and then I forgot to check the diff before I committed.

src/Hedgehog/Gen.fs
cmeeren: Nevermind, I found it.

cmeeren:
Now the diff is cleaner.

moodmosaic (approved):
:+1:

PR #251, opened 2020-12-11
Fix documentation file not being added to NuGet

cmeeren:
Currently, the documentation file is either not generated or not added to the nupkg. This PR fixes that, meaning that Intellisense will actually show the function docs.

### DIFF
--- a/src/Hedgehog/Hedgehog.fsproj
+++ b/src/Hedgehog/Hedgehog.fsproj
@@ -6,7 +6,7 @@
     <Description>Release with confidence.</Description>
     <Authors>Jacob Stanley;Nikos Baxevanis</Authors>
     <Copyright>Copyright © 2017</Copyright>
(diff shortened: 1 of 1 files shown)


src/Hedgehog/Hedgehog.fsproj
```diff
@@ -6,7 +6,7 @@
     <Description>Release with confidence.</Description>
     <Authors>Jacob Stanley;Nikos Baxevanis</Authors>
     <Copyright>Copyright © 2017</Copyright>
-    <DocumentationFile>bin\$(Configuration)\$(TargetFramework)\Hedgehog.XML</DocumentationFile>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
```
moodmosaic, line 9: Very nice. This looks like a [modern alternative to the old school one](https://docs.microsoft.com/en-us/dotnet/csharp/codedoc). :100:

moodmosaic (approved):
:+1:

PR #252, opened 2020-12-11
Improve and extend DateTime/DateTimeOffset generation

cmeeren:
This PR consists of these changes, each in a separate commit:
* Improve correctness of Gen.dateTime docs (`DateTime` is decidedly not an instant in time; `DateTimeOffset` is). This change was also done for consistency with the new functions added in later commits.
* Added `Gen.dateTimeOffset`. It's a "primitive" that is non-trivial to generate (care must be taken to avoid overflows due to adding offsets around the min/max values) and should be in fsharp-hedgehog
* Added ranged variants of `Gen.dateTime` and `Gen.dateTimeOffset`, allowing users to specify their own date ranges (I have myself come across the need to limit the date ranges due to complicated time zone stuff some country did a hundred years ago, which is not interesting to account for in my domain)
* Implemented `dateTime` and `dateTimeOffset` in terms of their ranged variants

There is one other change I'd like to make, but for some reason it won't compile: I'd like to change the default range of `dateTime` and `dateTimeOffset` from `constant` to either `linear` or `exponential`. I don't see a reason why it should be size-independent. But when I tried to change it, I get compile errors:

image

### DIFF
--- a/src/Hedgehog/Gen.fs
+++ b/src/Hedgehog/Gen.fs
@@ -537,7 +537,7 @@ module Gen =
               min 
                 (14L * 60L)
                 ((ticks - DateTimeOffset.MinValue.Ticks) / TimeSpan.TicksPerMinute)
-            let! offsetMinutes = int (Range.exponentialFrom 0 (Operators.int minOffsetMinutes) (Operators.int maxOffsetMinutes))
+            let! offsetMinutes = int (Range.linearFrom 0 (Operators.int minOffsetMinutes) (Operators.int maxOffsetMinutes))
             return System.DateTimeOffset(ticks, TimeSpan.FromMinutes (Operators.float offsetMinutes))
         }
 
--- a/src/Hedgehog/Gen.fs
+++ b/src/Hedgehog/Gen.fs
@@ -516,6 +516,11 @@ module Gen =
     }
 
     /// Generates a random DateTime using the specified range.
+    /// For example:
+    ///   let range =
+    ///      Range.constantFrom
+    ///          (System.DateTime (2000, 1, 1)) System.DateTime.MinValue System.DateTime.MaxValue
+    ///   Gen.dateTime range
     [<CompiledName("DateTime")>]
     let dateTime (range : Range<DateTime>) : Gen<System.DateTime> =
         gen {
--- a/src/Hedgehog/Gen.fs
+++ b/src/Hedgehog/Gen.fs
@@ -530,6 +530,24 @@ module Gen =
             return System.DateTime ticks
         }
 
+    /// Generates a random DateTimeOffset.
+    [<CompiledName("DateTimeOffset")>]
+    let dateTimeOffset : Gen<System.DateTimeOffset> =
+        let minTicks =
+            System.DateTimeOffset.MinValue.Ticks
+        let maxTicks =
+            System.DateTimeOffset.MaxValue.Ticks
+        gen {
+            let! ticks =
+                Range.constantFrom
+                    (System.DateTimeOffset (2000, 1, 1, 0, 0, 0, TimeSpan.Zero)).Ticks minTicks maxTicks
+                |> integral
+            let minOffsetMinutes = max (-14 * 60) (Operators.int ((maxTicks - ticks) / TimeSpan.TicksPerMinute) * -1)
+            let maxOffsetMinutes = min (14 * 60) (Operators.int ((ticks - minTicks) / TimeSpan.TicksPerMinute))
+            let! offsetMinutes = int (Range.exponentialFrom 0 minOffsetMinutes maxOffsetMinutes)
+            return System.DateTimeOffset(ticks, TimeSpan.FromMinutes (Operators.float offsetMinutes))
+        }
+
     //
     // Sampling
     //
--- a/src/Hedgehog/Gen.fs
+++ b/src/Hedgehog/Gen.fs
@@ -515,7 +515,7 @@ module Gen =
         return System.Guid bs
     }
 
-    /// Generates a random instant in time expressed as a date and time of day.
+    /// Generates a random DateTime.
     [<CompiledName("DateTime")>]
     let dateTime : Gen<System.DateTime> =
         let minTicks =
--- a/src/Hedgehog/Gen.fs
+++ b/src/Hedgehog/Gen.fs
@@ -515,6 +515,34 @@ module Gen =
         return System.Guid bs
     }
 
+    /// Generates a random DateTime using the specified range.
+    [<CompiledName("DateTimeRanged")>]
+    let dateTimeRanged (range : Range<DateTime>) : Gen<System.DateTime> =
+        let minTicks =
+            System.DateTime.MinValue.Ticks
+        let maxTicks =
+            System.DateTime.MaxValue.Ticks
+        gen {
+            let! ticks = range |> Range.map (fun dt -> dt.Ticks) |> integral
+            return System.DateTime ticks
+        }
+
+    /// Generates a random DateTimeOffset using the specified range.
+    [<CompiledName("DateTimeOffsetRanged")>]
+    let dateTimeOffsetRanged (range : Range<DateTimeOffset>) : Gen<System.DateTimeOffset> =
+        let minTicks =
+            System.DateTimeOffset.MinValue.Ticks
+        let maxTicks =
+            System.DateTimeOffset.MaxValue.Ticks
+        gen {
+            let! ticks = range |> Range.map (fun dt -> dt.Ticks) |> integral
+            // Ensure there is no overflow near the edges when adding the offset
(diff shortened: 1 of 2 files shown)


cmeeren:
We should probably add tests, at least for the `DateTimeOffset` generator(s). I experimented a bit with them now, and there may be a bug. I'll have a look at it tomorrow.

cmeeren:
The bug is now fixed. I have experimented with both `dateTimeOffset` and `dateTimeOffsetRanged` in FSI. They seem to produce sensible values according to the size (apart from the default range being `constant`, as previously mentioned), and never crash.

cmeeren:
To be clear, the only outstanding thing here is the compilation error I mentioned, and that can be done later (e.g. in another PR) or not at all (if you don't agree with the change). So feel free to review.

moodmosaic:
Interesting. Is that issue happening on master or just in v0.8.3?

cmeeren:
The compilation error? On master (and 0.8.4). I discovered it while working on this PR, so I don't see how 0.8.3 is relevant.

moodmosaic:
>There is one other change I'd like to make, but for some reason it won't compile: I'd like to change the default range of `dateTime` and `dateTimeOffset` from `constant` to either `linear` or `exponential`. I don't see a reason why it should be size-independent. But when I tried to change it, I get compile errors

Perhaps we're missing something in Numeric.fs(?) :thinking: Reminds me of https://github.com/hedgehogqa/fsharp-hedgehog/issues/156.

src/Hedgehog/Gen.fs
```diff
@@ -515,21 +515,43 @@ module Gen =
...
+                (14L * 60L)
+                ((ticks - DateTimeOffset.MinValue.Ticks) / TimeSpan.TicksPerMinute)
+            let! offsetMinutes = int (Range.linearFrom 0 (Operators.int minOffsetMinutes) (Operators.int maxOffsetMinutes))
+            return System.DateTimeOffset(ticks, TimeSpan.FromMinutes (Operators.float offsetMinutes))
+        }
+
+    /// Generates a random DateTime.
+    [<CompiledName("DateTime")>]
+    let dateTime : Gen<System.DateTime> =
+        dateTimeRanged (Range.constantFrom (DateTime (2000, 1, 1)) DateTime.MinValue DateTime.MaxValue)
```
moodmosaic, line 547: Perhaps we should just deprecate those in favor or the ranged ones with v0.9.0? (That is, we keep the same name `dateTime`, `dateTimeOffset` but we accept `Range` as parameter.)

After all, it's just a one liner to define the old one (assuming we name the ranged one as `dateTime`):

```f#
dateTime (Range.constantFrom (DateTime (2000, 1, 1)) DateTime.MinValue DateTime.MaxValue)
```

I understand that it's a breaking change, but we're v0.x so we should consider that trade-off if it helps keep the API clean.

moodmosaic (commented):
:+1: :100: I left a comment, but overall this looks great.

src/Hedgehog/Gen.fs
cmeeren: Sure. Do you want me to remove the non-ranged `datetimeoffset` then, since that's a new part of the API? No sense in adding something that's deprecated from day 1. :)

cmeeren: Or, when you say "deprecate", do you mean that we should just remove the non-ranged ones altogether?

moodmosaic: Yes, I think we should take them out. Are you OK with this @TysonMN?

TysonMN: If I correctly understand, then yes.  Here are my thoughts based on that linked code.  I think they are consistent with the conclusion being reached here.

There are two functions there: `dateTimeOffsetRanged` and `dateTimeOffset`.  Of course the preferred name is the shorter `dateTimeOffset`.  I think the function accepting `Range<DateTimeOffset>` should have the preferred name (i.e. `dateTimeOffset`).  Then we can either remove the other function or rename it to something like `dateTimeOffsetWithMaxRange`.

moodmosaic: Yep. I think that's one of the very few (if not the first) case in the last 10+ years where I think introducing a lil breaking change is actually fine.

moodmosaic (approved):
:+1:

cmeeren:
I have made the change now. Note:
* The test `dateTime shrinks to correct mid-value` still makes sense if one wants to test that it shrinks correctly based on the range. I left it in.
* When fixing the tests, I had to use `Range.constant`.  See OP for details. Also, it's not possible to use the `*Bounded` ranges, e.g. `Range.constantBounded` gives an error because `DateTime` does not support `Zero`.

moodmosaic:
>[..]I had to use `Range.constant` for the reason described in the OP. Furthermore, it's not possible to use the `*Bounded` ranges, e.g. `Range.constantBounded` gives an error because `DateTime` does not support `get_Zero`.

I'd like to look into this before merging, if that's OK.

cmeeren:
> I'd like to look into this before merging, if that's OK.

Absolutely.

cmeeren:
Any chance to have a look at this so it can be merged? :)

moodmosaic:
> > [..]I had to use `Range.constant` for the reason described in the OP. Furthermore, it's not possible to use the `*Bounded` ranges, e.g. `Range.constantBounded` gives an error because `DateTime` does not support `get_Zero`.
> 
> I'd like to look into this before merging, if that's OK.

That compiler error you were getting is because of the [`scaleLinear`](https://github.com/hedgehogqa/fsharp-hedgehog/blob/4e4f5c907cd67b796e66b279b89542b7502cd4df/src/Hedgehog/Range.fs#L180) function used internally by [`linearFrom`](https://github.com/hedgehogqa/fsharp-hedgehog/blob/4e4f5c907cd67b796e66b279b89542b7502cd4df/src/Hedgehog/Range.fs#L215). If you instead work on [`Ticks`](https://docs.microsoft.com/en-us/dotnet/api/system.datetime.ticks?view=net-5.0) and then map to DateTime, it should compile just fine:

```diff
@@ -53,11 +53,18 @@ let ``dateTime randomly generates value between max and min ticks`` () =
 [<Fact>]
 let ``dateTime shrinks to correct mid-value`` () =
     let result =
         property {
             let! actual = 
-              Range.constantFrom (System.DateTime (2000, 1, 1)) System.DateTime.MinValue System.DateTime.MaxValue
+              Range.linearFrom
+                (System.DateTime (2000, 1, 1))
+                     .Ticks
+                 System.DateTime.MinValue
+                     .Ticks
+                 System.DateTime.MaxValue
+                     .Ticks
+              |> Range.map System.DateTime
               |> Gen.dateTime
             System.DateTime.Now =! actual
         }
         |> Property.report
         |> Report.render
```

src/Hedgehog/Gen.fs
```diff
@@ -515,21 +515,32 @@ module Gen =
         return System.Guid bs
     }
 
-    /// Generates a random instant in time expressed as a date and time of day.
+    /// Generates a random DateTime using the specified range.
```
moodmosaic, line 518:
```suggestion
    /// Generates a random DateTime using the specified range.
    /// For example:
    ///   let range =
    ///      Range.constantFrom
    ///          (System.DateTime (2000, 1, 1)) System.DateTime.MinValue System.DateTime.MaxValue
    ///  Gen.dateTime range
```

moodmosaic (approved):
Thank you! Looking good :+1: I've suggested a small change in the docs. We can merge after that.

cmeeren:
> If you instead work on [`Ticks`](https://docs.microsoft.com/en-us/dotnet/api/system.datetime.ticks?view=net-5.0) and then map to DateTime, it should compile just fine:

Yes, but then you sidestep the whole generator and have effectively written your own generator. Furthermore, the same applies to `DateTimeOffset`, and that generator is significantly more complicated than just mapping from ticks to `DateTimeOffset`. Is there nothing that can be done to make these generators support non-constant ranges?

moodmosaic:
>Is there nothing that can be done to make these generators support non-constant ranges?

Not without tweaking [`scaleLinear`](https://github.com/hedgehogqa/fsharp-hedgehog/blob/4e4f5c907cd67b796e66b279b89542b7502cd4df/src/Hedgehog/Range.fs#L180) first.

>but then you sidestep the whole generator and have effectively written your own generator

I'm not sure I fully understand. I've added `.Ticks` to each one of the parameters here (and mapped to DateTime)

```
Range.constantFrom (System.DateTime (2000, 1, 1)) System.DateTime.MinValue System.DateTime.MaxValue
```

cmeeren:
> I'm not sure I fully understand. I've added `.Ticks` to each one of the parameters here (and mapped to DateTime)

My bad, I see you map the `Range` and not the `Gen` as I thought.

PR #253, opened 2020-12-12
Generate NuGet package during build

TysonMN:
How about generating a NuGet package as part of the build?

In general, it is useful to know if some change prevents the ability to create the NuGet package.  More specifically, this helps test unreleased code.  Of course it is possible to add a direct dependency on the project, but it is easier to obtain and share the NuGet package and depend on it after putting it in a local NuGet package source.  In particular, this change would make it easier to test PR #251.

### DIFF
--- a/src/Hedgehog/Hedgehog.fsproj
+++ b/src/Hedgehog/Hedgehog.fsproj
@@ -6,6 +6,7 @@
     <Description>Release with confidence.</Description>
     <Authors>Jacob Stanley;Nikos Baxevanis</Authors>
     <Copyright>Copyright © 2017</Copyright>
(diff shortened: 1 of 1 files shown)


cmeeren:
I always do it this way myself for my own packages. I have never found a downside. It remove the need for a separate `dotnet pack` step or similar.

moodmosaic:
Exactly, and we can take that section off from the README.md now, I suppose.

moodmosaic:
Not a huge deal, but I've run into this https://github.com/dotnet/sdk/issues/10335.

TysonMN:
Is the fix to remove the call to `dotnet pack` and just use the already built NuGet package?

moodmosaic:
Yep. I'll update the README.md 👀

Issue #154, opened 2018-01-24
Generate lazily evaluated sequences


cmeeren:
When writing tests for Hedgehog.Experimental I came across the need for just checking if a property holds at least some of the time. For example, when testing `withNull`, I create a property I run once that generates a list of, say, 1000 items and checks if it contains `null`.

When testing if a property holds some of the time, one might need to generate many items to avoid false negatives. But most of the time you might get a positive fairly quickly, say after 10 items and the rest of the 990 generated items are wasted. I thought about using `Gen.seq` instead of `Gen.list`, but it just generates a list and passes it to `Seq.ofList` and thus provides no performance benefits.

Is it possible to have `Gen.seq` actually generate a lazily evaluated sequence?

moodmosaic:
Perhaps with a custom generator `seq 'a` that uses [sequence expressions](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/sequences#sequence-expressions) internally?

cmeeren:
I hope that's possible, but I don't know how. AFAIK you can't easily mix computation expressions.

Porges:
Maybe we could invent some kind of ... computation expression transformer 😉

moodmosaic:
Sure! A computation expression (CE) is really a class (a type that represents objects that can have properties, methods, and events).

So it should be possible to define a CE with a non-parameterless constructor taking another CE as a parameter, similar to [this](https://stackoverflow.com/a/12340269/467754) one.

TysonMN:
It is easy to lazily obtain a given number of values from some generator.  However, I don't know how to create `Gen<seq<'a>>` in which the generated sequences are lazy.  The difference between these two statements is (integrated) shrinking. 
 The current implementation to create `Gen<List<'a>>` converts `List<Tree<'a>>` to `Tree<List<'a>>` (which goes by the name `sequence` in functional programming).  Maybe a compromise is that `Gen<seq<'a>>` would return lazy values before shrinking and `list`s cast as `seq`s after shrinking has started.  Not sure though.  This path is currently unclear to me.

[...]

Issue #192, opened 2020-01-10
Incorrect shrinking (mutable objects in generator)


cmeeren:
In Hedgehog.Experimental I have this:

```f#
/// Shuffles the case of the given string.
let shuffleCase (s: string) =
  gen {
    let sb = Text.StringBuilder()
    for i = 0 to s.Length - 1 do
      let! b = Gen.bool
      let f = if b then Char.ToUpperInvariant else Char.ToLowerInvariant
      sb.Append (f s.[i]) |> ignore
    return sb.ToString()
  }
```

Let's try to sample this:

```f#
shuffleCase "abc" |> Gen.printSample
```

Yields:

```
=== Outcome ===
"abc"
=== Shrinks ===
.
=== Outcome ===
"aBC"
=== Shrinks ===
"aBCb"
"aBCbc"
.
=== Outcome ===
"ABc"
=== Shrinks ===
"ABca"
"ABcab"
.
=== Outcome ===
"abC"
=== Shrinks ===
"abCc"
.
=== Outcome ===
"abc"
=== Shrinks ===
.
```

As you can see, there's something weird going on with shrinking. Additional characters are added.

If I instead implement it in an immutable way, it works. For example:

```f#
let shuffleCase (s: string) =
  let rec inner (i: int) (soFar: string) =
    gen {
      if i = s.Length then return soFar
      else
        let c = s.[i]
        let! b = Gen.bool
        let f = if b then Char.ToUpperInvariant else Char.ToLowerInvariant
        return! inner (i + 1) (soFar + string (f c))
    }
  inner 0 ""
```

Why does the first one fail? Is it a general rule that using (locally defined) mutable objects in a generator will cause issues for shrinking?

cmeeren:
Might be related to (and possible duplicate of) #157.

moodmosaic:
Yep, seems like a dupe of #157 which we've never get into it. I think what's happening is that everything above `let!` will execute once, and then the shrinking occurs below.

I'd be tempted to compare these results with the haskell version, because we had a similar issue reported there in https://github.com/hedgehogqa/haskell-hedgehog/issues/122.

cmeeren:
I see. The Hedgehog.Experimental library contains more generators that use mutable state. Do you think you'd be able to have a look at this in the foreseeable future, or should I prioritize implementing them without immutability? (That might impact performance, so I'd rather not.)

[...]

PR #260, opened 2021-01-02
add ListGen.traverse

TysonMN:
This PR adds `ListGen.traverse` and its specialization `sequence`.

### DIFF
--- a/src/Hedgehog/ListGen.fs
+++ b/src/Hedgehog/ListGen.fs
@@ -1,0 +1,17 @@
+﻿namespace Hedgehog
+
+module ListGen =
+
+    let traverse (f: 'a -> Gen<'b>) (ma: list<'a>) : Gen<list<'b>> =
+        let rec loop input output =
+            match input with
+            | [] -> output |> List.rev |> Gen.constant
+            | a :: input ->
+                gen {
+                    let! b = f a
+                    return! loop input (b :: output)
+                }
+        loop ma []
+
+    let sequence (gens : List<Gen<'a>>) : Gen<List<'a>> =
+        gens |> traverse id
--- a/src/Hedgehog/Hedgehog.fsproj
+++ b/src/Hedgehog/Hedgehog.fsproj
@@ -32,6 +32,7 @@ https://github.com/hedgehogqa/fsharp-hedgehog/blob/master/doc/tutorial.md
     <Compile Include="Random.fs" />
     <Compile Include="Shrink.fs" />
     <Compile Include="Gen.fs" />
+    <Compile Include="ListGen.fs" />
     <Compile Include="Journal.fs" />
     <Compile Include="Tuple.fs" />
     <Compile Include="Outcome.fs" />
--- a/tests/Hedgehog.Tests/GenTests.fs
+++ b/tests/Hedgehog.Tests/GenTests.fs
@@ -4,6 +4,16 @@ open Hedgehog
 open Swensen.Unquote
 open Xunit
 
+
+[<Fact>]
+let ``Large input to ListGen sequence does not cause stack overflow`` () =
+  let n = 10000
+  Gen.constant 1
+  |> List.replicate n
+  |> ListGen.sequence
+  |> Gen.sample 0 1
+  |> ignore
+
 [<Theory>]
 [<InlineData(  8)>]
 [<InlineData( 16)>]
--- a/tests/Hedgehog.Tests/GenTests.fs
+++ b/tests/Hedgehog.Tests/GenTests.fs
@@ -4,16 +4,6 @@ open Hedgehog
 open Swensen.Unquote
 open Xunit
 
-
-[<Fact>]
-let ``Large input to ListGen sequence does not cause stack overflow`` () =
-  let n = 10000
-  Gen.constant 1
-  |> List.replicate n
-  |> ListGen.sequence
-  |> Gen.sample 0 1
-  |> ignore
-
 [<Theory>]
 [<InlineData(  8)>]
 [<InlineData( 16)>]


TysonMN:
I think `traverse` can be used to simplify the immutable workaround @cmeeren gave at the end of the OP in #192.

TysonMN:
I think this code can be generalized to accept (a lazy) `'a seq` and return `Gen<'a seq>`, where the returned sequence is actually lazy.  This would resolve #154.

dharmaturtle:
[This SO answer](https://stackoverflow.com/a/1306267) doesn't advise using `seq` for recursive algorithms.

TysonMN:
Quoting that answer.
> Any time you break apart a seq using `Seq.hd` and `Seq.skip 1` you are almost surely falling into the trap of going `O(N^2)`.

Exactly.  That is what I told you I was concerned about last night.

Quoting that answer again.
> `IEnumerator<T>` is workable, but its API programming model is not so fun/easy to work with.

Indeed, I was thinking of using `IEnumerator<T>` instead.  I presented this first using `list`s since its API is fun and easy to work with.

src/Hedgehog/ListGen.fs
```diff
@@ -0,0 +1,16 @@
...
+            match input with
+            | [] -> output |> List.rev |> Gen.constant
+            | a :: input ->
+                gen {
+                    let! b = f a
+                    return! loop input (b :: output)
+                }
+        loop ma []
+
+    let sequence ma = traverse id ma
```
dharmaturtle, line 16: `mma` ;)

TysonMN: :D

src/Hedgehog/ListGen.fs
```diff
@@ -0,0 +1,16 @@
+﻿namespace Hedgehog
+
+module ListGen =
+
+    let traverse (f: 'a -> Gen<'b>) (ma: list<'a>) : Gen<list<'b>> =
```
adam-becker, line 5: Out of curiosity, since this returns a `Gen<list<'b>>`, shouldn't the module be named `GenList`?

adam-becker: This would be more consistent with `src/Hedgehog/Tuple.fs` in #247

TysonMN: > Out of curiosity, since this returns a `Gen<list<'b>>`, shouldn't the module be named `GenList`?

Great question.  I always have to think carefully about which module name I think is correct.

Functions are typically contained in modules named after their input type.  For example, the [function `isSome`](https://fsharp.github.io/fsharp-core-docs/reference/fsharp-core-optionmodule.html#isSome) with type `'a option -> bool` is in a module named `Option` instead of one named `Bool`.  In addition to being typical, I think this is also the correct naming convention.  Given an instance of type `Foo`, a programmer knows they can type `Foo.` in their IDE and get suggestions about what they could do with their instance.

Of course `traverse` doesn't have `Gen<'a> list` (or `'a Gen list` when consistently using the postfix notation for generic types) as input.  Nor does it have `Gen<'a list>` (or `'a list Gen`) as input.  However, `traverse` should be in the same module as `sequence`, which can be implemented as `traverse id`.  The lone input to `sequence` has type `Gen<'a> list` (or `'a Gen list`).  Hence the module name `ListGen`.

_\*reflects on what was just written\*_

I just realized that this naming convention also depends on a preference for the generic type notation.  If the prefix notation is preferred, then my argument above says the module should be named `ListGen`, but if the postfix notation is preferred, then my argument above says the the module name should be `GenList`.

I prefer the prefix notation.  I think this is because I primarily used Java in school for ten years and then C# professionally for four years before picking up F#.

_\*more thinking\*_

The postfix generic type syntax is only available for generic types with a single type argument, right?  If so, then I think the module naming convention should be based on the prefix syntax.  Consider the nested type `Either<Map<'k, 'v, 'e>`.  It is a functor in `'v` (among other things).  In what module should functions for it like `map` exist?  Since the only way to express this type is with the prefix notation, I think the functions should be in a module named `EitherMap`.  Then for consistency, I think all nested types should have their functions contained a module matching the name of their type when expressed using the prefix notation.

> This would be more consistent with `src/Hedgehog/Tuple.fs` in #247

For reference, that module and its functions are [here](https://github.com/hedgehogqa/fsharp-hedgehog/pull/247/files#diff-0498fde986354191be905aa346ef56e4f314b6b38e18de8287148496ed061d31R12-R19).

I think `ListGen` and `GenTuple` are consistently named.  In general, I think you correctly named that module `GenTuple` because the outer type is `Gen<>` and the inner type is a tuple.

However, it is not just any tuple.  It is specifically a pair.  I just that you change the name of that module to `GenPair`.  I will make the same comment there for completeness.

TysonMN:
> > `IEnumerator<T>` is workable, but its API programming model is not so fun/easy to work with.
> 
> Indeed, I was thinking of using `IEnumerator<T>` instead. I presented this first using `list`s since its API is fun and easy to work with.

I have thought more about this.  In particular, I have looked at the method [`Gen.Builder.For`](https://github.com/hedgehogqa/fsharp-hedgehog/blob/master/src/Hedgehog/Gen.fs#L140-L145), which uses `seq<>`/`IEnumerable<>`.  Both [`MoveNext`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.ienumerator.movenext#System_Collections_IEnumerator_MoveNext) and [`Current`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.ienumerator-1.current#System_Collections_Generic_IEnumerator_1_Current) are clunky to use but pose no technical difficulty.  The part that is easy to get wrong is that the instance of [`IEnumerator<>`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.ienumerator-1) should be disposed, and the most difficult case of that is when an exception is thrown.  The correct implementation might be just as simple as the implementation of `Gen.Builder.For`, but it still takes a lot of thought for me to convince myself that this implementation is correct.

dharmaturtle:
>The part that is easy to get wrong is that the instance of IEnumerator<> should be disposed, and the most difficult case of that is when an exception is thrown.

Both `use` and `using` dispose when an exception is thrown.

image

image

Code so you can test it yourself:

```
[<Fact>]
let ``use disposes`` () =
    let imDisposable =
        { new System.IDisposable with
            member _.Dispose() =
                printfn "Use disposed!"
        }
    use __ = imDisposable
    failwith "oh no!"
    ()

[<Fact>]
let ``using disposes`` () =
    let imDisposable =
        { new System.IDisposable with
            member _.Dispose() =
                printfn "Using Disposed!"
        }
    using imDisposable (fun _ ->
        failwith "oh no!"
    )
    ()
```

The Gen builder [currently uses `using`](https://github.com/hedgehogqa/fsharp-hedgehog/blob/master/src/Hedgehog/Gen.fs#L142).

TysonMN:
Yes.  The other concern is that you are on a knife's edge.  You have to both ensure that you dispose eventually (and ideally ASAP) as well as not too soon.  Your code shows how to achieve the first requirement.  Now you need to make sure the disposable doesn't happen too early (because of the lazy behavior).

In C# (where I am perfectly confident in my understanding of the problem),
```csharp
public Task<int> Method(Func<CancellationToken, Task<int>> f) {
  using (var cts = new CancellationTokenSource()) {
    return f(cts.Token);
  }
}
```

disposes too early.  Instead, the correct code is
```csharp
public async Task<int> Method(Func<CancellationToken, Task<int>> f) {
  using (var cts = new CancellationTokenSource()) {
    return await f(cts.Token);
  }
}
```

The nice thing about disposing too early is that it is easy to check if it is happening.  If essentially any nontrivial test of the function passes, then disposing never happens too early.  It is harder to test if disposing is happening too late, but using the same `using xse (fun xse -> ...)` syntax as `Gen.Builder.For` probably ensures that it is eventually happening (and right on time).

src/Hedgehog/ListGen.fs
adam-becker: > ... because the outer type is Gen<> and the inner type is a tuple.

That's what I was getting at. The same is true here, where `Gen<>` is the outer type, and `List<>` is an inner type. If one prefers lighter weight generic syntax, it could be written as `Gen (List a)`. I would argue for naming based on an "outside-in" approach, where this PR uses "inside-out". I'm not sure if there's an established naming convention for this. I'm sure Haskell-ers have something, since they must deal with this more.

Module functions can also be grouped on the return value, lest we forget about `val constant : 'a -> Gen<'a>`. No one would argue that isn't a `Gen` function. Since `traverse` and `sequence` do the functor-flip, but both result in a `Gen<List<...>>` I would argue that the return type is more significant to the module they're contained in.

I'm genuinely curious if an established naming convention exists though, as I'd like to follow the principal of least surprise here.

adam-becker: I'd also be totally fine with putting a `List` module under `Gen`, so accessing these functions would be `Gen.List.traverse`

adam-becker, line 5:
```suggestion
    let traverse (f : 'a -> Gen<'b>) (ma : list<'a>) : Gen<list<'b>> =
```

src/Hedgehog/ListGen.fs
```diff
@@ -0,0 +1,16 @@
...
+            match input with
+            | [] -> output |> List.rev |> Gen.constant
+            | a :: input ->
+                gen {
+                    let! b = f a
+                    return! loop input (b :: output)
+                }
+        loop ma []
+
+    let sequence mma = mma |> traverse id
```
adam-becker, line 16:
```suggestion
    let sequence (gens : List<Gen<'a>>) : Gen<List<'a>> =
        gens |> traverse id
```

adam-becker (changes requested):
Outside of the ongoing discussion about the module name, these were the only nitpicks I have.

src/Hedgehog/ListGen.fs
TysonMN: To summarize, there are three things to decide.
1. Outside-in approach vs inside-out approach
2. Prefix or postfix notation for generic types
3. Based on the type of an input or the output

> I would argue for naming based on an "outside-in" approach, where this PR uses "inside-out".

Me too.  I also want to name the modules based on the outside-in approach.

> > ... because the outer type is Gen<> and the inner type is a tuple.
> 
> `Gen<>` is the outer type, and `List<>` is an inner type [of the return type of `traverse`].

I want to name the modules based on the prefix notation of the generic type.  That quote makes me think that you do as well.

> Module functions can also be grouped on the return value, lest we forget about `val constant : 'a -> Gen<'a>`. No one would argue that isn't a `Gen` function. Since `traverse` and `sequence` do the functor-flip, but both result in a `Gen<List<...>>` I would argue that the return type is more significant to the module they're contained in.
> 
> I'm genuinely curious if an established naming convention exists though, as I'd like to follow the principle of least surprise here.

I think the only thing we disagree on there is whether the name of a module should match the type of an input or the output of the functions it contains.

I certainly admit that I want to put `traverse` in a module named `ListGen` even though `'a Gen list` isn't actually a type of one of its inputs.

I think your example of `constant` is misleading.  There are many functions merely meant to convert between types in a sense that is primarily syntactic.  A great example of this is `Seq.toList`.  Sometimes functions exist in both directions, such as `List.ofSeq`.  Sometimes only one of these to functions exists.  For your example, there is no containing type, so it is not even clear what name would be for the module containing the other function.  Sometimes one of these two functions is missing because F# makes it difficult to create cyclic dependencies.

In summary, there are many functions of the form `A.toB` and `B.ofA` both of type `'a -> 'b`.  My impression is that there are more such functions of the form `A.toB`.

> I'm not sure if there's an established naming convention for this.

Because there exists functions of both forms `A.toB` and `B.ofA`, any convention is not strictly followed.

When you questioned my naming, I was surprised to realize that there is more sublity than I had previously noticed and that I couldn't find any examples to justify my naming convention.

Here is one though, which is [the post by Scott Wlaschin about `traverse` and `sequence`](https://fsharpforfunandprofit.com/posts/elevated-world-4).  Specifically consider [this code](https://fsharpforfunandprofit.com/posts/elevated-world-4/#applicative-vs-monadic-versions-of-traverse:~:text=module%20List%20%3D,let%20rec%20traverseResultA%20f%20list%20%3D).
```fsharp
module List =

    /// Map a Result producing function over a list to get a new Result 
    /// using applicative style
    /// ('a -> Result<'b>) -> 'a list -> Result<'b list>
    let rec traverseResultA f list = [omitting implementation]
```

Scott's naming convention is to put `traverse` in a module named `List` to primarily match the last input type.  Of course it needs further distinguishing, so he appends `Result` to the function name.  (The `A` is orthogonal to our discussion.)  So Scott is primarily emphasizing `List` and secondarily emphasizing `Result`.  I agree with that order of emphasis.  I just think it is better to move `Result` up to a module and simplify the function name to just `traverse`.

> I'd also be totally fine with putting a `List` module under `Gen`, so accessing these functions would be `Gen.List.traverse`

This is a good idea, but of course I think it should be `List.Gen.traverse`.  This would also be more consistent with the fact that the last input of traverse is `'a list` (with `'a = _ Gen`).

> I'm sure Haskell-ers have something, since they must deal with this more.

This is not an issue for them.  Higher-kinded types allow them to simply write `map` instead of `Gen.map`, `List.map`, etc.

TysonMN: The F# style guide says to only put white space on the right side of those colons.
https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting#right-pad-function-argument-type-annotations

TysonMN:
As I said in https://github.com/hedgehogqa/fsharp-hedgehog/pull/238#discussion_r558298308, the code in this branch lacks tail calls.

TysonMN:
I pushed a failing test.  It is good to see that [an aborted test run](https://github.com/hedgehogqa/fsharp-hedgehog/actions/runs/488417802) (which is what happens when the stack overflows during a test run) is correctly recorded as failing the build step.

TysonMN:
@adam-becker, I think I figured out how to write this in a way that won't stack overflow.  This should help with `replicate` in PR #238.  I will give another update when I have more time.

src/Hedgehog/ListGen.fs
adam-becker: Be that as it may, the rest of the code in this code base doesn't do that. I value internal consistency over edicts :smile:

adam-becker: We could probably investigate #118 in a follow up PR to address style concerns like this, it would definitely be beneficial.

TysonMN: I don't mind the formatting change.  I just wanted to provide the link for reference.

TysonMN: I have my style preferences, but I most concerned with consistency.

TysonMN:
I pushed a commit with a different implementation (one that works with `seq` instead of `list` @dharmaturtle), but it there is still a stack overflow.  I think there are at least two reasons for this.

First, it just dawned on me that `Gen.appy` is implemented by calling `Gen.bind`.  This is certainly possible (because every monad is an applicative functor), but it is the best way to do it.  I will make an issue for this.

Second, even in my code at work where I use this implementation of `traverse` with my own implementation of the validation monad (and I "correctly" implemented `apply`), there is still a stack overflow.  I don't know why though (and don't have time to investigate that now).

TysonMN:
> Second, even in my code at work where I use this implementation of `traverse` with my own implementation of the validation monad (and I "correctly" implemented `apply`), there is still a stack overflow. I don't know why though (and don't have time to investigate that now).

I partially figured this out.  There was a problem with the functions used to (lazily) create the sequence.  There was no stack overflow after I switched everything from `seq` to `list`.  For now, I am satisfied with using `list` instead of `seq`.

Therefore, I switched this new implementation in this branch to use `list`.

The test in this branch still fails.  I think that is because `apply` is implemented via `bind`, which is not the most efficient implementation (c.f. #272).  I could be wrong about that though.

Maybe it is the laziness of `Gen` is "causing" the stack to overflow (just like the laziness of `seq` was "causing" the stack to overflow in my code at work).  There must be a way to do both laziness and not overflow the stack.  It might involve continuations.  Not sure.

TysonMN:
> > Second, even in my code at work where I use this implementation of `traverse` with my own implementation of the validation monad (and I "correctly" implemented `apply`), there is still a stack overflow. I don't know why though (and don't have time to investigate that now).
> 
> I partially figured this out. There was a problem with the functions used to (lazily) create the sequence. There was no stack overflow (_in my code at work_) after I switched everything from `seq` to `list`. For now, I am satisfied with using `list` instead of `seq`.

I am confusing myself.

I just checked again.  Using `seq` in my code at works passes the test with `n = 10000000`.  It fails for the next order of magnitude because of an `OutOfMemoryException`.

So I now feel confident switching this implementation back to being lazy (by using `seq`).  However, I will wait until the test passes using `List`, which I think is blocked by #272.

TysonMN:
Similar to https://github.com/hedgehogqa/fsharp-hedgehog/pull/238#discussion_r563175602, I don't think we should hold up this PR because the code an overflow the stack.  As such, I have removed the test that cases a stack overflow.

src/Hedgehog/ListGen.fs
adam-becker: > I don't mind the formatting change. I just wanted to provide the link for reference.

Understood.

> I have my style preferences, but I most concerned with consistency.

Same. I wouldn't write code the way it's written in this code base, but I seldom break with tradition on matters like this.

adam-becker:
> Similar to [#238 (comment)](https://github.com/hedgehogqa/fsharp-hedgehog/pull/238#discussion_r563175602), I don't think we should hold up this PR because the code can overflow the stack. As such, I have removed the test that cases a stack overflow.

I think we should still try to investigate removing stack overflows if possible, I can make an issue for it.

@moodmosaic are you comfortable merging these changes as long as we keep track of the fact that these functions can overflow? I can make an issue for these functions overflowing once they're in `master`.

src/Hedgehog/ListGen.fs
moodmosaic: Most of the style is back from 2016-2017 by the way. Is there some official style/formatting guide by Microsoft or by the F# community?

TysonMN: Yes.  [Microsoft has an F# style guide](https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/).

adam-becker: I vote we apply the style guide as a separate PR so we can consider it on its own merits without having to also consider logic changes. I'm in favor of using the style guide, but I don't want to block features.

moodmosaic:
@adam-becker, happy to merge if you think it's good to go.

I apologize for not being able to get through all of the open pull requests. They all look promising. It's just that I'm travelling<sup>*</sup> at the moment, so the time I have is limited.

---

<sup>*</sup>At the same time, I don't want to be the bottleneck for merging pull requests made by @TysonMN and you. If you guys review each others work and approve, we could merge.

Though, at some point, we need to release v0.10.0, and another, final, release with breaking changes before v1.0.0 where breaking changes are no longer desirable, unless there's a major version bump.

src/Hedgehog/ListGen.fs
moodmosaic: (FWIW, item no. 3 from that style guide originates from https://twitter.com/dsymetweets/status/843849405212098560.)

PR #261, opened 2021-01-05
automated publishing of NuGet package

### LINKED ISSUE #215: Automate release process

TysonMN:
Resolves #215

This PR adds the GitHub Action called [Publish NuGet](https://github.com/marketplace/actions/publish-nuget).  When executed, it publishes a NuGet package if the version (as specified in the `Version` element in the project file) has changed.

I found it because it is being used in `fsharp-hedgehog-experimental` [here](https://github.com/hedgehogqa/fsharp-hedgehog-experimental/blob/master/.github/workflows/dotnet-core.yml#L24-L29) (and with help from @dharmaturtle).

I didn't copy things exactly as they are in `fsharp-hedgehog-experimental`.  Instead, I made it very clear when the publish NuGet package code will execute: only when `master` changes.  If the files were left together, I was concerned that anyone could create pull request that changes the version and cause a NuGet package release even without the pull request being completed.  Maybe that wouldn't be possible, but I don't think that is obvious.  By separating the files, it is obvious that this bad behavior cannot occur.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,8 +1,6 @@
-name: .NET Core
+name: Pull Request
 
 on:
-  push:
-    branches: [ master ]
   pull_request:
     types:
       - opened
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -1,0 +1,33 @@
+name: master
+
+on:
+  push:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: windows-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Ensure no Windows line endings
+      run: ./validate.sh
+    - name: Setup .NET Core
+      uses: actions/setup-dotnet@v1
+      with:
+        dotnet-version: 3.1.403
+    - name: Install dependencies
+      run: dotnet restore
+    - name: Build
+      run: dotnet build --no-restore --configuration Release
+    - name: Test (Hedgehog.Tests)
+      run: dotnet test --no-build --configuration Release --verbosity normal tests/Hedgehog.Tests/Hedgehog.Tests.fsproj
+    - name: Test (Hedgehog.CSharp.Tests)
+      run: dotnet test --no-build --configuration Release --verbosity normal tests/Hedgehog.CSharp.Tests/Hedgehog.CSharp.Tests.csproj
+    - name: Publish NuGet
+      uses: brandedoutcast/publish-nuget@v2.5.5
+      with:
+        PROJECT_FILE_PATH: "src/Hedgehog/Hedgehog.fsproj"
+        PACKAGE_NAME: "Hedgehog"
+        NUGET_KEY: ${{secrets.NUGET_API_KEY}}
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-fsharp-hedgehog [![NuGet][nuget-shield]][nuget] ![](https://github.com/hedgehogqa/fsharp-hedgehog/workflows/.NET%20Core/badge.svg)
+fsharp-hedgehog [![NuGet][nuget-shield]][nuget] ![](https://github.com/hedgehogqa/fsharp-hedgehog/workflows/master/badge.svg)
 ========
 
 > Hedgehog will eat all your bugs.


moodmosaic (commented):
Looks great 👍 Can't wait to see this in action. I must set the NuGet key in the settings and then we can merge. 🎸

dharmaturtle:
It's pretty simple. Generate the api key here: https://www.nuget.org/account/apikeys

image

And stick it in here:

image

.github/workflows/publish_nuget.yml
```diff
@@ -0,0 +1,33 @@
+name: Publish Nuget
+
+on:
+  push:
+    branches: [ master ]
```
adam-becker, line 5: Does Github expose an `on:push:tags` for actions?

dharmaturtle: Yes: https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#create

TysonMN:
> I must set the NuGet key in the settings and then we can merge.

Oh, yep.  I forgot to mention that.  Good catch @moodmosaic.

And thanks for the tutorial-esque screenshots @dharmaturtle! :D

.github/workflows/publish_nuget.yml
adam-becker: Should we use that to create a release instead? Then a tag for `1.2.0` would kick off the release process for that version. I understand wanting more of the "nightly" kind of builds too so pushing to master makes sense, but I think doing the actual release with tags makes more sense.

dharmaturtle: Depends on how the maintainers want to manage this. For me personally, I prefer managing the version number from the `*.fsproj`. Making a tag in git to trigger a package release doesn't suit my style. I'm also unaware of any github action that uploads to Nuget upon tag creation, so we might have to make one for ourselves.

Side note, [the github action being used](https://github.com/marketplace/actions/publish-nuget) automatically creates a tag with the appropriate version number (if the push to Nuget is successful.)

.github/workflows/publish_nuget.yml
```diff
@@ -0,0 +1,33 @@
...
+    - name: Test (Hedgehog.Tests)
+      run: dotnet test --no-build --configuration Release --verbosity normal tests/Hedgehog.Tests/Hedgehog.Tests.fsproj
+    - name: Test (Hedgehog.CSharp.Tests)
+      run: dotnet test --no-build --configuration Release --verbosity normal tests/Hedgehog.CSharp.Tests/Hedgehog.CSharp.Tests.csproj
+    - name: Publish NuGet
+      uses: brandedoutcast/publish-nuget@v2.5.5
+      with:
+        PROJECT_FILE_PATH: "src/Hedgehog/Hedgehog.fsproj"
+        PACKAGE_NAME: "Hedgehog"
+        NUGET_KEY: ${{secrets.NUGET_API_KEY}}
```
moodmosaic, line 33: Done. Added `NUGET_API_KEY` as per @dharmaturtle's screenshot. :surfer:

.github/workflows/pull_request.yml
```diff
@@ -1,8 +1,6 @@
-name: .NET Core
+name: Pull Request
```
moodmosaic, line 1: Good catch. That's a better name.

.github/workflows/publish_nuget.yml
```diff
@@ -0,0 +1,33 @@
...
+    - name: Install dependencies
+      run: dotnet restore
+    - name: Build
+      run: dotnet build --no-restore --configuration Release
+    - name: Test (Hedgehog.Tests)
+      run: dotnet test --no-build --configuration Release --verbosity normal tests/Hedgehog.Tests/Hedgehog.Tests.fsproj
+    - name: Test (Hedgehog.CSharp.Tests)
+      run: dotnet test --no-build --configuration Release --verbosity normal tests/Hedgehog.CSharp.Tests/Hedgehog.CSharp.Tests.csproj
+    - name: Publish NuGet
+      uses: brandedoutcast/publish-nuget@v2.5.5
```
moodmosaic, line 29: Looks like this is exactly what we need :+1:

moodmosaic (approved):
:+1: :100:

.github/workflows/publish_nuget.yml
moodmosaic: We can try it and see how it goes. Nothing stops us from using that other action that gets triggered upon tag creation.

moodmosaic:
Ready to merge if we all agree :eyes:

dharmaturtle:
>I made it very clear when the publish NuGet package code will execute: only when master changes. If the files were left together, I was concerned that anyone could create pull request that changes the version and cause a NuGet package release even without the pull request being completed. Maybe that wouldn't be possible, but I don't think that is obvious. By separating the files, it is obvious that this bad behavior cannot occur.

I tested this [here](https://github.com/dharmaturtle/fsharp-hedgehog-xunit/pull/1) and [here](https://github.com/dharmaturtle/fsharp-hedgehog-xunit/pull/2). The resulting runs [here](https://github.com/dharmaturtle/fsharp-hedgehog-xunit/actions/runs/464964507) and [here](https://github.com/dharmaturtle/fsharp-hedgehog-xunit/actions/runs/464973958) complain `NUGET_KEY not given`. So the github action is run against the PR's author's repo's credentials.

TysonMN:
> Ready to merge if we all agree 👀

I agree

> I tested this [here](https://github.com/dharmaturtle/fsharp-hedgehog-xunit/pull/1) and [here](https://github.com/dharmaturtle/fsharp-hedgehog-xunit/pull/2). The resulting runs [here](https://github.com/dharmaturtle/fsharp-hedgehog-xunit/actions/runs/464964507) and [here](https://github.com/dharmaturtle/fsharp-hedgehog-xunit/actions/runs/464973958) complain `NUGET_KEY not given`. So the github action is run against the PR's author's repo's credentials.

What if the PR is from an owner of the repo using a branch in the repo?

moodmosaic:
>So the github action is run against the PR's author's repo's credentials.

That's OK as those PRs (that bump the version) are only being made by maintainers.

TysonMN:
That is certainly the intention, but it is more a security concern that maybe anyone _could_ create a PR that causes a release (by increased the version).  However, @dharmaturtle's tests show that this is not possible.

dharmaturtle:
What if the PR is from an owner of the repo using a branch in the repo?

image

It's possible that an _open_ PR will bump the version number.

I like this PR as is.

dharmaturtle:
Aaaaaaactually we have a badge on the Readme that reflects the build status:

image

If we change the name of the workflows, we need to update that too.

Do we want to display `Publish Nuget | Passing` or `Pull Request | Passing` on the Readme?

moodmosaic:
@dharmaturtle

> Do we want to display `Publish Nuget | Passing` or `Pull Request | Passing` on the Readme?

Perhap's `master | Passing`? (From the branch name.)

moodmosaic:
@TysonMN 

>but it is more a security concern that maybe anyone *could* create a PR that causes a release

From what I can tell, he or she must be a member of the repository otherwise the GitHub Action won't be able to push the package.

Still, if we're not 100% sure, we can merge this pull request *just* before doing the next release.

dharmaturtle:
Ah. `master | passing` will mean renaming the workflow since workflow badges don't support customizing the name.

Some docs: https://docs.github.com/en/free-pro-team@latest/actions/managing-workflow-runs/adding-a-workflow-status-badge

There are actions out there to build a custom badge, but that might be more effort than it's worth.

moodmosaic:
Then it's up to @TysonMN or you to decide. Either of those look fine. (`Publish Nuget | Passing` or `Pull Request | Passing`)

TysonMN:
I force pushed a change to make name of the workflow `master` and changed the badge on the README to match.  Keeping the workflows separate makes sense because of that badge.



> From what I can tell, he or she must be a member of the repository otherwise the GitHub Action won't be able to push the package.
> 
> Still, if we're not 100% sure, we can merge this pull request _just_ before doing the next release.

@dharmaturtle tested.  I am convinced that there is no issue.

I think this PR is ready to merge.

dharmaturtle (approved):

moodmosaic:
🧨 💯

PR #264, opened 2021-01-11
Add section to readme, rename namespace.

### LINKED ISSUE #262: Add C# section on the README
Just so that C# users are aware of the Hedgehog.CSharp namespace and its capabilities.

### LINKED ISSUE #263: Rename Hedgehog.CSharp to Hedgehog.Linq
That's a friendlier namespace for VB.NET users, in a similar fashion to System.Linq.

adam-becker:
Renamed the `Hedgehog.CSharp` namespace to `Hedgehog.Linq`. Also added some documentation to `README.md`.

Fixes #262
Fixes #263


adam-becker:
@moodmosaic @TysonMN

moodmosaic:
Looks good! Build failed, it looks like we'd have to update some other parts as well (e.g. test code).

adam-becker:
Woops, got lazy and it bit me.

adam-becker:
Ready for review again.

On a side note, few things are more amusing than a commit that says "Fix build" with a big red "X" next to it.

README.md
```diff
@@ -39,13 +39,40 @@ You can then load the module in F# Interactive, and run it:
...
+```csharp
+using Hegehog.Linq;
+
+class Generators
+{
+    static void Example()
+    {
+        // This creates a generator for 10 character strings, with only alphabetical characters.
+        var stringLength = 10;
+        var stringGen = Gen.String(Gen.Alpha, Range.FromValue(stringLength));
```
adam-becker, line 63: Now seeing this used, I don't think I like `Range.FromValue`, maybe `Range.Constant` would be better after all.

moodmosaic: I think `FromValue` is a friendlier name for developers that are new to functional programming. I've seen developers confusing singleton with the *singleton* lifetime when doing dependency injection, for example.

OTOH, if we don't push towards the correct terminology, the world will stay forever in this weird state of trying to map OO to FP in a funky way. (Perhaps that's one of the reasons Mark Seemann uses the terms as-is, e.g. *Either* instead of *ROP*.)

moodmosaic: We can tackle this separately, of course.

PR #266, opened 2021-01-12
Rearrange parameters for better chaining.

### LINKED ISSUE #265: Hedgehog bind functions are backwards
Axiomatically, the `bind` function in F# has a different parameter order than the `Bind` method in a computation expression. This is well established, and Hedgehog does it wrong.

```fsharp
module Gen =
    val bind : Gen<'a> -> ('a -> Gen<'b>) -> Gen<'b>

module Property =
    val bind : Property<'a> -> ('a -> Property<'b>) -> Property<'b>

module Random =
    val bind : Random<'a> -> ('a -> Random<'b>) -> Random<'b>

module Tree =
    val bind : Tree<'a> -> ('a -> Tree<'b>) -> Tree<'b>
```

Other modules in F# are consistent on this.

```fsharp
module Option =
    val bind : ('a -> Option<'b>) -> Option<'a> -> Option<'b>

module Result =
    val bind : ('a -> Result<'b, 'error>) -> Result<'a, 'error> -> Result<'b, 'error>

module Seq =
    val collect : ('a -> seq<'b>) -> seq<'a> -> seq<'b>

module Array =
    val collect : ('a -> Array<'b>) -> Array<'a> -> Array<'b>

module List =
    val collect : ('a -> List<'b>) -> List<'a> -> List<'b>
```

It's done this way to allow simple chaining, which is how I noticed this.

adam-becker:
Fixes #265.

The goal is to allow better data flow through the combinators, like:

```fsharp
Gen.alpha
|> Gen.tryWith (fun exn -> ...)
|> Gen.bind (fun ch -> ...)
|> Gen.map (fun ch -> ...)
```

### DIFF
--- a/src/Hedgehog/Gen.fs
+++ b/src/Hedgehog/Gen.fs
@@ -512,12 +512,11 @@ module Gen =
                 printfn "%A" (Tree.outcome shrink)
             printfn "."
 
+    module Operators =
+        let (<!>) f g = map f g
+        let (<*>) gf g = apply g gf
+        let (>>=) g f = bind f g
+
 [<AutoOpen>]
 module GenBuilder =
     let gen = Gen.Builder ()
-
-[<AutoOpen>]
-module GenOperators =
-    let (<!>) f g = Gen.map f g
-    let (<*>) gf g = Gen.apply g gf
-    let (>>=) g f = Gen.bind f g
--- a/src/Hedgehog/Gen.fs
+++ b/src/Hedgehog/Gen.fs
@@ -518,5 +518,5 @@ module GenBuilder =
 
 [<AutoOpen>]
 module GenOperators =
-    let (<!>) = Gen.map
-    let (<*>) = Gen.apply
+    let (<!>) f g = Gen.map f g
+    let (<*>) gf g = Gen.apply g gf
--- a/src/Hedgehog/Gen.fs
+++ b/src/Hedgehog/Gen.fs
@@ -30,7 +30,7 @@ module Gen =
     let constant (x : 'a) : Gen<'a> =
         Tree.singleton x |> Random.constant |> ofRandom
 
-    let private bindRandom (m : Random<Tree<'a>>) (k : 'a -> Random<Tree<'b>>) : Random<Tree<'b>> =
+    let private bindRandom (k : 'a -> Random<Tree<'b>>) (m : Random<Tree<'a>>) : Random<Tree<'b>> =
         Hedgehog.Random (fun seed0 size ->
             let seed1, seed2 =
                 Seed.split seed0
@@ -41,7 +41,7 @@ module Gen =
             Tree.bind (run seed1 m) (run seed2 << k))
 
     let bind (k : 'a -> Gen<'b>) (m : Gen<'a>) : Gen<'b> =
-        bindRandom (toRandom m) (toRandom << k) |> ofRandom
+        toRandom m |> bindRandom (toRandom << k) |> ofRandom
 
     let apply (gx : Gen<'a>) (gf : Gen<'a -> 'b>) : Gen<'b> =
         gf |> bind (fun f ->
--- a/src/Hedgehog/Gen.fs
+++ b/src/Hedgehog/Gen.fs
@@ -520,3 +520,4 @@ module GenBuilder =
 module GenOperators =
     let (<!>) f g = Gen.map f g
     let (<*>) gf g = Gen.apply g gf
+    let (>>=) g f = Gen.bind f g
--- a/src/Hedgehog/Gen.fs
+++ b/src/Hedgehog/Gen.fs
@@ -43,7 +43,7 @@ module Gen =
     let bind (k : 'a -> Gen<'b>) (m : Gen<'a>) : Gen<'b> =
         bindRandom (toRandom m) (toRandom << k) |> ofRandom
 
-    let apply (gf : Gen<'a -> 'b>) (gx : Gen<'a>) : Gen<'b> =
+    let apply (gx : Gen<'a>) (gf : Gen<'a -> 'b>) : Gen<'b> =
         gf |> bind (fun f ->
         gx |> bind (f >> constant))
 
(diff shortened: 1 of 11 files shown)


TysonMN:
The functions [`map2`, `map3`, and `map4` in `Gen`](https://github.com/hedgehogqa/fsharp-hedgehog/blob/7058b4c38c5b35aac1658894d01799207232d996/src/Hedgehog/Gen.fs#L59-L75) are not necessary.  Instead, `Gen.apply`
https://github.com/hedgehogqa/fsharp-hedgehog/blob/7058b4c38c5b35aac1658894d01799207232d996/src/Hedgehog/Gen.fs#L46-L48

should have its arguments reversed to
```fsharp
let apply (gx : Gen<'a>) (gf : Gen<'a -> 'b>) : Gen<'b> =
```

Then given
```fsharp
let flip f b a = f a b
let g = (Gen.constant 1)
```

we can replace
```fsharp
(fun a b c d e -> a + b + c + d + e)
|> flip Gen.map g
|> Gen.map4 g g g g
```

with
```fsharp
(fun a b c d -> a + b + c + d)
|> flip Gen.map g
|> Gen.apply g
|> Gen.apply g
|> Gen.apply g
```

Clearly the advantage of the latter is that it scales to a (curried) function with any number of inputs.

Just as I explained in https://github.com/hedgehogqa/fsharp-hedgehog/issues/265#issuecomment-758728308, the arguments to the function named `apply` should be flipped compared to the apply operator (which is denoted by [`<*>` in Haskell](http://hackage.haskell.org/package/base-4.14.1.0/docs/Control-Applicative.html)).

I am not actually requesting that `map2`, `map3`, and `map4` be removed.  I would remove them, but they can stay if others want them.

Instead, I am specifically requesting that the arguments to `Gen.apply` be swapped.

Also, I find it annoying to call `flip` before `map` in both of those cases shown above.  In my personal code, I thought of a solution that I like.  I created a function called `apply1` that is equivalent to `flip map`.  Then the code would look like
```fsharp
(fun a b c d -> a + b + c + d)
|> Gen.apply1 g
|> Gen.apply g
|> Gen.apply g
|> Gen.apply g
```

Therefore, I also suggest adding this `apply1` function.

adam-becker:
@TysonMN There are a few ways to skin that cat, you could also write the applicative chain like this:

```fsharp
let g = Gen.constant 1

Gen.success (fun a b c d -> a + b + c + d)
|> Gen.apply g
|> Gen.apply g
|> Gen.apply g
|> Gen.apply g
```

Or like this:

```fsharp
let (<*>) = Gen.apply
let (<!>) = Gen.map
let g = Gen.constant 1

(fun a b c d -> a + b + c + d) <!> g <*> g <*> g <*> g
```

`Gen.apply` would have to be adjusted to make this work, as you've noted. Is there a use case that `apply1` would cover that above doesn't?

Also on the topic of applicatives, we should probably add support for the new applicative syntax in F# 5 at some point.

TysonMN:
> Or like this:
> 
> ```fsharp
> let (<*>) = Gen.apply
> let (<!>) = Gen.map
> let g = Gen.constant 1
> 
> (fun a b c d -> a + b + c + d) <!> g <*> g <*> g <*> g
> ```

Indeed.  In fact, I didn't realize that those operators are defined.
https://github.com/hedgehogqa/fsharp-hedgehog/blob/7058b4c38c5b35aac1658894d01799207232d996/src/Hedgehog/Gen.fs#L507-L510

TysonMN:
> ```fsharp
> Gen.success (fun a b c d -> a + b + c + d)
> |> Gen.apply g
> |> Gen.apply g
> |> Gen.apply g
> |> Gen.apply g
> ```

I think you mean `Gen.constant` instead of `Gen.success`.  Yes, that would also work instead of using `apply1`.

adam-becker:
> I think you mean `Gen.constant` instead of `Gen.success`. Yes, that would also work instead of using `apply1`.

Ah, yes you're correct. It was literally in the previous snippet :smile:

adam-becker:
@TysonMN I've flipped the parameters for `apply`, then I forgot to update the `<*>` operator so I've fixed that as well and put some regression tests in place.

TysonMN:
Yes, those changes look good.  Thanks @adam-becker! :)

TysonMN (approved):

moodmosaic:
This branch has conflicts that must be resolved, then we can merge it.

adam-becker:
@moodmosaic Going to do a quick skim to see if there are other functions that need to be rearranged.

TysonMN:
Double check if the `map` operator has its inputs ordered correctly.  I think the desired order should allow
```fsharp
Gen.bool <!> not <!> not
```
while the function form would look like
```fsharp
Gen.bool |> Gen.map not |> Gen.map not
```

adam-becker:
> Double check if the `map` operator has its inputs ordered correctly. I think the desired order should allow
> 
> ```fsharp
> Gen.bool <!> not <!> not
> ```
> 
> while the function form would look like
> 
> ```fsharp
> Gen.bool |> Gen.map not |> Gen.map not
> ```

I've always seen `map` defined the exact same way as `<!>`. It allows for applicative programming this way, a la

```fsharp
let y = // Gen<int>
    (fun a b -> a + b) // int -> int -> int
    <!> Gen.constant 1 // Gen<int -> int>
    <*> Gen.constant 2 // Gen<int>
```

I'm not entirely sold on using the operator versions anyways, and I think they should definitely not be marked with `[<AutoOpen>]`.

I've looked for references, but almost everywhere I looked omitted the `<!>` operator. Instead I see the ordering you've suggested exposed as `|>>` (FParsec, FSharpPlus). This would turn your example into:

```fsharp
Gen.bool |>> not |>> not
```

We can consider adding this operator, but as I said above I think using these operators should be something a user opts in to.

TysonMN:
Ok.  You have definitely thought about this more than me.  I am fibre with whatever you want.  I don't use the operators anyway.

I also think not auto opening the operators is reasonable.

moodmosaic:
Agreed with @adam-becker. Infix operators in F# must be defined per type (unless you do some funky usage of statically resolved type parameter tricks), so using these operators should be only something a user opts in to.

adam-becker:
@moodmosaic I've made the change to opt-in to operators for `Gen`, I believe this branch is merge-able now.

moodmosaic (commented):
Looks good to me! Massive piece of work. If @TysonMN also agrees, we can merge 🚀

TysonMN:
Yep, looks great.  Excellent work @adam-becker! :)

Issue #271, opened 2021-01-15
Investigate Fantomas for consistent formatting


adam-becker:
Issues with formatting/style have cropped up a few times in PRs, so we might want to look into using an auto-formatting tool. Fantomas appears to be the front runner in this domain.

I'm curious what other members think (as well as contributors, of course). /cc @moodmosaic @TysonMN

moodmosaic:
I'd be surprised if Fantomas could handle this particular formatting/style. I'm all in for this, though, unless Tyson has a different opinion.

TysonMN:
When programming in C#, I am satisfied with the amount of `.editorconfig` settings supported by Visual Studio.  For F#, Visual Studio only supports a few settings, and I have not felt the need to in my F# programming to find an automated solution to enforce additional settings.

Even so, I still like having automated solutions for such things, so I am in too.

I have my preferences about formatting, but I care much more that the formatting is consistent.

I just created PR #273, which adds a `.editorconfig` file.

moodmosaic:
An alternative can be just a markdown file explaining what the style is, similar to https://github.com/hedgehogqa/haskell-hedgehog/blob/master/STYLE_GUIDE.md

dharmaturtle:
Both [Fantomas](https://github.com/fsprojects/fantomas/blob/master/docs/Documentation.md#fsharp_space_before_colon) and [FSharpLint](https://fsprojects.github.io/FSharpLint/reference/FSharpLint.Core/fsharplint-rules-typeditemspacing.html) support spaces before/after colons, which I believe was the instigating issue.

adam-becker:
Generally I think formatting tools go too far, and/or have less tolerance for certain things. I tried running [Fantomas][fantomas-live] on the `Gen.fs` file, and the results seemed sporadic, like `Gen (fun x -> ...` would be split across multiple lines sometimes but wouldn't be in other situations.

Another issue is that most of the code I've seen in this repo has a function body on a new line, whereas Fantomas will only do that for "complex" expressions, or expressions that are less then `n` characters. If the tool were "dumber" and just always formatted the code a certain way, I'd have a better feeling about it. I feel like they've gone to great lengths to make the tool more flexible in it's output rules, but in doing so it makes the formatted code artifact look frantic.

[fantomas-live]: https://fsprojects.github.io/fantomas-tools/#/fantomas/preview

moodmosaic:
>and the results seemed sporadic

[...]

PR #273, opened 2021-01-16
add .editorconfig file

TysonMN:
Partially addresses #271

[EditorConfig](https://editorconfig.org/) is a IDE-independent way of expressing some formatting settings.

The file contains my preferred settings.  Feel free to suggest different settings.  I care much more that a setting exists than whether it is set to my preference.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,7 @@
+root = true
+
+[*] # rules that follow apply to files with any extension
+insert_final_newline = false
+trim_trailing_whitespace = true
+indent_style = space
+indent_size = 4
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,7 +1,7 @@
 root = true
 
 [*] # rules that follow apply to files with any extension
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,7 @@
+root = true
+
+[*] # rules that follow apply to files with any extension
+insert_final_newline = false
```
moodmosaic, line 4: I think we're inserting a final new line everywhere, or in most places, but I could be wrong.

(I'm out and about at the moment so I can't check.)

moodmosaic (commented):
👍

.editorconfig
moodmosaic: If that's the case then this must be set to true.

TysonMN: Yep, I just checked.  I changed that setting.

moodmosaic: 👍

moodmosaic:
💯

PR #274, opened 2021-01-16
Add render functions

### LINKED ISSUE #226: toString versions of print functions

adam-becker:
Fixes #226 the name for `Gen.printSample` would be awkward (`Gen.toStringSample`) if the convention were followed.

/cc @moodmosaic @TysonMN

### DIFF
--- a/src/Hedgehog/Gen.fs
+++ b/src/Hedgehog/Gen.fs
@@ -502,7 +502,7 @@ module Gen =
         toRandom g
         |> Random.run seed 30
 
-    let formatSample (gen : Gen<'a>) : string =
+    let renderSample (gen : Gen<'a>) : string =
         String.concat Environment.NewLine [
             let forest = sampleTree 10 5 gen
             for tree in forest do
@@ -513,9 +513,6 @@ module Gen =
                     yield sprintf "%A" (Tree.outcome shrink)
                 yield "."
         ]
-
-    let printSample (gen : Gen<'a>) : unit =
-        printfn "%s" (formatSample gen)
 
     module Operators =
         let (<!>) f g = map f g
--- a/src/Hedgehog/Gen.fs
+++ b/src/Hedgehog/Gen.fs
@@ -502,15 +502,20 @@ module Gen =
         toRandom g
         |> Random.run seed 30
 
+    let asString (g : Gen<'a>) : string =
+        String.concat Environment.NewLine [
+            let forest = sampleTree 10 5 g
+            for tree in forest do
+                yield "=== Outcome ==="
+                yield sprintf "%A" (Tree.outcome tree)
+                yield "=== Shrinks ==="
+                for shrink in Tree.shrinks tree do
+                    yield sprintf "%A" (Tree.outcome shrink)
+                yield "."
+        ]
+
     let printSample (g : Gen<'a>) : unit =
-        let forest = sampleTree 10 5 g
-        for tree in forest do
-            printfn "=== Outcome ==="
-            printfn "%A" (Tree.outcome tree)
-            printfn "=== Shrinks ==="
-            for shrink in Tree.shrinks tree do
-                printfn "%A" (Tree.outcome shrink)
-            printfn "."
+        printfn "%s" (asString g)
 
     module Operators =
         let (<!>) f g = map f g
--- a/src/Hedgehog/Gen.fs
+++ b/src/Hedgehog/Gen.fs
@@ -502,9 +502,9 @@ module Gen =
         toRandom g
         |> Random.run seed 30
 
-    let asString (g : Gen<'a>) : string =
+    let formatSample (gen : Gen<'a>) : string =
         String.concat Environment.NewLine [
-            let forest = sampleTree 10 5 g
+            let forest = sampleTree 10 5 gen
             for tree in forest do
                 yield "=== Outcome ==="
                 yield sprintf "%A" (Tree.outcome tree)
@@ -514,8 +514,8 @@ module Gen =
                 yield "."
         ]
 
-    let printSample (g : Gen<'a>) : unit =
-        printfn "%s" (asString g)
+    let printSample (gen : Gen<'a>) : unit =
+        printfn "%s" (formatSample gen)
 
     module Operators =
         let (<!>) f g = map f g
--- a/src/Hedgehog/Property.fs
+++ b/src/Hedgehog/Property.fs
@@ -241,31 +241,36 @@ module Property =
     let recheckBool (size : Size) (seed : Seed) (g : Property<bool>) : unit =
         g |> bind ofBool |> recheck size seed
 
-    let printWith (config : PropertyConfig) (p : Property<unit>) : unit =
-        reportWith config p
-        |> Report.render
-        |> printfn "%s"
-
-    let formatWith (n : PropertyConfig) (p : Property<unit>) : string =
+    let renderWith (n : PropertyConfig) (p : Property<unit>) : string =
         reportWith n p
         |> Report.render
 
-    let format (p : Property<unit>) : string =
+    let render (p : Property<unit>) : string =
         report p
         |> Report.render
 
     let print (p : Property<unit>) : unit =
-        format p
+        render p
         |> printfn "%s"
 
-    let printBoolWith (config : PropertyConfig) (p : Property<bool>) : unit =
+    let renderBool (property : Property<bool>) : string =
+        reportBool property
+        |> Report.render
+
+    let renderBoolWith (config : PropertyConfig) (p : Property<bool>) : string =
         reportBoolWith config p
         |> Report.render
+
+    let printBoolWith (config : PropertyConfig) (p : Property<bool>) : unit =
+        renderBoolWith config p
+        |> printfn "%s"
+
+    let printWith (config : PropertyConfig) (p : Property<unit>) : unit =
+        renderWith config p
         |> printfn "%s"
 
     let printBool (p : Property<bool>) : unit =
-        reportBool p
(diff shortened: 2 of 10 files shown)


src/Hedgehog/Property.fs
```diff
@@ -418,15 +418,17 @@ module Property =
...
+        Report.render <| report' n p
+
+    let toString (p : Property<unit>) : string =
+        Report.render <| report p
+
     let print' (n : int<tests>) (p : Property<unit>) : unit =
-        report' n p
-        |> Report.render
-        |> printfn "%s"
+        printfn "%s" <| toString' n p
```
moodmosaic, line 428: Just a thought... Perhaps `asString` instead of `toString`? (/cc @TysonMN) — In the Hedgehog.Linq namespace, for example, `ToString` may cause confusion because of the built-in [ToString](https://docs.microsoft.com/en-us/dotnet/api/system.object.tostring) that all types inherit.

TysonMN: What is it called in the Haskell code? `render`?

dharmaturtle: Overriding `ToString` is actually recommended in C#. From [the docs](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/how-to-override-the-tostring-method):

>When you create a custom class or struct, you should override the ToString method in order to provide information about your type to client code.

I think `toString`/`ToString` is fine, and since we're in dotnet land, I think we should work within their conventions (and not Haskell's - at least in this matter).

TysonMN: I mostly indifferent about this naming.  What @dharmaturtle says sounds good to me.

adam-becker: Maybe we should just override `ToString` for these types, and not make it part of the LINQ interface?

TysonMN: Then I no longer want to call this `ToString`.  I want to prioritise #F conventions (like separating types from behavior) over C# ones (like combing types and behavior).

adam-becker: Renamed all instances to `asString`/`AsString`

dharmaturtle: [Someone else ran into this problem.](https://github.com/fsharp/fslang-suggestions/issues/691)

I'm fine with `asString`.

src/Hedgehog/Property.fs
```diff
@@ -418,15 +418,17 @@ module Property =
     let recheckBool (size : Size) (seed : Seed) (g : Property<bool>) : unit =
         bind g ofBool |> recheck size seed
 
+    let asString' (n : int<tests>) (p : Property<unit>) : string =
+        Report.render <| report' n p
```
moodmosaic, line 422: Do we really want to keep using `<|` in our code? https://youtu.be/1AZA1zoP-II?t=2130

src/Hedgehog/Gen.fs
```diff
@@ -490,15 +490,20 @@ module Gen =
         toRandom g
         |> Random.run seed 30
 
+    let asString (g : Gen<'a>) : string =
```
moodmosaic, line 493: This looks good but I believe it can also be given some more thought. 

I would expect most users to understand that *print* + *sample* does exactly that:
* *print* (to the console) a *sample* of `g`

Now, if we introduce `asString` then it *may* cause confusion (IME a small % of the actual users are following each and every PR). So as a user that just upgraded to latest Hedgehog I would expect that `asString`:
* it returns a string as `g` in its fully qualified name?
* it returns a string representing some structure of `g` prior being run/sampled/shrinked?

---

Perhaps *sample* is the name we would want, after all?

src/Hedgehog/Property.fs
moodmosaic: > What is it called in the Haskell code? `render`?

IIRC, it's called `sample` the equivalent of the one we're tackling here.

src/Hedgehog/Property.fs
adam-becker: There are plenty of other spots in the code where they are used, so mostly sticking to convention. I agree though that we should probably avoid it, and use `x |> f` or `(f x)` instead. Should I tackle that in this PR or put up another quick one after this?

src/Hedgehog/Gen.fs
adam-becker: `sample` is already defined, though. :disappointed:

src/Hedgehog/Property.fs
adam-becker: I've created #276 for this.

src/Hedgehog/Gen.fs
TysonMN: I suggest we remove this function (called `Gen.printSample` in `master` and `Gen.asString` in this PR).

Instead, move the body of the `for` loop to a new function called `Tree.asString`.  Then users can easily implement this function themselves if they really want to see exactly five shrink tree printed to standard out.

adam-becker: I renamed these functions to `format`, `formatWith`, and `formatSample`.

src/Hedgehog/Gen.fs
```diff
@@ -502,15 +502,20 @@ module Gen =
...
-    let printSample (g : Gen<'a>) : unit =
-        let forest = sampleTree 10 5 g
-        for tree in forest do
-            printfn "=== Outcome ==="
-            printfn "%A" (Tree.outcome tree)
-            printfn "=== Shrinks ==="
-            for shrink in Tree.shrinks tree do
-                printfn "%A" (Tree.outcome shrink)
-            printfn "."
+    let formatSample (gen : Gen<'a>) : string =
```
moodmosaic, line 505: I would expect this to be taking a [format provider](https://docs.microsoft.com/en-us/dotnet/api/system.iformatprovider?view=net-5.0) since it's called format(?) :thinking:

adam-becker: &lt;insert joke about naming in computer science here&gt;

We're getting to the point where all we have left are silly names:

- `toString`/`asString`/`stringify`
- `toPretty`/`asPretty`/`prettify`
- `toDisplay`/`asDisplay`/`displayify` (that was intended as a joke)

Do we maybe want to expose a function that uses `IFormatProvider`?

TysonMN: How about `render`?

moodmosaic: `render` is closer to Haskell Hedgehog naming so I think it's a good option.

adam-becker: @moodmosaic @TysonMN They are now all a variation of `render`.

src/Hedgehog/Gen.fs
```diff
@@ -502,15 +502,20 @@ module Gen =
...
+            for tree in forest do
+                yield "=== Outcome ==="
+                yield sprintf "%A" (Tree.outcome tree)
+                yield "=== Shrinks ==="
+                for shrink in Tree.shrinks tree do
+                    yield sprintf "%A" (Tree.outcome shrink)
+                yield "."
+        ]
+
+    let printSample (gen : Gen<'a>) : unit =
```
TysonMN, line 517: I think we should drop the `print*` family of functions

adam-becker: @moodmosaic Thoughts on this?

src/Hedgehog/Gen.fs
moodmosaic: > How about `render`?

👍

src/Hedgehog/Gen.fs
moodmosaic: I believe we should, yes. After all they're one liners e.g. `printfn "%s" (render gen)` (btw, it's still formatSample in the diff, I thought we're going with `render` instead?)

moodmosaic: Edit: Now I see, `render` is on the property level.

moodmosaic: https://github.com/hedgehogqa/fsharp-hedgehog/pull/274/files#r576824359

adam-becker: Dropped the `print*` variations, and renamed the missed functions to `render*`.

Do we want `Gen.renderSample` to just be `Gen.render`?

TysonMN: I don't think so.  `Sample` has meaning here.  It is not the entire `Gen`<_>` that is being rendered.  Just a few samples (5 or 10, I forgot which argument is which).

TysonMN (approved):

moodmosaic (approved):
👍

PR #275, opened 2021-01-17
Hide unit type from LINQ interface

adam-becker:
The purpose of this PR is to hide another F# detail from the LINQ interface, in this case the `unit` type. When used as a return type, `unit` becomes `void`, but when used as a type argument then it remains as `FSharp.Core.Unit`. To solve this I've introduced the `Property` type, that just conceals a `Property<unit>` (ie, this is _not_ an alias). Then from C#/VB a user would just see this simple type instead of `Property<FSharp.Core.Unit>`.

/cc @moodmosaic @TysonMN

### DIFF
--- a/src/Hedgehog/Linq/Property.fs
+++ b/src/Hedgehog/Linq/Property.fs
@@ -6,28 +6,22 @@ open System
 open System.Runtime.CompilerServices
 open Hedgehog
 
-type PropertyThrows =
-    private
-        PropertyThrows of Property<unit>
+type Property = private Property of Property<unit> with
 
-[<Extension>]
-[<AbstractClass; Sealed>]
-type Property private () =
-
-    static member Failure : PropertyThrows =
+    static member Failure : Property =
         Property.failure
-        |> PropertyThrows
+        |> Property
 
-    static member Discard : PropertyThrows =
+    static member Discard : Property =
         Property.discard
-        |> PropertyThrows
+        |> Property
 
     static member Success (value : 'T) : Property<'T> =
         Property.success value
 
-    static member FromBool (value : bool) : PropertyThrows =
+    static member FromBool (value : bool) : Property =
         Property.ofBool value
-        |> PropertyThrows
+        |> Property
 
     static member FromGen (gen : Gen<Journal * Result<'T>>) : Property<'T> =
         Property.ofGen gen
@@ -35,9 +29,9 @@ type Property private () =
     static member FromResult (result : Result<'T>) : Property<'T> =
         Property.ofResult result
 
-    static member FromThrowing (throwingFunc : Action<'T>, arg : 'T) : PropertyThrows =
+    static member FromThrowing (throwingFunc : Action<'T>, arg : 'T) : Property =
         Property.ofThrowing throwingFunc.Invoke arg
-        |> PropertyThrows
+        |> Property
 
     static member Delay (f : Func<Property<'T>>) : Property<'T> =
         Property.delay f.Invoke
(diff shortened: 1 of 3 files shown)


src/Hedgehog/Linq/Property.fs
```diff
@@ -6,39 +6,48 @@ open System
 open System.Runtime.CompilerServices
 open Hedgehog
 
+type PropertyThrows =
```
moodmosaic, line 9: Can't we just named this `Property` and play with attributes as in https://github.com/hedgehogqa/fsharp-hedgehog/blob/f0ea8a1a35eefd7ae0f5734fdef5eb1d3b75064e/src/Hedgehog/Seed.fs#L37?

adam-becker: I was concerned that the names would collide in odd ways, but I was able to ride the line here and got the name `Property` to work. The API will contain:

- `Property`: Holds static factory functions (`ForAll`, `Success`, etc), and is the wrapper for `Property<unit>`.
- `PropertyExtensions`: Contains extension methods that target a `Property`/`Property<'a>`.
- `Property<'a>`: The type declared in `src/Hedgehog/Property.fs`.

TysonMN: I like using `Property` in place of `Property<unit>` in the C# code.

src/Hedgehog/Linq/Property.fs
```diff
@@ -6,46 +6,53 @@ open System
...
 
     static member ForAll (gen : Gen<'T>, k : Func<'T, Property<'TResult>>) : Property<'TResult> =
         Property.forAll gen k.Invoke
 
     static member ForAll (gen : Gen<'T>) : Property<'T> =
         Property.forAll' gen
 
+[<Extension>]
+[<AbstractClass; Sealed>]
+type PropertyExtensions private () =
```
moodmosaic, line 54: The only disadvantage with `PropertyExtensions` is that the rest of the classes (Range, Gen) in Hedgehog.Linq don't have the *Extensions* suffix — (fortunately).

.NET extensions are sometimes considered as a [*feature envy*](https://twitter.com/ploeh/status/253426390718881792). If we'd want to be consistent I'd suggest we consider naming this `PropertyEnvy`, and then `GenEnvy`, `RangeEnvy`.

These can be ugly names for some people going through the codebase, so perhaps it's worth trying a trick like this
https://github.com/hedgehogqa/fsharp-hedgehog/blob/f0ea8a1a35eefd7ae0f5734fdef5eb1d3b75064e/src/Hedgehog/Seed.fs#L37

The rest of the PR looks good, and I don't want this comment of mine to be a blocker issue. If @adam-becker or @TysonMN have a different opinion we can consider that as well.

TysonMN: I think the name `PropertyExtension` is fine.  This is very standard in C#.

For consistency, we can also move the `Gen<_>` extensions and `Range<_>` extension to `GenExtensions` and `RangeExtensions`.

moodmosaic: OK, sounds like a plan.

moodmosaic: So are we ready to merge or do we want to rename also the types inside /Linq/Gen.fs and /Ling/Range.fs so that they end with *Extensions*? (I wouldn't yet rename the files, because we might end up adding more stuff to those as it happened here.)

TysonMN: Making that change in another PR is fine with me.

And I wasn't thinking just a rename.  I was thinking that the extensions in `Gen` move to `GenExtensions` and the non-extensions stay on `Gen`.

moodmosaic: Thanks! Let's wait for @adam-becker (just in case he plans to do that in this PR) before we can merge.

adam-becker: Thanks for waiting! I went ahead and made this change as well.

adam-becker:
@moodmosaic This has been rebased and is ready to merge.

moodmosaic (commented):
:+1: :100: :1st_place_medal:

PR #276, opened 2021-01-18
Remove backward pipe

adam-becker:
As part of a discussion on #274 there was a question about removing usages of the `<|` operator. This PR demonstrates what that would look like, and is also merge-able.

/cc @moodmosaic @TysonMN


doc/tutorial.md
```diff
@@ -27,18 +27,18 @@ One way to use Hedgehog to check the above property is to use the `property` com
 
 ```fs
 property {
-    let! xs = Gen.list (Range.linear 0 100) <| Gen.int (Range.constant 0 1000)
+    let! xs = Gen.list (Range.linear 0 100) (Gen.int (Range.constant 0 1000))
```
TysonMN, line 30: I think piping right is more readable because then the code is read in the order it is executed.  In this case, I prefer
```fsharp
let! xs = Range.constant 0 1000 |> Gen.int |> Gen.list (Range.linear 0 100)
```

adam-becker: Agreed, there are going to be quite a few of those to visit and fix. I'll get to that tomorrow most likely :smile:

src/Hedgehog/Gen.fs
```diff
@@ -242,68 +245,70 @@ module Gen =
...
             | Some x ->
-                Tree.map Some x |> Random.constant
+                Random.constant (Tree.map Some x)
+
+        let flipBind f ma = Random.bind ma f
+
+        toRandom g
+        |> tryFilterRandom p
+        |> flipBind filter
+        |> ofRandom
```
moodmosaic, line 282: Are we sure this one works the same (I don't have a compiler in front of me right now because I'm travelling). The safest would be to just add `( .. )` instead of `<|`.

src/Hedgehog/Gen.fs
```diff
@@ -242,68 +245,70 @@ module Gen =
...
 
     /// Runs an option generator until it produces a 'Some'.
     let some (g : Gen<'a option>) : Gen<'a> =
-        bind (filter Option.isSome g) <| function
-        | Some x ->
-            constant x
-        | None ->
-            invalidOp "internal error, unexpected None"
+        bind (filter Option.isSome g) (Option.get >> constant)
```
moodmosaic, line 286: :+1: — IIRC, `get` throws an `invalidArg` instead of an `invalidOp` but that should be OK as a change (as long as we're v0.x).

src/Hedgehog/Gen.fs
```diff
@@ -416,27 +422,28 @@ module Gen =
...
 
     /// Generates a random 64-bit floating point number.
     let float (range : Range<float>) : Gen<float> =
-        (double range) |> map float
+        double range |> map float
 
     /// Generates a random 32-bit floating point number.
     let single (range : Range<single>) : Gen<single> =
-      double (Range.map ExtraTopLevelOperators.double range) |> map single
+        double (Range.map ExtraTopLevelOperators.double range) |> map single
```
moodmosaic, line 434: Good catch :+1:

src/Hedgehog/Gen.fs
```diff
@@ -416,27 +422,28 @@ module Gen =
...
 
     /// Generates a random decimal floating-point number.
     let decimal (range : Range<decimal>) : Gen<decimal> =
-      double (Range.map ExtraTopLevelOperators.double range) |> map decimal
+        double (Range.map ExtraTopLevelOperators.double range) |> map decimal
```
moodmosaic, line 438: Good catch :+1:

moodmosaic (commented):
Looks good. We could merge after you're done with @TysonMN's [suggestion](https://github.com/hedgehogqa/fsharp-hedgehog/pull/276#discussion_r559318411) (or when you and Tyson are ready). :100:

src/Hedgehog/Gen.fs
```diff
@@ -223,12 +223,15 @@ module Gen =
...
     /// <i>The first argument (i.e. the non-recursive input list) must be non-empty.</i>
     let choiceRec (nonrecs : seq<Gen<'a>>) (recs : seq<Gen<'a>>) : Gen<'a> =
-        sized <| fun n ->
+        sized (fun n ->
             if n <= 1 then
                 choice nonrecs
             else
-                let halve x = x / 2
-                choice <| Seq.append nonrecs (Seq.map (scale halve) recs)
+                recs
```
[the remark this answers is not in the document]
TysonMN: If both branches could start with `choice` then I think it is best when neither branch even contains `choice`.  I would implement it as
```fsharp
if n <= 1 then
    Seq.empty
else
    recs |> Seq.map (scale (fun x -> x / 2))
|> Seq.append nonrecs
|> choice
```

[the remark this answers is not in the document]
moodmosaic: Very good point.

src/Hedgehog/Gen.fs
TysonMN: > Are we sure this one works the same

They are the same.  Here is a proof.  All lines are equivalent.
```fsharp
ofRandom << Random.bind gen <| f
(ofRandom << Random.bind gen) <| f
f |> (ofRandom << Random.bind gen)
f |> (Random.bind gen >> ofRandom)
f |> Random.bind gen |> ofRandom
Random.bind gen f |> ofRandom
flip Random.bind f gen |> ofRandom
gen |> flip Random.bind f |> ofRandom
```

Frist,
```fsharp
function
| None -> None |> Tree.singleton
| Some t -> t |> Tree.map Some
```
is also known as `OptionTree.sequence` (according to [how I name nested containers](https://github.com/hedgehogqa/fsharp-hedgehog/pull/260#discussion_r551046128), or `Option.Tree.sequence`).

This can be added in a separate PR.

Second, let's add
```fsharp
[<AutoOpen>]
module internal AutoOpen =
    let flip f b a = f a b
```

This can also be in a separate PR.

Then after completing those PRs and rebasing the branch in this PR, I would implement this as
```fsharp
toRandom g
|> tryFilterRandom p
|> flip Random.bind (OptionTree.sequence >> Random.constant)
|> ofRandom
```

As a reminder the need for `flip` there is being removed in PR #266.

src/Hedgehog/Gen.fs
adam-becker: I made this change because in actuality this should never throw anyways, but you are correct it's a different exception.

src/Hedgehog/Gen.fs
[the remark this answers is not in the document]
adam-becker: @TysonMN the only change I'd make from your version is I'd put bind the `if`/`else` expression to a name first, then pass that to the pipeline.

src/Hedgehog/Gen.fs
adam-becker: I'm definitely in favor of using the `|>` here, since the previous version was very difficult to grok.

adam-becker:
I think this is ready for another round of review.

src/Hedgehog/Gen.fs
TysonMN: Yes, I was also thinking that it would never throw.

src/Hedgehog/Gen.fs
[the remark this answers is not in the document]
TysonMN: That's fine too

doc/tutorial.md
```diff
@@ -251,10 +251,10 @@ let version =
     |> Gen.tuple3
     |> Gen.map (fun (ma, mi, bu) -> Version (ma, mi, bu))
 
-Property.print <| property {
+Property.print (property {
     let! xs = Gen.list (Range.linear 0 100) version
     return xs |> List.rev = xs
-    }
+})
```
TysonMN, line 257: I would do either
```fsharp
property {
} |> Property.print
```
or
```fsharp
property {
}
|> Property.print
```
but this is fine too.

There are other cases like this.  I won't comment on them until we decide what is preferred here.

doc/tutorial.md
```diff
@@ -631,8 +630,8 @@ let tryAdd a b =
     if a > 100 then None // Nasty bug.
     else Some (a + b)
 
-property { let! a = Gen.int <| Range.constantBounded ()
-           let! b = Gen.int <| Range.constantBounded ()
+property { let! a = Gen.int (Range.constantBounded ())
+           let! b = Gen.int (Range.constantBounded ())
```
TysonMN, line 634:
```suggestion
property { let! a = Range.constantBounded () |> Gen.int
           let! b = Range.constantBounded () |> Gen.int
```

doc/tutorial.md
```diff
@@ -658,8 +657,8 @@ let tryAdd a b =
     if a > 100 then None // Nasty bug.
     else Some(a + b)
 
-property { let! a = Gen.int <| Range.constantBounded ()
-           let! b = Gen.int <| Range.constantBounded ()
+property { let! a = Gen.int (Range.constantBounded ())
+           let! b = Gen.int (Range.constantBounded ())
```
TysonMN, line 661:
```suggestion
property { let! a = Range.constantBounded () |> Gen.int
           let! b = Range.constantBounded () |> Gen.int
```

doc/tutorial.md
```diff
@@ -681,8 +680,8 @@ let tryAdd a b =
...
```
TysonMN, line 684:
```suggestion
property { let! a = Range.constantBounded () |> Gen.int
           let! b = Range.constantBounded () |> Gen.int
```

src/Hedgehog/Gen.fs
```diff
@@ -179,7 +183,7 @@ module Gen =
         if Array.isEmpty xs then
             return crashEmpty "xs"
         else
-            let! ix = integral <| Range.constant 0 (Array.length xs - 1)
+            let! ix = integral (Range.constant 0 (Array.length xs - 1))
```
TysonMN, line 186:
```suggestion
            let! ix = Array.length xs - 1 |> Range.constant 0 |> integral
```

src/Hedgehog/Gen.fs
```diff
@@ -202,7 +206,7 @@ module Gen =
                 else
                     pick (n - k) ys
 
-        let! n = integral <| Range.constant 1 total
+        let! n = integral (Range.constant 1 total)
```
TysonMN, line 209:
```suggestion
        let! n = Range.constant 1 total |> integral
```

src/Hedgehog/Gen.fs
```diff
@@ -213,7 +217,7 @@ module Gen =
         if Array.isEmpty xs then
             return crashEmpty "xs" xs
         else
-            let! ix = integral <| Range.constant 0 (Array.length xs - 1)
+            let! ix = integral (Range.constant 0 (Array.length xs - 1))
```
TysonMN, line 220:
```suggestion
            let! ix = Array.length xs - 1 |> Range.constant 0 |> integral
```

src/Hedgehog/Gen.fs
```diff
@@ -242,68 +252,70 @@ module Gen =
...
         loop ()
         |> ofRandom
 
     /// Tries to generate a value that satisfies a predicate.
     let tryFilter (p : 'a -> bool) (g : Gen<'a>) : Gen<'a option> =
-        ofRandom << Random.bind (toRandom g |> tryFilterRandom p) <| function
+        let filter = function
             | None ->
-                None |> Tree.singleton |> Random.constant
+                Random.constant (Tree.singleton None)
```
TysonMN, line 280:
```suggestion
                Tree.singleton None |> Random.constant
```

src/Hedgehog/Gen.fs
```diff
@@ -416,27 +429,28 @@ module Gen =
...
+        double (Range.map ExtraTopLevelOperators.double range) |> map decimal
 
     //
     // Combinators - Constructed
     //
 
     /// Generates a random globally unique identifier.
     let guid : Gen<Guid> = gen {
-        let! bs = array (Range.constant 16 16) (byte <| Range.constantBounded ())
+        let! bs = array (Range.singleton 16) (byte (Range.constantBounded ()))
```
TysonMN, line 453:
```suggestion
        let! bs = Range.constantBounded () |> byte |> array (Range.singleton 16)
```

tests/Hedgehog.Tests/MinimalTests.fs
```diff
@@ -53,19 +53,23 @@ let rec tryFindSmallest (p : 'a -> bool) (Node (x, xs) : Tree<'a>) : 'a option =
...
+            Var <!> genName
+        ]
+
+        let nonrecs = [
+            Lam <!> Gen.zip genName genExp
+            App <!> Gen.zip genExp genExp
+        ]
+
+        let choiceRec = Gen.choiceRec recs nonrecs
+        Gen.shrink shrinkExp choiceRec) // comment this out to see the property fail
```
TysonMN, line 68:
```suggestion
        Gen.choiceRec recs nonrecs
        |> Gen.shrink shrinkExp // comment this out to see the property fail
    )
```

TysonMN (approved):

src/Hedgehog/Gen.fs
TysonMN: I think that was the newly added `.editorconfig` file.  There were some other lines changed because white space was removed from the end of the line.

src/Hedgehog/Gen.fs
adam-becker: I'm not sure I like this piping, since arithmetic is being used before the pipe. Maybe that sub-expression can be given a name?

TysonMN: Not a big deal.  You can leave it as it.

doc/tutorial.md
adam-becker: I think I would prefer

```fsharp
let name =
    property {
        // ..
    }
Property.print name
```

Often times F# code plays chicken with the off-side rule and I'd rather just be blunt and direct.

This is actually one situation where I think `<|` makes perfect sense to use. Passing a CE directly to a function as the last (or only) parameter, when no other piping is being used.

```fsharp
Property.print <| property {
    // ...
}
```

src/Hedgehog/Gen.fs
adam-becker: I made the range construction part of the range module, under a function called `Range.ofArray`. This function just creates a `constant 0 (Array.length xs - 1)` as a helper, but it also cleaned up this part entirely. With the added bonus of removing duplication. :+1: 

The new function is marked `internal` until we decide to expose it.

doc/tutorial.md
TysonMN: I am fine with anything here

src/Hedgehog/Range.fs
```diff
@@ -81,6 +81,13 @@ module Range =
 
         constantFrom zero lo hi
 
+    //
+    // Factories
+    //
+
+    let internal ofArray (xs : 'a array) =
```
moodmosaic, line 88:
```suggestion
    let internal ofArray (xs : 'a array) : Range<int> =
```

* And since it's monomorphic, can we move it inside Gen.fs?

(You've already done a massive work, so I'm happy to do that change myself if you're tired.)

adam-becker: I think this would make more sense as a `Range` function, but it can live wherever.

moodmosaic: Well, actually, I'm not sure it even makes sense as a function at all since it's used in [less than 3 places](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) but also [explicit is better than implicit](https://www.python.org/dev/peps/pep-0020/). But having it inside Gen.fs will make it easier for those [reading](https://www.goodreads.com/quotes/835238-indeed-the-ratio-of-time-spent-reading-versus-writing-is) the code, at least.

adam-becker: `/grump on`

In the wikipedia article you linked, there's a "See Also" for the DRY principle, which I'm more likely to follow in my PRs. I've been in software for the better part of two decades and had never heard of the "Rule of three", as I find any repetition in my code abhorrent.

I don't see what a python manifesto has to do with an F# code base either :smile:

`/grump off`

Kidding aside though, I believe the best place for this function is in the `Range` module, since it's a constructor function for a `Range` that might be useful to users as well. The discussion around exposing this is separate though, so my vote would be to just leave it where it is but alas I am only a single vote :sweat_smile: 

As far as readability goes, plenty of other functions in `Gen.fs` call over to functions in `Range.fs` and readability hasn't been hindered for me in that regard. As long as the function being called has a sensible name, I can make a quick assumption as to its behavior as I'm reading. In my mind, putting functions where they don't belong was a big motivating factor for #247, where a lot of things had coalesced in the `Property.fs` file.

Perhaps we need @TysonMN to break the tie! :tada: 

@moodmosaic I appreciate the feedback as always :+1:

adam-becker:
Pipeline failed because of some random "Unable to download dependencies" issue, is there a way to re-run the build?

moodmosaic:
I just rerun the pipeline. 👍

src/Hedgehog/Range.fs
moodmosaic: >but alas I am only a single vote 

A single vote with enough power at this point. You've done a massive work. Feel free to leave it there as-is. I shouldn't be nitpicking that much. 😥

moodmosaic:
Alrighty, ready to merge?

TysonMN:
Yep

moodmosaic:
:rocket:

PR #282, opened 2021-01-20
Add OptionTree.traverse

TysonMN:
This PR adds `OptionTree.traverse`.  It uses the same naming convention I have been arguing for in PR https://github.com/hedgehogqa/fsharp-hedgehog/pull/260#discussion_r551046128.

However, to hopefully avoid the same discussion here, I made that module internal.  That way, we can easily change the name (i.e. without a breaking change).

### DIFF
--- a/src/Hedgehog/OptionTree.fs
+++ b/src/Hedgehog/OptionTree.fs
@@ -1,0 +1,11 @@
+﻿[<RequireQualifiedAccess>]
+module internal Hedgehog.OptionTree
+
+
+let traverse (f: 'a -> Tree<'b>) (maybeTree: 'a option) : Tree<'b option> =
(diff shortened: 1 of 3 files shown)


src/Hedgehog/OptionTree.fs
```diff
@@ -0,0 +1,11 @@
+﻿[<RequireQualifiedAccess>]
+module internal Hedgehog.OptionTree
+
+
```
adam-becker, line 4: Just an extra line here, everything else looks fine.

adam-becker (approved):

src/Hedgehog/OptionTree.fs
TysonMN: Removed

Issue #283, opened 2021-01-20
Flaky tests


adam-becker:
Some of our tests can fail randomly, to make CI more reliable we should address these.

```
X Hedgehog.Linq.Tests.LinqTests.CanUseWhereWithBool [28ms]
  Error Message:
   System.Exception : *** Gave up after 100 discards, passed 17 tests.
  Stack Trace:
     at Hedgehog.ReportModule.tryRaise(Report report) in D:\a\fsharp-hedgehog\fsharp-hedgehog\src\Hedgehog\Report.fs:line 107
   at Hedgehog.Linq.PropertyExtensions.Check(Property`1 property, Int32 tests) in D:\a\fsharp-hedgehog\fsharp-hedgehog\src\Hedgehog\Linq\Property.fs:line 106
   at Hedgehog.Linq.Tests.LinqTests.CanUseWhereWithBool() in D:\a\fsharp-hedgehog\fsharp-hedgehog\tests\Hedgehog.Linq.Tests\LinqTests.cs:line 104
```

The test above fails because it exhausted the allowed discards. The way to fix this would be to use a `Gen` to always gives us the data we want, and "filter" that to ensure that we can use `where`, as the test mandates.

/cc @moodmosaic @TysonMN

moodmosaic:
Yep, we shouldn't have [erratic tests](http://xunitpatterns.com/Erratic%20Test.html) :+1:

TysonMN:
@cmeeren described a similar issue in #154.  I provided a solution to his problem in PR https://github.com/hedgehogqa/fsharp-hedgehog-experimental/pull/42.  I suggest we remove our test flakiness by using the same approach.

PR #284, opened 2021-01-20
Update remote dev files

adam-becker:
This PR addresses some issues/changes with the project, just a quick overview:

- Use a `docker-compose.yml` file to mount in a volume to cache vscode extensions.
- Switch to using the `.editorconfig` that was added in #273.
- Ionide version 5 switched to requiring .NET 5, which this project doesn't yet support. So I've pinned the last version that works with .NET Core 3.0.
- Also, there's an issue with these extensions working across language barriers, this part I can remove. I changed `Hedgehog.Linq.Tests.csproj` to reference the local path for `Hedgehog.dll` so that omnisharp can find it. This was more useful when I was writing the LINQ interface.

### DIFF
--- a/tests/Hedgehog.Linq.Tests/Hedgehog.Linq.Tests.csproj
+++ b/tests/Hedgehog.Linq.Tests/Hedgehog.Linq.Tests.csproj
@@ -15,5 +15,8 @@
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\..\src\Hedgehog\Hedgehog.fsproj" />
+    <Reference Include="Hedgehog">
+      <HintPath>..\..\src\Hedgehog\bin\Debug\netstandard2.0\Hedgehog.dll</HintPath>
+    </Reference>
   </ItemGroup>
 </Project>
--- a/tests/Hedgehog.Linq.Tests/Hedgehog.Linq.Tests.csproj
+++ b/tests/Hedgehog.Linq.Tests/Hedgehog.Linq.Tests.csproj
@@ -15,8 +15,5 @@
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\..\src\Hedgehog\Hedgehog.fsproj" />
-    <Reference Include="Hedgehog">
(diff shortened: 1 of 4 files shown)


adam-becker:
I'm really getting bit by #283 lately.

moodmosaic:
Build failed. I just rerun it.

tests/Hedgehog.Linq.Tests/Hedgehog.Linq.Tests.csproj
```diff
@@ -15,5 +15,8 @@
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\..\src\Hedgehog\Hedgehog.fsproj" />
+    <Reference Include="Hedgehog">
+      <HintPath>..\..\src\Hedgehog\bin\Debug\netstandard2.0\Hedgehog.dll</HintPath>
```
moodmosaic, line 19: Would it be possible to add a comment here so that others can see the reason we're adding a direct DLL reference? 👀

We can merge this as-is but the comment can help everyone else who hasn't paid attention to this PR.

moodmosaic (approved):

tests/Hedgehog.Linq.Tests/Hedgehog.Linq.Tests.csproj
adam-becker: I can just remove this, it's a hack to get a specific tool to work. Thinking about it, this doesn't need to be part of the mainline.

moodmosaic: It's up to you. Unless @TysonMN has a strong opinion on this.

TysonMN: I am all for improving the experience in various IDEs (especially an IDE of our most active contributor!), but this doesn't feel like the right solution to me.  If you still want to pursue this, then I suggest creating a new issue and providing more details about the problem.

PR #285, opened 2021-01-22
Housekeeping

moodmosaic:
Feel free to reject any of those cleanups. /cc @adam-becker @TysonMN 

:rescue_worker_helmet:

### DIFF
--- a/src/Hedgehog/Outcome.fs
+++ b/src/Hedgehog/Outcome.fs
@@ -7,14 +7,19 @@ type Outcome<'a> =
 
 module Outcome =
 
-    let cata result failure discard success =
-        match result with
+    let private cata
+        (outcome : Outcome<'a>)
+        (failure : unit -> 'b)
+        (discard : unit -> 'b)
+        (success :   'a -> 'b)
+            : 'b =
+        match outcome with
         | Failure ->
-            failure()
+            failure ()
         | Discard ->
-            discard()
-        | Success(x) ->
-            success(x)
+            discard ()
+        | Success x ->
+            success x
 
     [<CompiledName("Map")>]
     let map (f : 'a -> 'b) (result : Outcome<'a>) : Outcome<'b> =
@@ -27,7 +32,7 @@ module Outcome =
     let filter (f : 'a -> bool) (result : Outcome<'a>) : Outcome<'a> =
         let successOrDiscard x =
             if f x then
-                Success(x)
+                Success x
             else
                 Discard
 
--- a/src/Hedgehog/Outcome.fs
+++ b/src/Hedgehog/Outcome.fs
@@ -11,8 +11,7 @@ module Outcome =
         (outcome : Outcome<'a>)
         (failure : unit -> 'b)
         (discard : unit -> 'b)
-        (success :   'a -> 'b)
-            : 'b =
+        (success : 'a -> 'b) : 'b =
         match outcome with
         | Failure ->
             failure ()
--- a/src/Hedgehog/Seed.fs
+++ b/src/Hedgehog/Seed.fs
@@ -145,10 +145,10 @@ module Seed =
                 seed |> nextBigInt
                     (bigint Int32.MinValue)
                     (bigint Int32.MaxValue)
-            let scaled_x =
+            let scaledX =
                   (0.5 * lo + 0.5 * hi) + ((0.5 * hi - 0.5 * lo) / (0.5 * 4294967296.0)) * float x
 
-            scaled_x, seed'
+            scaledX, seed'
 
     /// Splits a random number generator in to two.
     let split (s0 : Seed) : Seed * Seed =
--- a/src/Hedgehog/Seed.fs
+++ b/src/Hedgehog/Seed.fs
@@ -46,24 +46,6 @@ module Seed =
     let [<Literal>] private goldenGamma : uint64 =
         0x9e3779b97f4a7c15UL
 
-    let private crashUnless (cond : bool) (msg : string) : unit =
-        if cond then
-            ()
-        else
-            failwith msg
-
-    /// Create a new 'Seed' from a 32-bit integer.
-    let ofInt32 (s0 : int) : Seed =
-        // We want a non-negative number, but we can't just take the abs
-        // of s0 as -Int32.MinValue == Int32.MinValue.
-        let s = s0 &&& Int32.MaxValue
-
-        // TODO remove crashUnless
-        crashUnless (s >= 0) "s >= 0"
-
-        { Value = uint64 s
-          Gamma = goldenGamma }
-
     let private mix64 (s0 : uint64) : uint64 =
         let s = s0
         let s = (s ^^^ (s >>> 33)) * 0xff51afd7ed558ccdUL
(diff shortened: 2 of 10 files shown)


tests/Hedgehog.Linq.Tests/NameTests.cs
```diff
@@ -36,8 +37,10 @@ public static IEnumerable<object[]> AllPublicMembers()
                     }
 
                     var mi = member as MethodInfo;
-                    if (mi != null &&
-                        (mi.IsSpecialName || mi.IsConstructor || mi.Name.StartsWith("get_")))
+                    if (mi != null && (
+                           mi.IsSpecialName
+                        || mi.IsConstructor
+                        || mi.Name.StartsWith("get_")))
```
adam-becker, line 43: It would be better to just extract a method for the expression on the right of `&&`.

src/Hedgehog/Outcome.fs
```diff
@@ -7,14 +7,19 @@ type Outcome<'a> =
...
 module Outcome =
 
-    let cata result failure discard success =
-        match result with
+    let private cata
+        (outcome : Outcome<'a>)
+        (failure : unit -> 'b)
+        (discard : unit -> 'b)
+        (success :   'a -> 'b)
+            : 'b =
```
adam-becker, line 15: I don't think we need to whitespace align `'a -> 'b`, and as a personal preference, I'd put `: 'b =` on the previous line.

See `takeSmallest` in `Property.fs`.

src/Hedgehog/Seed.fs
```diff
@@ -46,24 +46,6 @@ module Seed =
...
         0x9e3779b97f4a7c15UL
 
-    let private crashUnless (cond : bool) (msg : string) : unit =
-        if cond then
-            ()
-        else
-            failwith msg
-
-    /// Create a new 'Seed' from a 32-bit integer.
-    let ofInt32 (s0 : int) : Seed =
```
adam-becker, line 56: Guess we didn't need this function anymore?

src/Hedgehog/Outcome.fs
```diff
@@ -7,14 +7,19 @@ type Outcome<'a> =
...
+            failure ()
         | Discard ->
-            discard()
-        | Success(x) ->
-            success(x)
+            discard ()
+        | Success x ->
+            success x
 
     [<CompiledName("Map")>]
```
adam-becker, line 24: We might not need the `CompiledName` attributes here, since we have `Hedgehog.Linq`.

adam-becker (changes requested):
Looks good, just left a few discretionary comments.

While we're housekeeping, could we remove the snake-cased variable names in `./src/Hedgehog/{Numeric,Range,Seed}.fs`? To be clear, I mean the variable/function names like `call_2`, `scaled_x` should be `call2` and `scaledX` respectively.

tests/Hedgehog.Linq.Tests/NameTests.cs
```diff
@@ -36,8 +37,10 @@ public static IEnumerable<object[]> AllPublicMembers()
                     }
 
                     var mi = member as MethodInfo;
```
adam-becker, line 39: This could be turned into an "if pattern match"

```csharp
if (member is MethodInfo mi && ...)
```

src/Hedgehog/Seed.fs
moodmosaic: There was a [bug in the SplitMix paper](https://github.com/hedgehogqa/haskell-hedgehog/issues/191) so after I've fixed it in Haskell I applied the changes in F#. I probably left this function in there by accident (my memory just can't go so many years back :confused:).

src/Hedgehog/Outcome.fs
moodmosaic: :thinking:

tests/Hedgehog.Linq.Tests/NameTests.cs
moodmosaic: Thank you @adam-becker. If we can switch to C# 8.0 we can use recursive patterns and static local functions:

```c#
switch (member)
{
    // Ignore special members like indexers etc.
    case PropertyInfo { IsSpecialName: true }:
    case MethodInfo mi when NotOfInterest(mi):
        continue;
    default:
        // Avoid covariant conversion from MemberInfo[] to object[].
        yield return new object[] { member };
        break;
}

static bool NotOfInterest(MethodBase mi) =>
       mi.IsSpecialName
    || mi.IsConstructor
    || mi.Name.StartsWith("get_")
       // Static inline methods, by convention starting with '`'.
    || mi.Name.StartsWith("`");
```

moodmosaic: So we can specify the C# language version in the *.csproj*, e.g.:

```diff
diff --git a/tests/Hedgehog.Linq.Tests/Hedgehog.Linq.Tests.csproj b/tests/Hedgehog.Linq.Tests/Hedgehog.Linq.Tests.csproj
index e9833ba..3d8748c 100644
--- a/tests/Hedgehog.Linq.Tests/Hedgehog.Linq.Tests.csproj
+++ b/tests/Hedgehog.Linq.Tests/Hedgehog.Linq.Tests.csproj
@@ -2,6 +2,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFrameworks>netcoreapp3.0;net461</TargetFrameworks>
+    <LangVersion>8.0</LangVersion>
     <ApplicationIcon />
     <OutputType>Library</OutputType>
     <StartupObject />
```

moodmosaic: https://github.com/hedgehogqa/fsharp-hedgehog/pull/285/commits/29e5293c4c9fe2292414720c817df9ac38172758

src/Hedgehog/Outcome.fs
moodmosaic: https://github.com/hedgehogqa/fsharp-hedgehog/pull/285/commits/68d4dc2bec0b0d38ab76007200ec3f127c227bdf

moodmosaic:
Thanks for the feedback, @adam-becker. I've added 2 commits.

moodmosaic:
>the variable/function names like call_2, scaled_x should be call2 and scaledX respectively.

Yes, we can rename those. I might just do it as part of this PR.

adam-becker (approved):
Looks good to me!

moodmosaic:
> > the variable/function names like call_2, scaled_x should be call2 and scaledX respectively.
> 
> Yes, we can rename those. I might just do it as part of this PR.

Done, except those in Numeric.fs, I'd propose to leave those as-is, to be identical with [the original code](https://github.com/gusty/FsControl/blob/f125b96d252080f722a1ff7a9efcfaf53420a339/FsControl.Core/Numeric.fs#L47).

src/Hedgehog/Seed.fs
TysonMN: Can you add to the comments in `Seed.fs` something about the bug in the original paper and link to [that haskell-hedgehog issue](https://github.com/hedgehogqa/haskell-hedgehog/issues/191)?

PR #291, opened 2021-01-30
Slightly improved docs

dharmaturtle:
I have no idea how the new fsdocs/gh-pages work, so here are some markdown edits. Perhaps we should wait until a release is cut before merging them though?

I'd like to make more drastic changes, but I'll wait until the fsdocs/gh-pages start to gain structure.

Also, there are many references to the "gen expression" and "property expression". Is there a reason why we aren't calling them "computation expressions"?

Also also, should we [disable the wiki](https://docs.github.com/en/github/building-a-strong-community/disabling-wikis) since we're going in the gh-pages direction?

### DIFF
--- a/doc/index.md
+++ b/doc/index.md
@@ -289,7 +289,7 @@ You can find out more about integrated vs type-based shrinking in [this](http://
 
 ### Generators
 
-Hedgehog's `Gen` module exports some basic generators and plenty combinators for making new generators. Here's a generator of alphanumeric chatracters:
+Hedgehog's `Gen` module exports some basic generators and plenty combinators for making new generators. Here's a generator of alphanumeric characters:
 
 ```fs
 Gen.alphaNum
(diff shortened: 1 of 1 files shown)


adam-becker:
> I have no idea how the new fsdocs/gh-pages work, so here are some markdown edits. Perhaps we should wait until a release is cut before merging them though?

The assumption is that the docs will be generated when a new release is cut, so any documentations changes supporting the next release should be made ahead of time.

> I'd like to make more drastic changes, but I'll wait until the fsdocs/gh-pages start to gain structure.

Feel free to do this. Now that we have a nice way to represent the docs, I believe they should be greatly expanded. Perhaps the work you're envisioning could be where the structure begins :smile: 

> Also, there are many references to the "gen expression" and "property expression". Is there a reason why we aren't calling them "computation expressions"?

Since they're both computation expressions, it would make sense to call them that when referring to both simultaneously. Otherwise, I'm fine with the verbiage of `gen expression` for `gen {...}` and `property expression` for `property {...}`

> Also also, should we [disable the wiki](https://docs.github.com/en/github/building-a-strong-community/disabling-wikis) since we're going in the gh-pages direction?

Yes.

adam-becker (approved):
LGTM :+1:

moodmosaic:
🎸

moodmosaic:
Thank you 💯 

We could also add Hedgehog.Xunit to the README 👀

>

moodmosaic:
> > Also also, should we [disable the wiki](https://docs.github.com/en/github/building-a-strong-community/disabling-wikis) since we're going in the gh-pages direction?
> 
> Yes.

Done.

Issue #296, opened 2021-02-01
Each function has at least one test


adam-becker:
Now that we have a single test project, it would be nice to see at least one test verifying the behavior of each function. Particularly, `Tree` and `Seed` seem to be lacking test cases.

/cc @moodmosaic @TysonMN

moodmosaic:
FWIW, that one [parameterized test in `Seed`](https://github.com/hedgehogqa/fsharp-hedgehog/pull/161/files) originates from a [fix](https://github.com/hedgehogqa/fsharp-hedgehog/pull/160/files) I originally did for `haskell-hedgehog` [here](https://github.com/hedgehogqa/haskell-hedgehog/commit/39b15b9b4d147f6001984c4b7edab00878269da7) and it basically verifies that we won't accidentally run into a bug found in the SplitMix paper.

It'd be nice with some more test coverage *as a first step* :+1: What's also important is to have a battery of statistical tests (e.g. [dieharder](https://webhome.phy.duke.edu/~rgb/General/dieharder.php)) for measuring the actual quality of the generators, something similar to what I did [for haskell-hedgehog](https://github.com/hedgehogqa/haskell-hedgehog/pull/185/files#diff-58a102a4d2236792ee09f7d51e097c491e92682d262d0f2982af2cc85baa2237) maybe.

adam-becker:
> FWIW, that one [parameterized test in `Seed`](https://github.com/hedgehogqa/fsharp-hedgehog/pull/161/files) originates from a [fix](https://github.com/hedgehogqa/fsharp-hedgehog/pull/160/files) I originally did for `haskell-hedgehog` [here](https://github.com/hedgehogqa/haskell-hedgehog/commit/39b15b9b4d147f6001984c4b7edab00878269da7) and it basically verifies that we won't accidentally run into a bug found in the SplitMix paper.

Sure, what I was pointing out is that it's the _only_ test we have for `Seed`, which is a central component of the project.

> It'd be nice with some more test coverage _as a first step_ +1 What's also important is to have a battery of statistical tests (e.g. [dieharder](https://webhome.phy.duke.edu/~rgb/General/dieharder.php)) for measuring the actual quality of the generators, something similar to what I did [for haskell-hedgehog](https://github.com/hedgehogqa/haskell-hedgehog/pull/185/files#diff-58a102a4d2236792ee09f7d51e097c491e92682d262d0f2982af2cc85baa2237) maybe.

[...]

PR #297, opened 2021-02-02
Fix flaky tests

### LINKED ISSUE #283: Flaky tests

adam-becker:
fixes #283

This addresses an issue we've had in CI, where the tests for `PropertyExtensions.Where` hit a discard limit and fail. If we generate the data such that the predicate in `PropertyExtensions.Where` always evaluates to true, we will no longer hit the discard limit randomly.

/cc @moodmosaic @TysonMN

### DIFF
--- a/tests/Hedgehog.Linq.Tests/LinqTests.cs
+++ b/tests/Hedgehog.Linq.Tests/LinqTests.cs
@@ -81,9 +81,9 @@ namespace Hedgehog.Linq.Tests
         public void CanUseWhereWithAssertion()
         {
             var property =
-                from x in ForAll(Gen.Bool)
+                from x in ForAll(Gen.FromValue(true))
                 where x == true
-                from y in ForAll(Gen.Bool)
+                from y in ForAll(Gen.FromValue(false))
                 where y == false
                 select Assert.True(x && !y);
 
@@ -94,9 +94,9 @@ namespace Hedgehog.Linq.Tests
         public void CanUseWhereWithBool()
         {
             var property =
-                from x in ForAll(Gen.Bool)
+                from x in ForAll(Gen.FromValue(true))
                 where x == true
-                from y in ForAll(Gen.Bool)
+                from y in ForAll(Gen.FromValue(false))
                 where y == false
                 select x && !y;
 


TysonMN:
I don't like this.  It is barely testing anything.  I just suggested an alternative approach in https://github.com/hedgehogqa/fsharp-hedgehog/issues/283#issuecomment-771350930.

adam-becker:
> I don't like this. It is barely testing anything. I just suggested an alternative approach in [#283 (comment)](https://github.com/hedgehogqa/fsharp-hedgehog/issues/283#issuecomment-771350930).

I think there might be some confusion on what these tests did before this PR. These tests were _never_ meant to verify any runtime behavior, they simply exist to make sure that the LINQ query syntax duck-typing works with our extension methods. The tests are aptly named `CanUseWhereWithAssertion` and `CanUseWhereWithBool`. I could remove the `property.Check` portion of these tests and get the same effect.

A little further down in the same file, you'll see tests for `Gen` and `Range` written in a similar fashion.

I believe if we want to verify that `Where(...)` actually filters properly, that should be a different test. Perhaps in the spirit of #296, and preferably one that doesn't randomly fail.

TysonMN:
Oh!  Yes, I had missed that.

In that case, this is a great improvement.

TysonMN (approved):

tests/Hedgehog.Linq.Tests/LinqTests.cs
```diff
@@ -94,9 +94,9 @@ public void CanUseWhereWithAssertion()
         public void CanUseWhereWithBool()
         {
             var property =
-                from x in ForAll(Gen.Bool)
+                from x in ForAll(Gen.FromValue(true))
                 where x == true
-                from y in ForAll(Gen.Bool)
+                from y in ForAll(Gen.FromValue(false))
                 where y == false
                 select x && !y;
```
moodmosaic, line 102: Perhaps we can take this a step further and just remove the call to property `check` so that those don't actually run (that is, we only care that the syntax works). It doesn't have to be done now, it's just a thought.

moodmosaic:
>I could remove the `property.Check` portion of these tests and get the same effect.

Yep, exactly.

PR #302, opened 2021-02-03
Housekeeping

moodmosaic:
Some housekeeping I did while going through the new Fable test-suite. Feel free to discuss and/or reject any or or all of those:

* remove redundant space
* remove unused `open` directives
* `testCaseNoFable` renamed to `fableIgnore`
* where possible maximum line width at 80 characters
* where possible use `expected`/`actual` test terminology

/cc @adam-becker @TysonMN

### DIFF
--- a/tests/Hedgehog.Tests/GenTests.fs
+++ b/tests/Hedgehog.Tests/GenTests.fs
@@ -1,68 +1,91 @@
 module Hedgehog.Tests.GenTests
 
+open System
 open Hedgehog
 open Hedgehog.Gen.Operators
 open TestDsl
 
 let genTests = testList "Gen tests" [
-    yield! testCases "dateTime creates System.DateTime instances" [ 8; 16; 32; 64; 128; 256; 512 ] <| fun count->
-        let actual = Gen.dateTime (Range.constant System.DateTime.MinValue System.DateTime.MaxValue) |> Gen.sample 0 count
+    yield! testCases "dateTime creates DateTime instances"
+        [ 8; 16; 32; 64; 128; 256; 512 ] <| fun count->
+
+        let actual =
+            (Range.constant
+                DateTime.MinValue
+                DateTime.MaxValue)
+            |> Gen.dateTime
+            |> Gen.sample 0 count
+
         actual
         |> List.distinct
         |> List.length
         =! actual.Length
 
     testCase "unicode doesn't return any surrogate" <| fun _ ->
-        let actual = Gen.sample 100 100000 Gen.unicode
-        [] =! List.filter System.Char.IsSurrogate actual
+        let actual =
+            Gen.sample 100 100000 Gen.unicode
+        [] =! List.filter Char.IsSurrogate actual
 
-    yield! testCases "unicode doesn't return any noncharacter" [ 65534; 65535 ] <| fun nonchar ->
-        let actual = Gen.sample 100 100000 Gen.unicode
+    yield! testCases "unicode doesn't return any noncharacter"
+        [ 65534; 65535 ] <| fun nonchar ->
+
+        let actual =
+            Gen.sample 100 100000 Gen.unicode
         [] =! List.filter (fun ch -> ch = char nonchar) actual
 
     testCase "dateTime randomly generates value between max and min ticks" <| fun _ ->
-        let seed0 = Seed.random()
+        let seed0 = Seed.random ()
         let (seed1, _) = Seed.split seed0
         let range =
             Range.constant
-                System.DateTime.MinValue.Ticks
-                System.DateTime.MaxValue.Ticks
+                DateTime.MinValue.Ticks
+                DateTime.MaxValue.Ticks
         let ticks =
             Random.integral range
             |> Random.run seed1 0
-        let expected = System.DateTime ticks
 
-        let actual = Gen.dateTime (Range.constant System.DateTime.MinValue System.DateTime.MaxValue)
+        let actual =
+            Range.constant DateTime.MinValue DateTime.MaxValue
+            |> Gen.dateTime
+            |> Gen.toRandom
+            |> Random.run seed0 0
+            |> Tree.outcome
 
-        let result = actual |> Gen.toRandom |> Random.run seed0 0 |> Tree.outcome
-        expected =! result
+        let expected =
+            DateTime ticks
+        expected =! actual
 
     testCase "dateTime shrinks to correct mid-value" <| fun _ ->
-        let result =
+        let actual =
             property {
                 let! actual =
-                  Range.constantFrom (System.DateTime (2000, 1, 1)) System.DateTime.MinValue System.DateTime.MaxValue
+                  (Range.constantFrom
+                       (DateTime (2000, 1, 1))
+                        DateTime.MinValue
+                        DateTime.MaxValue)
                   |> Gen.dateTime
-                System.DateTime.Now =! actual
+                DateTime.Now =! actual
             }
             |> Property.report
             |> Report.render
-            |> (fun x -> x.Split ([|System.Environment.NewLine|], System.StringSplitOptions.None))
+            |> (fun x -> x.Split ([|Environment.NewLine|], StringSplitOptions.None))
             |> Array.item 1
-            |> System.DateTime.Parse
-        System.DateTime (2000, 1, 1) =! result
+            |> DateTime.Parse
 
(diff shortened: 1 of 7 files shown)


tests/Hedgehog.Tests/TestDsl.fs
```diff
@@ -1,43 +1,38 @@
...
 #endif
 
-let inline (=!) actual expected = Expect.equal expected actual "Should be equal"
+let inline (=!) (actual : 'a) (expected : 'a) : unit =
+    Expect.equal expected actual "Should be equal"
 
 [<RequireQualifiedAccess>]
 module Expect =
-    let inline isTrue value = Expect.isTrue value "Should be true"
+    let inline isTrue value =
```
moodmosaic, line 37: I'm not sure off the top of my head what's the IL diff between inlining vs no-inlining those. (I think, last time I had to deal with IL was [in 2012](https://github.com/AutoFixture/SemanticComparison/commit/af632fc95297b6bf1fa30a5cc0bf42472e46b773#diff-cb5da29d5d5e182dadee4df2aa3f982f4dc48c161e001feaf9045e6fd6baa79d) :open_mouth:, or [is it similar to GHC's optimizations](https://downloads.haskell.org/~ghc/8.6.3/docs/html/users_guide/glasgow_exts.html#inline-pragma)?)

adam-becker: I'm completely in the dark there too, `inline` makes perfect sense to me when compiling, but how a compiled form retains the ability to be inlined is beyond me.

moodmosaic: Perhaps we can take those out. (Less is more(?))

tests/Hedgehog.Tests/GenTests.fs
```diff
@@ -5,23 +5,35 @@ open Hedgehog.Gen.Operators
 open TestDsl
 
 let genTests = testList "Gen tests" [
-    yield! testCases "dateTime creates System.DateTime instances" [ 8; 16; 32; 64; 128; 256; 512 ] <| fun count->
-        let actual = Gen.dateTime (Range.constant System.DateTime.MinValue System.DateTime.MaxValue) |> Gen.sample 0 count
+    yield! testCases "dateTime creates System.DateTime instances"
+        [ 8; 16; 32; 64; 128; 256; 512 ] <| fun count->
```
adam-becker, line 9: I'd like to see if we can get rid of the `<|` in these tests, only because we decided to do it in #276.

tests/Hedgehog.Tests/TestDsl.fs
adam-becker: Since this test project isn't a library, I think inline here doesn't hurt anything. Usually `inline` is used to mitigate overhead, and to do fun ad-hoc things like we have in `Numeric.fs`. We don't really need either in this case, so it's up to you @moodmosaic.

tests/Hedgehog.Tests/GenTests.fs
```diff
@@ -30,39 +42,49 @@ let genTests = testList "Gen tests" [
...
+                       (System.DateTime (2000, 1, 1))
+                        System.DateTime.MinValue
+                        System.DateTime.MaxValue)
                   |> Gen.dateTime
                 System.DateTime.Now =! actual
             }
             |> Property.report
             |> Report.render
-            |> (fun x -> x.Split ([|System.Environment.NewLine|], System.StringSplitOptions.None))
+            |> (fun x -> x.Split [| '\n' |])
```
adam-becker, line 70: Do we want to always assume `\n` here? Does this break Windows users?

moodmosaic: No, I made a mistake here. I'll revert that one. Very good catch :+1:

tests/Hedgehog.Tests/GenTests.fs
moodmosaic: Yes, I think so too. Let's do this on a separate PR.

tests/Hedgehog.Tests/TestDsl.fs
moodmosaic: Let's remove it. I just forced-pushed on the branch.

tests/Hedgehog.Tests/GenTests.fs
moodmosaic: Done.

moodmosaic:
I just force-pushed the branch from 667f37d to d694831. Thanks for the feedback @adam-becker :+1:

tests/Hedgehog.Tests/GenTests.fs
```diff
@@ -30,18 +42,26 @@ let genTests = testList "Gen tests" [
...
     testCase "dateTime shrinks to correct mid-value" <| fun _ ->
-        let result =
+        let actual =
             property {
                 let! actual =
-                  Range.constantFrom (System.DateTime (2000, 1, 1)) System.DateTime.MinValue System.DateTime.MaxValue
+                  (Range.constantFrom
+                       (System.DateTime (2000, 1, 1))
+                        System.DateTime.MinValue
+                        System.DateTime.MaxValue)
```
adam-becker, line 64: I think we can open `System` and not fully qualify `System.DateTime`. If `DateTime` is going to collide, we can use a type alias.

tests/Hedgehog.Tests/GenTests.fs
```diff
@@ -50,19 +70,21 @@ let genTests = testList "Gen tests" [
             |> (fun x -> x.Split ([|System.Environment.NewLine|], System.StringSplitOptions.None))
             |> Array.item 1
             |> System.DateTime.Parse
-        System.DateTime (2000, 1, 1) =! result
 
-    testCaseNoFable "int64 can create exponentially bounded integer" <| fun _ ->
+        System.DateTime (2000, 1, 1) =! actual
```
adam-becker, line 74: This value is repeated a bit, can we make a binding for it?

tests/Hedgehog.Tests/SeedTests.fs
```diff
@@ -25,7 +24,5 @@ let seedTests = testList "Seed tests" [
           (0x76442b62dddf926cUL,  8100917074368564322UL, 12297829382473034421UL) ] <| fun (x, value, gamma) ->
 
         { Value = value
-          Gamma = gamma }
-        =! Seed.from x
-
+          Gamma = gamma } =! Seed.from x
```
adam-becker, line 27:
```fsharp
let actual =
  { Value = value
    Gamma = gamma }

actual =! Seed.from x
```

tests/Hedgehog.Tests/GenTests.fs
```diff
@@ -30,18 +42,26 @@ let genTests = testList "Gen tests" [
         let ticks =
             Random.integral range
             |> Random.run seed1 0
-        let expected = System.DateTime ticks
 
-        let actual = Gen.dateTime (Range.constant System.DateTime.MinValue System.DateTime.MaxValue)
+        let actual =
+            (Range.constant System.DateTime.MinValue System.DateTime.MaxValue)
```
adam-becker, line 47: The parens aren't needed here.

adam-becker (approved):
LGTM, left a few more comments, but none of them are blocking.

tests/Hedgehog.Tests/GenTests.fs
moodmosaic: :+1:

tests/Hedgehog.Tests/SeedTests.fs
moodmosaic: :+1: (Space [isn't needed](https://blog.ploeh.dk/2013/06/24/a-heuristic-for-formatting-code-according-to-the-aaa-pattern/) there.)

tests/Hedgehog.Tests/GenTests.fs
moodmosaic: It's just in 2 places so I left it as-is.

tests/Hedgehog.Tests/GenTests.fs
moodmosaic: Done. I didn't see any issues.

moodmosaic:
Just forced-pushed. Thanks for the review. :+1:

adam-becker:
I'll let @TysonMN weigh in before merging.

TysonMN (approved):
LGTM :)

PR #307, opened 2021-02-07
made Seq module internal and moved to own file

### LINKED ISSUE #306: Seq module scope
I would prefer if the `Seq` module had internal scope.  Currently it has public scope.

https://github.com/hedgehogqa/fsharp-hedgehog/blob/f346f77b40231978e6f7904b67f96b559780d548/src/Hedgehog/Shrink.fs#L3-L17

TysonMN:
Fixes #306


moodmosaic (approved):
LGTM /cc @adam-becker

TysonMN:
...but not to the compiler 😩

moodmosaic:
Remove the inline keyword, maybe?

TysonMN:
I think understand now.

`Shrink.towards` is marked as inlined, so when that code is inlined in user code, then either `Seq.consNub` needs to be inlined as well or it needs to be accessible, which in this case is `public` scope.

adam-becker:
LGTM 👍🏼

PR #309, opened 2021-02-07
Remove PropertyConfig.coalesce

adam-becker:
Instead of making an overload that accepts a null `PropertyConfig` then converting that to a valid one, this PR creates separate overloads so this can be handled statically with overload resolution.

/cc @moodmosaic @TysonMN


moodmosaic (approved):
Much cleaner! In the spirit of https://www.simplehaskell.org/ but for F# :+1:

PR #310, opened 2021-02-07
Property isn't a true functor.

adam-becker:
Experimental, this is a rather large shift. Feel free to close if this is too much.

Every necessary property function up until now worked with a `Property<unit>` or a `Property<bool>`, this PR accepts this fact, and changes the type definition from `Property<'a>` to `Property`, as a consequence of this the `Outcome<'a>` type was also changed to `Outcome`.

This required little to no changes in the other files, which sort of proves the point of this PR. Logically, `Property` should be constructed from a `Gen<bool>`, where the `bool` indicates success.

Also, all overloads dealing with `Property<bool>` were removed, so we no longer have to implement each of those functions twice.

/cc @moodmosaic @TysonMN


TysonMN:
This idea seems reasonable.  I didn't check all the details though.

> Property isn't a true functor

I don't know what you mean by "true functor".  My guess is that you mean that `Property` is not a (polymorphic) functor in `'a`.  However, a type need not have a type parameter to be a functor.  [Monomorphic functors also exist](https://blog.ploeh.dk/2020/10/19/monomorphic-functors/).

moodmosaic:
>where the bool indicates success

Do we care about properties throwing exceptions?

adam-becker:
> I don't know what you mean by "true functor".

I just mean that although we treat it like one, at the end of the day we only care about `Property<bool>` or `Property<unit>`. The rest of the polymorphic stuff can be handled in closures or it can be erased by the time a property is constructed.

TysonMN:
My current intuition is that our property should remain polymorphic, but I might my wing about this.  (I still have not looked at the did for this PR, but I will eventually.)

adam-becker:
Closing this for now, we can re-evaluate using the new contributing process.

PR #312, opened 2021-02-07
Add property args structure.

adam-becker:
This PR adds a structure to keep all the run args in a single place. This will allow us to more easily add support for #25.

/cc @moodmosaic @TysonMN


moodmosaic:
I think this looks good, however, perhaps we can merge once we're done with #296, unless @TysonMN has a better idea.

moodmosaic (commented):
I think this looks good, however, perhaps we can merge once we're done with #296, unless @TysonMN has a better idea.

adam-becker:
> I think this looks good, however, perhaps we can merge once we're done with #296, unless @TysonMN has a better idea.

That works for me 👍🏼

TysonMN:
Also looks good to me.  I think we should merge now instead of waiting for #296.

Issue #286, opened 2021-01-23
Follow the F# style guide


adam-becker:
For reference: https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/

/cc @moodmosaic @TysonMN

moodmosaic:
This can be a prerequisite for #182. IIRC, @cmeeren follows a different style there.

cmeeren:
I don't care too much about the style used in Hedgehog.Experimental. I'd be surprised if it's at all consistent there; there's probably one style from stuff originally from someone else, one style for old stuff I made, and one style for new stuff I made. I haven't spent much effort trying to be consistent, since it's such a small project. Feel free to reformat however you like,

I suggest using Fantomas and configuring it using an `.editorconfig`. Fantomas isn't bug-free, but its aim is to have defaults that match the F# style guide. It's very nearly at a point where I can begin using it for my own projects at work.

adam-becker:
@cmeeren We had a brief discussion about using Fantomas in #271, I think we came to an agreement that it would be easier to just apply it once as a single PR, then exercise discipline in subsequent PRs. If you know how to better configure Fantomas and could get it working that would be great though.

cmeeren:
Configuring Fantomas is simple. Just place an `.editorconfig` in the repo root, have it contain [everything listed here](https://github.com/fsprojects/fantomas/blob/master/docs/Documentation.md#configuration), and tweak to your liking (see the rest of the document linked above for a description of each setting). If you use VS, download the "F# Formatting" extension which enables formatting in F# files using Fantomas with the settings from `.editorconfig` (or as configured in the plugin settings, but that's of course not portable). Any setting left at the default can be removed, but AFAIK it's safer to leave it, since otherwise someone may have a device-specific configuration of Fantomas (e.g. the F# Formatting extension config) that does not match your chosen (default) settings.

TysonMN:
> If you use VS, download the "F# Formatting" extension which enables formatting in F# files using Fantomas with the settings from `.editorconfig` (or as configured in the plugin settings, but that's of course not portable). Any setting left at the default can be removed, but AFAIK it's safer to leave it, since otherwise someone may have a device-specific configuration of Fantomas (e.g. the F# Formatting extension config) that does not match your chosen (default) settings.

Including an explicit version of these settings in our own `.editorconfig` file is better both because this ensures consistency among developers and because it is more readable.

cmeeren:
[...]

PR #315, opened 2021-02-11
Rename the parameters of Range.exponentialFrom (and friends)

### LINKED ISSUE #313: Rename the parameters of Range.exponentialFrom (and friends)
https://github.com/hedgehogqa/fsharp-hedgehog/blob/368469461f3862256eddbd6a380b5d374403da00/src/Hedgehog/Range.fs#L172-L174

`z`, `x`, and `y` are pretty opaque. How do we feel about renaming these to `origin`, `lowerBound`, and `upperBound` respectively?

I intend the rename to also affect `exponentialBounded`, `linear*`, and `constant*`.

[Haskell uses `z`, `x`, and `y`](https://hackage.haskell.org/package/hedgehog-1.0.4/docs/src/Hedgehog.Internal.Range.html#exponentialFrom), but it looks like they have docs on each specific parameter... which F# [can't do yet](https://github.com/dotnet/fsharp/issues/4154) 🙄

Thoughts?

dharmaturtle:
Closes #313

### DIFF
--- a/src/Hedgehog/Range.fs
+++ b/src/Hedgehog/Range.fs
@@ -96,12 +96,16 @@ module Range =
     [<AutoOpen>]
     [<System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)>]
     module Internal =
-        // The functions in this module where initially marked as internal
+        // The functions in this module where initially marked as `internal`
         // but then the F# compiler complained with the following message:
         //
         // The value 'linearFrom' was marked inline but its implementation
         // makes use of an internal or private function which is not
         // sufficiently accessible.
+        //
+        // But `inline` is required to work with the statically resolved type
+        // parameters on `minValue ()` and `maxValue ()`, so these functions
+        // are hidden from the public API with `EditorBrowsableState.Never`.
 
         /// Truncate `value` so it stays within bounds.
         let clamp (bound1 : 'a) (bound2 : 'a) (value : 'a) : 'a =
--- a/src/Hedgehog/Range.fs
+++ b/src/Hedgehog/Range.fs
@@ -102,12 +102,12 @@ module Range =
         // makes use of an internal or private function which is not
         // sufficiently accessible.
 
-        /// Truncate a value so it stays within some range.
-        let clamp (x : 'a) (y : 'a) (bound : 'a) : 'a =
-            if x > y then
-                min x (max y bound)
+        /// Truncate `value` so it stays within bounds.
+        let clamp (bound1 : 'a) (bound2 : 'a) (value : 'a) : 'a =
+            if bound1 > bound2 then
+                min bound1 (max bound2 value)
             else
-                min y (max x bound)
+                min bound2 (max bound1 value)
 
         /// Scale an integral linearly with the size parameter.
         let inline scaleLinear (sz : Size) (origin : 'a) (bound : 'a) : 'a =
--- a/src/Hedgehog/Range.fs
+++ b/src/Hedgehog/Range.fs
@@ -65,12 +65,12 @@ module Range =
 
     /// Construct a range which is unaffected by the size parameter with a
     /// origin point which may differ from the bounds.
-    let constantFrom (z : 'a) (x : 'a) (y : 'a) : Range<'a> =
-        Range (z, fun _ -> x, y)
+    let constantFrom (origin : 'a) (lowerBound : 'a) (upperBound : 'a) : Range<'a> =
+        Range (origin, fun _ -> lowerBound, upperBound)
 
     /// Construct a range which is unaffected by the size parameter.
-    let constant (x : 'a) (y : 'a) : Range<'a> =
-        constantFrom x x y
+    let constant (lowerBound : 'a) (upperBound : 'a) : Range<'a> =
+        constantFrom lowerBound lowerBound upperBound
 
     /// Construct a range which is unaffected by the size parameter using the
     /// full range of a data type.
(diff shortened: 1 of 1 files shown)


src/Hedgehog/Range.fs
```diff
@@ -109,52 +109,52 @@ module Range =
...
             let diff =
-                ((n - z) * bigint sz) / (bigint 99)
+                ((n - o) * bigint sz) / (bigint 99)
 
-            fromBigInt (z + diff)
+            fromBigInt (o + diff)
 
         /// Scale an integral exponentially with the size parameter.
-        let inline scaleExponential (lo : 'a) (hi : 'a) (sz0 : Size) (z0 : 'a) (n0 : 'a) : 'a =
+        let inline scaleExponential (lo : 'a) (hi : 'a) (sz0 : Size) (o0 : 'a) (n0 : 'a) : 'a =
```
adam-becker, line 128: I think these should be consistently named with everything else (`lowerBound`, `upperBound`, etc.)

Also, I never liked that the parameters had numerals in their names (`sz0`, `z0`, etc.) If we must keep this naming, then the parameters should be named simply, and names with numerals should only be internally visible.

src/Hedgehog/Range.fs
```diff
@@ -36,8 +36,8 @@ module Range =
     ///
     /// When using a 'Range' to generate numbers, the shrinking function will
     /// shrink towards the origin.
-    let origin (Range (z, _) : Range<'a>) : 'a =
-        z
+    let origin (Range (o, _) : Range<'a>) : 'a =
+        o
```
adam-becker, line 40: `o` should be `origin`, for readability's sake.

src/Hedgehog/Range.fs
```diff
@@ -20,8 +20,8 @@ module Range =
     let private bimap (f : 'a -> 'b) (g : 'c -> 'd) (a : 'a, b : 'c) : 'b * 'd =
         f a, g b
 
-    let map (f : 'a -> 'b) (Range (z, g) : Range<'a>) : Range<'b> =
-        Range (f z, fun sz ->
+    let map (f : 'a -> 'b) (Range (o, g) : Range<'a>) : Range<'b> =
+        Range (f o, fun sz ->
```
adam-becker, line 24: This function could be moved down a little, and the `origin` function could be used to get the origin from the range. Less to change if the type definition changes.

adam-becker (approved):
LGTM, left some notes that should be resolved before we merge.

src/Hedgehog/Range.fs
```diff
@@ -102,59 +103,44 @@ module Range =
         // sufficiently accessible.
 
         /// Truncate a value so it stays within some range.
-        let clamp (x : 'a) (y : 'a) (n : 'a) : 'a =
+        let clamp (x : 'a) (y : 'a) (bound : 'a) : 'a =
```
adam-becker, line 106: should `x` and `y` be `minimum` and `maximum`?

dharmaturtle: I don't think so; see the new commit.

src/Hedgehog/Range.fs
```diff
@@ -93,6 +94,7 @@ module Range =
     //
 
     [<AutoOpen>]
+    [<System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)>]
```
dharmaturtle, line 97: ref: https://stackoverflow.com/questions/27253613/why-editorbrowsablestate-never-exists

moodmosaic: It'd be nice to be able to discover those via IntelliSense internally, when working in this codebase. Is `Never` going to prevent that also?

src/Hedgehog/Range.fs
```diff
@@ -101,60 +103,45 @@ module Range =
...
-            let diff =
-                ((n - z) * bigint sz) / (bigint 99)
-
-            fromBigInt (z + diff)
+        let inline scaleLinear (sz : Size) (origin : 'a) (bound : 'a) : 'a =
+            let sz = max 0 (min 99 sz)
+            let origin = toBigInt origin
+            let bound = toBigInt bound
+            let diff = ((bound - origin) * bigint sz) / (bigint 99)
+            fromBigInt (origin + diff)
```
moodmosaic, line 119: I'd rather keep the formatting as-is until we move on with #286.

src/Hedgehog/Range.fs
```diff
@@ -101,60 +103,45 @@ module Range =
...
-            let n =
-                toBigInt n0
-
-            let diff =
-                 (((float (abs (n - z) + 1I)) ** (float sz / 99.0)) - 1.0) * float (sign (n - z))
+        let inline scaleExponential (lo : 'a) (hi : 'a) (sz : Size) (origin : 'a) (bound : 'a) : 'a =
+            let sz = clamp 0 99 sz
+            let origin = toBigInt origin
+            let bound = toBigInt bound
+            let diff = (((float (abs (bound - origin) + 1I)) ** (float sz / 99.0)) - 1.0) * float (sign (bound - origin))
```
moodmosaic, line 126: I'd rather keep the formatting as-is until we move on with #286.

src/Hedgehog/Range.fs
```diff
@@ -171,21 +158,17 @@ module Range =
...
             let scale =
                 // https://github.com/hedgehogqa/fsharp-hedgehog/issues/185
-                scaleExponential x y sz z
-            let xSized =
-                scale x
-            let ySized =
-                scale y
-            xSized, ySized)
+                scaleExponential lowerBound upperBound sz origin
+            scale lowerBound, scale upperBound)
```
moodmosaic, line 166: 👍

moodmosaic (changes requested):
Looking good. Left a few comments wrt styling.

src/Hedgehog/Range.fs
dharmaturtle: Yes, it prevents you from seeing it even when you're in the same fsproj.

FWIW we only have one occurrence of `Range.Internal` since its `[<AutoOpen>]`

moodmosaic: Is it practical, though? Perhaps we can just take it out of the `Range.Internal` namespace and mark it as `internal`. Can be done separately, of course.

dharmaturtle: Unfortunately, `inline` functions can't have `internal` scope if they're used in publicly scoped functions.

I'm not sure why they're `inline` to begin with, honestly.

moodmosaic: I can't remember why I did it that way, but it's a good tradeoff to have it as `internal` and get rid of the namespace and the inline part, which my gut feeling says it won't hurt the performance.

dharmaturtle: `inline` is required to work with SRTPs - I elaborated on the comment on the `Internal` module in a new commit.

dharmaturtle:
Just checking - I believe I'm done addressing the comments/reviews on this PR. Please let me know if I overlooked something.

adam-becker:
Sorry for the delay in getting this through. @moodmosaic are you okay with these changes?

moodmosaic (approved):

moodmosaic:
LGTM 👍 💯

adam-becker:
Thanks @dharmaturtle for the work you've done, it's always appreciated!

PR #321, opened 2021-02-23
Remove duplicates in frequency shrink tree

### LINKED ISSUE #319: Frequency generator should immediately shrink
F# Hedgehog might contain [this bug that was fixed in Haskell Hedgehog](https://github.com/hedgehogqa/haskell-hedgehog/pull/406) about the frequency generator testing the same values more than once.

If F# Hedgehog has this bug, then we should fix it.  Either way, we should have a test that would fail if we had this bug.

I don't know if we have that bug or have such a test.  I will work on this (if no one beats me to it).

TysonMN:
Fixes #319

We had the same bug that Haskell had.  They fixed this bug in PR https://github.com/hedgehogqa/haskell-hedgehog/pull/406.  I tried to mimic their solution here.  I think I was successful.

CC @HuwCampbell

I did one thing differently related to their `shrink` function.  I think the input `n` of the lambda to `shrink` is put into the root of the shrink tree by `shrink`.  Instead, I avoided an off-by-one error by first paring the weights and then calling `takeWhile`.  Without pairing, the resulting sequence does not include the weight corresponding to the weight `k` in the base case of `pick`.

### DIFF
--- a/src/Hedgehog/Gen.fs
+++ b/src/Hedgehog/Gen.fs
@@ -206,8 +206,26 @@ module Gen =
                 else
                     pick (n - k) ys
 
+        let f n =
+            let smallWeights =
+                xs
+                |> List.map fst
+                |> List.scan (+) 0
+                |> List.pairwise
+                |> List.takeWhile (fun (a, _) -> a < n)
+                |> List.map snd
+                |> List.toArray
+            let length = smallWeights |> Array.length
+            Shrink.createTree 0 (length - 1)
+            |> Tree.map (fun i -> smallWeights.[i])
+
         gen {
-            let! n = Range.constant 1 total |> integral
+            let! n =
+                Range.constant 1 total
+                |> integral
+                |> toRandom
+                |> Random.map (Tree.outcome >> f)
+                |> ofRandom
             return! pick n xs
         }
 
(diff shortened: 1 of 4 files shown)


TysonMN:
> I tried to mimic their solution here. I think I was successful.

To be clear, I mean I think I was successful in fixing the bug.  My implementation is probably not as elegant as @HuwCampbell's.

I force pushed the branch with some new commits that add another test.  This test ensures that the entire shrink tree produced by `frequency` is balanced.  Because a test failure means recursing into the corresponding subtree and a test success means moving on to the next child, a balanced shrink tree is one in which, at each subtree/`Node`, the number of children and the depth of that subtree differ by at most `1`.

@HuwCampbell, although I was able to understand your Haskell code well enough, I don't understand it perfectly.  Does your implementation produce balanced shrink trees?

HuwCampbell:
No it didn't produce a balanced tree. We did it before our other research on balanced trees. I think it should be possible though.

TysonMN:
Maybe that can be your best improvement then.  I am still curious, after that change, if the resulting tree is balanced.

adam-becker:
@TysonMN do you remember if this is ready to merge?

TysonMN:
I think so.  As I recall, the tests here are very strong.

adam-becker:
Merged. 👍🏼

PR #318, opened 2021-02-22
Implement Property.select via bind to avoid bug

TysonMN:
Fixes #317

I added tests corresponding to those given in #317 except that they pass when the behavior is correct and fail otherwise.  Then I fixed the issue by implementing `Property.select` via `Property.bind` and especially `Property.ofThrowing`, which maps a thrown exception to `Outcome<_>.Failure`.

PR #324, opened 2021-03-02
Special processing for printing ResizeArray<_> in Property.forAll #323

### LINKED ISSUE #323: A way to render a list completely in the report?
Given a property that takes a list generator, such as this
```cs
[Test]
public void Hedgehog_List_IsPrintedIncomplete()
{
    var property =
        from xs in Property.ForAll(Gen.Int32(Range.Constant(1, 255)).List(Range.Constant(1, 100)))
        select xs.All(x => x < 255);
    property.Check(PropertyConfig.Default.WithShrinks(0));
}
```
When the property fails, the output only shows the first several elements of the list
```
System.Exception : *** Failed! Falsifiable (after 5 tests):
seq [188; 144; 215; 37; ...]
```
It can be beneficial to be able to see all elements of the list in the test output. This can be achieved by wrapping the input in a class and overriding its ToString(), but it takes significantly more code, so a simpler way would be preferable.
```cs
class Wrapper
{
    public List<int> Xs;

    public override string ToString()
    {
        return $"[{string.Join(", ", Xs)}]";
    }
}

[Test]
public void Hedgehog_ListWithinClassWithACustomToString_IsPrintedComplete()
{
    var gen =
        from xs in Gen.Int32(Range.Constant(1, 255)).List(Range.Constant(1, 100))
        select new Wrapper {Xs = xs};
    var property =
        from wrapper in Property.ForAll(gen)
        select wrapper.Xs.All(x => x < 255);
    property.Check(PropertyConfig.Default.WithShrinks(0));
}
```

altxt:
Tried to fix #323

### DIFF
--- a/src/Hedgehog/Property.fs
+++ b/src/Hedgehog/Property.fs
@@ -81,24 +81,23 @@ module Property =
     let bind (k : 'a -> Property<'b>) (m : Property<'a>) : Property<'b> =
         bindGen (toGen << k) (toGen m) |> ofGen
 
+
+    // sprintf "%A" is not prepared for printing ResizeArray<_> (C# List<T>) so we prepare the value instead
+    let private prepareForPrinting (value: obj) : obj =
 #if FABLE_COMPILER
-    let private printValue (value) : string =
-        sprintf "%A" value
+        value
 #else
-    let private (|IsResizeArray|_|) (candidate : obj) : obj list option =
-        let t = candidate.GetType()
+        let t = value.GetType()
         // have to use TypeInfo due to targeting netstandard 1.6
         let t = System.Reflection.IntrospectionExtensions.GetTypeInfo(t)
         let isList = t.IsGenericType && t.GetGenericTypeDefinition() = typedefof<ResizeArray<_>>
         if isList
-        then Some (candidate :?> System.Collections.IEnumerable |> Seq.cast<obj> |> List.ofSeq)
-        else None
+        then value :?> System.Collections.IEnumerable |> Seq.cast<obj> |> List.ofSeq :> obj
+        else value
+#endif
 
     let private printValue (value) : string =
-        match value with
-        | IsResizeArray list -> sprintf "%A" list
-        | _ -> sprintf "%A" value
-#endif
+        value |> prepareForPrinting |> sprintf "%A"
 
     let forAll (k : 'a -> Property<'b>) (gen : Gen<'a>) : Property<'b> =
         let handle (e : exn) =
--- a/src/Hedgehog/Property.fs
+++ b/src/Hedgehog/Property.fs
@@ -81,22 +81,21 @@ module Property =
     let bind (k : 'a -> Property<'b>) (m : Property<'a>) : Property<'b> =
         bindGen (toGen << k) (toGen m) |> ofGen
 
-
-    // sprintf "%A" is not prepared for printing ResizeArray<_> (C# List<T>) so we prepare the value instead
-    let private prepareForPrinting (value: obj) : obj =
-#if FABLE_COMPILER
-        value
-#else
-        let t = value.GetType()
-        // have to use TypeInfo due to targeting netstandard 1.6
-        let t = System.Reflection.IntrospectionExtensions.GetTypeInfo(t)
-        let isList = t.IsGenericType && t.GetGenericTypeDefinition() = typedefof<ResizeArray<_>>
-        if isList
-        then value :?> System.Collections.IEnumerable |> Seq.cast<obj> |> List.ofSeq :> obj
-        else value
-#endif
-
     let private printValue (value) : string =
+        // sprintf "%A" is not prepared for printing ResizeArray<_> (C# List<T>) so we prepare the value instead
+        let prepareForPrinting (value: obj) : obj =
+        #if FABLE_COMPILER
+            value
+        #else
+            let t = value.GetType()
+            // have to use TypeInfo due to targeting netstandard 1.6
+            let t = System.Reflection.IntrospectionExtensions.GetTypeInfo(t)
+            let isList = t.IsGenericType && t.GetGenericTypeDefinition() = typedefof<ResizeArray<_>>
+            if isList
+            then value :?> System.Collections.IEnumerable |> Seq.cast<obj> |> List.ofSeq :> obj
+            else value
+        #endif
+
         value |> prepareForPrinting |> sprintf "%A"
 
     let forAll (k : 'a -> Property<'b>) (gen : Gen<'a>) : Property<'b> =
--- a/src/Hedgehog/Property.fs
+++ b/src/Hedgehog/Property.fs
@@ -81,25 +81,23 @@ module Property =
     let bind (k : 'a -> Property<'b>) (m : Property<'a>) : Property<'b> =
         bindGen (toGen << k) (toGen m) |> ofGen
 
-#if !FABLE_COMPILER
-    let private (|IsGenericList|_|) (candidate : obj) : seq<obj> option =
+#if FABLE_COMPILER
+    let private printValue (value) : string =
+        sprintf "%A" value
+#else
+    let private (|IsResizeArray|_|) (candidate : obj) : obj list option =
         let t = candidate.GetType()
         // have to use TypeInfo due to targeting netstandard 1.6
         let t = System.Reflection.IntrospectionExtensions.GetTypeInfo(t)
         let isList = t.IsGenericType && t.GetGenericTypeDefinition() = typedefof<ResizeArray<_>>
         if isList
-        then Some (candidate :?> System.Collections.IEnumerable |> Seq.cast<obj>)
+        then Some (candidate :?> System.Collections.IEnumerable |> Seq.cast<obj> |> List.ofSeq)
         else None
 
     let private printValue (value) : string =
         match value with
-        | IsGenericList list ->
-            let printedElements = Seq.map (fun element -> sprintf "%A" element) list
-            "[" + (String.concat ", " printedElements) + "]"
+        | IsResizeArray list -> sprintf "%A" list
         | _ -> sprintf "%A" value
-#else
-    let private printValue (value) : string =
-        sprintf "%A" value
 #endif
 
     let forAll (k : 'a -> Property<'b>) (gen : Gen<'a>) : Property<'b> =
--- a/src/Hedgehog/Property.fs
+++ b/src/Hedgehog/Property.fs
@@ -81,6 +81,7 @@ module Property =
     let bind (k : 'a -> Property<'b>) (m : Property<'a>) : Property<'b> =
         bindGen (toGen << k) (toGen m) |> ofGen
 
+#if !FABLE_COMPILER
     let private (|IsGenericList|_|) (candidate : obj) : seq<obj> option =
         let t = candidate.GetType()
         // have to use TypeInfo due to targeting netstandard 1.6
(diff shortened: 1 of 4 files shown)


src/Hedgehog/Property.fs
```diff
@@ -81,11 +81,32 @@ module Property =
     let bind (k : 'a -> Property<'b>) (m : Property<'a>) : Property<'b> =
         bindGen (toGen << k) (toGen m) |> ofGen
 
+#if !FABLE_COMPILER
```
TysonMN, line 84: Can you use positive logic by writing the compiler directives this way?
```
#if FABLE_COMPILER
...
#else
...
#endif
```

tests/Hedgehog.Tests/PropertyTests.fs
```diff
@@ -0,0 +1,21 @@
+﻿module Hedgehog.Tests.PropertyTests
+
+open Hedgehog
+open Expecto
+open TestDsl
+
+let propertyTests = testList "Property tests" [
+    fableIgnore "generated C# list of five elements is not abbreviated in the failure report" <| fun _ ->
+        Expect.throwsC
```
TysonMN, line 9: Instead of asserting that some code throws an exception, I think it would be better to test if the rendered property doesn't contain the substring `"..."`.  See [here for a C# example](https://github.com/hedgehogqa/fsharp-hedgehog/pull/318/files#diff-c5321d72bf30f64c2b6fb902119ac3bdf4b9c9720755247f774ff7c13c0b4208R34-R37) of how to do that.

tests/Hedgehog.Tests/PropertyTests.fs
```diff
@@ -0,0 +1,21 @@
...
+            (fun () ->
+                property {
+                    let! xs = Range.singleton 0 |> Gen.int |> Gen.list (Range.singleton 5) |> Gen.map ResizeArray
+                    return Seq.forall ((>) 0) xs
```
TysonMN, line 13: I think this test can be simplified by replacing `int` with `unit`.

src/Hedgehog/Property.fs
```diff
@@ -81,11 +81,32 @@ module Property =
...
+    let private (|IsGenericList|_|) (candidate : obj) : seq<obj> option =
+        let t = candidate.GetType()
+        // have to use TypeInfo due to targeting netstandard 1.6
+        let t = System.Reflection.IntrospectionExtensions.GetTypeInfo(t)
+        let isList = t.IsGenericType && t.GetGenericTypeDefinition() = typedefof<ResizeArray<_>>
+        if isList
+        then Some (candidate :?> System.Collections.IEnumerable |> Seq.cast<obj>)
```
TysonMN, line 91: Is it possible to directly cast to `ResizeArray<obj>`?

src/Hedgehog/Property.fs
```diff
@@ -81,11 +81,32 @@ module Property =
     let bind (k : 'a -> Property<'b>) (m : Property<'a>) : Property<'b> =
         bindGen (toGen << k) (toGen m) |> ofGen
 
+#if !FABLE_COMPILER
+    let private (|IsGenericList|_|) (candidate : obj) : seq<obj> option =
```
TysonMN, line 85: What about renaming this to `IsResizeArray`?

src/Hedgehog/Property.fs
```diff
@@ -81,11 +81,32 @@ module Property =
...
+        let isList = t.IsGenericType && t.GetGenericTypeDefinition() = typedefof<ResizeArray<_>>
+        if isList
+        then Some (candidate :?> System.Collections.IEnumerable |> Seq.cast<obj>)
+        else None
+
+    let private printValue (value) : string =
+        match value with
+        | IsGenericList list ->
+            let printedElements = Seq.map (fun element -> sprintf "%A" element) list
+            "[" + (String.concat ", " printedElements) + "]"
```
TysonMN, line 98: How about mapping the value `list` to be of type `obj list`?  Then no additional "to string" code is needed.

Then I expect this function could be renamed to something like `improveRuntimeType` and the implementation below would be
```fsharp
x |> improveRuntimeType |> sprintf "%A"
```

TysonMN (changes requested):
This is also the first solution that came to my mind.

It would be interesting to see if there are any bug reports or feature requests against F# for this behavior.

src/Hedgehog/Property.fs
altxt: `candidate :?> ResizeArray<obj>` throws at runtime
```
Unable to cast object of type 'System.Collections.Generic.List`1[System.Int32]' to type 'System.Collections.Generic.List`1[System.Object]
```
Possibly due to F# issues with variance https://github.com/fsharp/fslang-suggestions/issues/162

src/Hedgehog/Property.fs
TysonMN: Oh, yes.  That makes sense.  Otherwise it would suggest that adding any instance of type `obj` were possible, which is not the case.

src/Hedgehog/Property.fs
altxt: Done

tests/Hedgehog.Tests/PropertyTests.fs
altxt: Done

tests/Hedgehog.Tests/PropertyTests.fs
altxt: Replaced condition with `return false`

src/Hedgehog/Property.fs
altxt: Done

src/Hedgehog/Property.fs
altxt: I tried to convert the active pattern to a regular function but it did not typecheck (obj vs obj list)
Otherwise - done

TysonMN: I like the use of an active pattern here.  I am not suggesting removing it.

src/Hedgehog/Property.fs
TysonMN: (Weird.  I thought I already posted this.  Oh well.  Here it is again.)

Oh, yes.  That makes sense.  Otherwise, it would be possible to add any instance of type `obj`, which is not possible.

src/Hedgehog/Property.fs
```diff
@@ -81,11 +81,30 @@ module Property =
...
+        let t = System.Reflection.IntrospectionExtensions.GetTypeInfo(t)
+        let isList = t.IsGenericType && t.GetGenericTypeDefinition() = typedefof<ResizeArray<_>>
+        if isList
+        then Some (candidate :?> System.Collections.IEnumerable |> Seq.cast<obj> |> List.ofSeq)
+        else None
+
+    let private printValue (value) : string =
+        match value with
+        | IsResizeArray list -> sprintf "%A" list
+        | _ -> sprintf "%A" value
```
TysonMN, line 100: Now that I remembered that we can't cast to `ResizeArray<obj>`, I don't think the active pattern helps.  I think it is misleading for the pattern to be named `IsResizeArray` but then not return a `ResizeArray<_>`.

What do you think of this code?  I think it is simpler.  In addition to this suggestion, you would also need to make a change below, where I recommend `x |> improveRuntimeType |> sprintf "%A"`.

```suggestion
    let private improveRuntimeType (value: obj) : obj =
#if FABLE_COMPILER
        value
#else
        let t = value.GetType()
        // have to use TypeInfo due to targeting netstandard 1.6
        let t = System.Reflection.IntrospectionExtensions.GetTypeInfo(t)
        let isList = t.IsGenericType && t.GetGenericTypeDefinition() = typedefof<ResizeArray<_>>
        if isList
        then value :?> System.Collections.IEnumerable |> Seq.cast<obj> |> List.ofSeq
        else value
```

altxt: Done, but I used a different name for the function that feels more descriptive

src/Hedgehog/Property.fs
```diff
@@ -81,11 +81,29 @@ module Property =
...
+        // have to use TypeInfo due to targeting netstandard 1.6
+        let t = System.Reflection.IntrospectionExtensions.GetTypeInfo(t)
+        let isList = t.IsGenericType && t.GetGenericTypeDefinition() = typedefof<ResizeArray<_>>
+        if isList
+        then value :?> System.Collections.IEnumerable |> Seq.cast<obj> |> List.ofSeq :> obj
+        else value
+#endif
+
+    let private printValue (value) : string =
+        value |> prepareForPrinting |> sprintf "%A"
```
TysonMN, line 100: I think this can be inlined

tests/Hedgehog.Tests/PropertyTests.fs
```diff
@@ -0,0 +1,16 @@
...
+open Expecto
+open TestDsl
+
+let propertyTests = testList "Property tests" [
+    fableIgnore "generated C# list of five elements is not abbreviated in the failure report" <| fun _ ->
+        let prop = property {
+            let! xs = Range.singleton 0 |> Gen.int |> Gen.list (Range.singleton 5) |> Gen.map ResizeArray
+            return false
+        }
+        let report = Property.renderWith (PropertyConfig.withShrinks 0<shrinks> PropertyConfig.defaultConfig) prop
```
TysonMN, line 13:
```suggestion
        let report = property {
            let! xs = Range.singleton 0 |> Gen.int |> Gen.list (Range.singleton 5) |> Gen.map ResizeArray
            return false
        }
        |> Property.renderWith (PropertyConfig.withShrinks 0<shrinks> PropertyConfig.defaultConfig)
```

src/Hedgehog/Property.fs
altxt: It can, but isn't it better to let the compiler do the inlining? This way we know the intent of the code, captured in a name, and are ready to reuse it should the need arise.

TysonMN: Ok, good point.

tests/Hedgehog.Tests/PropertyTests.fs
altxt: It looks like putting parentheses around the RHS of the let binding is required for this to compile. Do you think it looks better still?

TysonMN: How about that?

altxt: Done

TysonMN (approved):
Great job @altxt!  I think this is ready to merge.

src/Hedgehog/Property.fs
```diff
@@ -81,11 +81,28 @@ module Property =
...
+            let t = value.GetType()
+            // have to use TypeInfo due to targeting netstandard 1.6
+            let t = System.Reflection.IntrospectionExtensions.GetTypeInfo(t)
+            let isList = t.IsGenericType && t.GetGenericTypeDefinition() = typedefof<ResizeArray<_>>
+            if isList
+            then value :?> System.Collections.IEnumerable |> Seq.cast<obj> |> List.ofSeq :> obj
+            else value
+        #endif
+
+        value |> prepareForPrinting |> sprintf "%A"
```
adam-becker, line 99: I think it might be possible to extend the `Gen<'a>` type to include a format function for the type. For most values, this could be `fun value -> sprintf "%A" value`, and for `ResizeArray<'a>` it could be `fun resizeArray -> <your list logic above>`. This would eliminate the need for reflection, and keep us from having to add special cases for each type.

@TysonMN What are your thoughts on the approach I outlined above?

TysonMN: Seems reasonable.  However, I would create a new issue for that alternative design and merge this PR as is since it fixes a bug without changing the public API.

adam-becker: Done, the new issues is #326.

adam-becker:
Thanks for contributing, @altxt!

PR #328, opened 2021-04-14
Fix for counterexample crashing bug #327

### LINKED ISSUE #327: Bug: `counterexample` breaks in presence of exceptions
Version: **0.10** from NuGet

It looks like when an exception is thrown inside of a property test, it works just fine unless there is a call to `counterexample`.

Here is the test:

```fsharp
[<Fact>]
let ``Test exceptions`` () =
    property {
        let! a = Gen.string (Range.linear 2 10) Gen.alphaNum
        
//        counterexample a
        Assert.Equal(a, "")
    } |> Property.check
```

with `counterexample` commented, it works as expected showing this output:

```
System.Exception: *** Failed! Falsifiable (after 1 test and 3 shrinks):

System.Exception
*** Failed! Falsifiable (after 1 test and 3 shrinks):
"aa"
Xunit.Sdk.EqualException: Assert.Equal() Failure
          ↓ (pos 0)
Expected: aa
Actual:   
          ↑ (pos 0)
   at Xunit.Assert.Equal(String expected, String actual, Boolean ignoreCase, Boolean ignoreLineEndingDifferences, Boolean ignoreWhiteSpaceDifferences) in C:\Dev\xunit\xunit\src\xunit.assert\Asserts\StringAsserts.cs:line 244
```

However, with `counterexample` uncommented it produces the following output:

```
Xunit.Sdk.EqualException
Assert.Equal() Failure
          ↓ (pos 0)
Expected: BB
Actual:   
          ↑ (pos 0)
   at Tests.Test exceptions@50-6.Invoke(String _arg2) in /Users/alexey/src/ep/connector-classroom/EP.Connector.Classroom.Tests/Tests.fs:line 50
   at Hedgehog.Property.bind@108-2.Invoke(a x)
   at Hedgehog.Property.bindGen@98.Invoke(Tuple`2 tupledArg)
   at Hedgehog.Gen.bind@44-1.Invoke(a x)
   at Hedgehog.Gen.bindRandom@41-1.Invoke(a x)
   at Hedgehog.Tree.bind[a,b](FSharpFunc`2 k, Tree`1 _arg1)
   at Hedgehog.Gen.bindRandom@34.Invoke(Seed seed0, Int32 size)
   at Hedgehog.Property.loop@168-11(Boolean renderRecheck, PropertyConfig config, Random`1 random, Seed seed, Int32 size, Int32 tests, Int32 discards)
   at Hedgehog.Property.reportWith'(Boolean renderRecheck, Int32 size0, Seed seed, PropertyConfig config, Property`1 p)
   at Hedgehog.Property.reportWith(PropertyConfig config, Property`1 p)
   at Hedgehog.Property.check(Property`1 p)
```

So as we can see, there is no `after N tests and K shrinks` message, the value is `BB` and not `aa` which suggests that shrinking has not even happened and, as it seems, the test just crashed.

`Assert.Equal` just throws and exception internally, so it can be replaced with `failwith` with the same result:

```fsharp
let ``Test exceptions`` () =
    property {
        let! a = Gen.string (Range.linear 2 10) Gen.alphaNum
        
        counterexample a
        failwith "boom!"
    } |> Property.check
```

Which probably means that when a function under test throws an exception Hedgehog may not work properly if `counterexample` is present in the test.

### Expected behaviour

I expect that test works with `counterexample` present the same way as it works without `counterexample` except, of course, `counterexample` printing the message.

TysonMN:
Fixes #327

I think this PR fixes the bug where an exception isn't caught when checking a `Property<_>` containing `counterexample`.  I got the idea for this fix by looking at `Property.forall`, which is the only other place that `Property.counterexample` is called.  In the last commit, I removed the exception handling code from `Property.forall` since it is now happening in `Property.bind`.

How does this compare to the code in Haskell Hedgehog?

@AlexeyRaga, does the branch in this PR give the behavior that you expect?

### DIFF
--- a/src/Hedgehog/Property.fs
+++ b/src/Hedgehog/Property.fs
@@ -56,15 +56,19 @@ module Property =
             failure
 
     let counterexample (msg : unit -> string) : Property<unit> =
-        Gen.constant (Journal.singleton msg, Success ()) |> ofGen
+        (Journal.singleton msg, Success ()) |> Gen.constant |> ofGen
 
     let private mapGen
             (f : Gen<Journal * Outcome<'a>> -> Gen<Journal * Outcome<'b>>)
-            (x : Property<'a>) : Property<'b> =
-        toGen x |> f |> ofGen
+            (p : Property<'a>) : Property<'b> =
+        p |> toGen |> f |> ofGen
 
     let map (f : 'a -> 'b) (x : Property<'a>) : Property<'b> =
-        (mapGen << GenTuple.mapSnd << Outcome.map) f x
+        let g = f |> Outcome.map |> GenTuple.mapSnd |> mapGen
+        g x
+
+    let private set (a: 'a) (property : Property<'b>) : Property<'a> =
+        property |> map (fun _ -> a)
 
     let private bindGen
             (k : 'a -> Gen<Journal * Outcome<'b>>)
@@ -79,10 +83,18 @@ module Property =
                 GenTuple.mapFst (Journal.append journal) (k x))
 
     let private handle (e : exn) =
-        Gen.constant (Journal.singletonMessage (string e), Failure) |> ofGen
+        (Journal.singletonMessage (string e), Failure) |> Gen.constant
 
     let bind (k : 'a -> Property<'b>) (m : Property<'a>) : Property<'b> =
-        bindGen (fun a -> (try k a with e -> handle e) |> toGen) (toGen m) |> ofGen
+        let kTry a =
+            try
+                k a |> toGen
+            with e ->
+                handle e
+        m
+        |> toGen
+        |> bindGen kTry
+        |> ofGen
 
     let private printValue (value) : string =
         // sprintf "%A" is not prepared for printing ResizeArray<_> (C# List<T>) so we prepare the value instead
@@ -101,13 +113,11 @@ module Property =
 
         value |> prepareForPrinting |> sprintf "%A"
 
-    let private handle (e : exn) =
-        Gen.constant (Journal.singletonMessage (string e), Failure) |> ofGen
-
     let forAll (k : 'a -> Property<'b>) (gen : Gen<'a>) : Property<'b> =
         let prepend (x : 'a) =
             counterexample (fun () -> printValue x)
-            |> bind (fun _ -> try k x with e -> handle e)
+            |> set x
+            |> bind k
             |> toGen
 
         gen |> Gen.bind prepend |> ofGen
@@ -211,7 +221,7 @@ module Property =
         try
             success (f a)
         with e ->
-            handle e
+            handle e |> ofGen
 
     let reportRecheckWith (size : Size) (seed : Seed) (config : PropertyConfig) (p : Property<unit>) : Report =
         let args = {
--- a/src/Hedgehog/Property.fs
+++ b/src/Hedgehog/Property.fs
@@ -78,8 +78,11 @@ module Property =
             | Success x ->
                 GenTuple.mapFst (Journal.append journal) (k x))
 
+    let private handle (e : exn) =
+        Gen.constant (Journal.singletonMessage (string e), Failure) |> ofGen
+
     let bind (k : 'a -> Property<'b>) (m : Property<'a>) : Property<'b> =
-        bindGen (toGen << k) (toGen m) |> ofGen
+        bindGen (fun a -> (try k a with e -> handle e) |> toGen) (toGen m) |> ofGen
 
     let private printValue (value) : string =
         // sprintf "%A" is not prepared for printing ResizeArray<_> (C# List<T>) so we prepare the value instead
--- a/tests/Hedgehog.Tests/PropertyTests.fs
+++ b/tests/Hedgehog.Tests/PropertyTests.fs
@@ -14,4 +14,19 @@ let propertyTests = testList "Property tests" [
             |> Property.renderWith (PropertyConfig.withShrinks 0<shrinks> PropertyConfig.defaultConfig)
         Expect.isNotMatch report "\.\.\." "Abbreviation (...) found"
 
+    testCase "counterexample example" <| fun () ->
+        // based on examlpe from https://hedgehogqa.github.io/fsharp-hedgehog/index.html#Custom-Operations
+        let guid = System.Guid.NewGuid().ToString()
+        let tryAdd a b =
+            if a > 100 then None // Nasty bug.
+            else Some(a + b)
+        let report =
+            property {
+                let! a = Range.constantBounded () |> Gen.int
+                let! b = Range.constantBounded () |> Gen.int
+                counterexample guid
+                Some(a + b) =! tryAdd a b
+            }
+            |> Property.renderWith (PropertyConfig.withShrinks 0<shrinks> PropertyConfig.defaultConfig)
+        Expect.stringContains report guid "Missing counterexample text"
 ]
--- a/tests/Hedgehog.Tests/PropertyTests.fs
+++ b/tests/Hedgehog.Tests/PropertyTests.fs
@@ -28,5 +28,10 @@ let propertyTests = testList "Property tests" [
                 Some(a + b) =! tryAdd a b
             }
             |> Property.renderWith (PropertyConfig.withShrinks 0<shrinks> PropertyConfig.defaultConfig)
+#if FABLE_COMPILER
+// See the discussion in this PR for what needs to happen for Fable to support Expect.stringContains https://github.com/hedgehogqa/fsharp-hedgehog/pull/328
+        Expect.isTrue (report.Contains guid)
+#else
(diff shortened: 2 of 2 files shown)


TysonMN:
The build fails because
> [`Expecto_Expect_stringContains` is not defined](https://github.com/hedgehogqa/fsharp-hedgehog/runs/2345577625?check_suite_focus=true#step:5:332)

when compiled by Fable.  This is related to a test that I added.

@ThisFunctionalTom, do you have an opinion about how to fix this?  Of course the brute force solution is to have Fable ignore this test.

ThisFunctionalTom:
> The build fails because
> 
> > [`Expecto_Expect_stringContains` is not defined](https://github.com/hedgehogqa/fsharp-hedgehog/runs/2345577625?check_suite_focus=true#step:5:332)
> 
> when compiled by Fable. This is related to a test that I added.
> 
> @ThisFunctionalTom, do you have an opinion about how to fix this? Of course the brute force solution is to have Fable ignore this test.

I guess we can make a pull request in mocha to also implement `Expect.stringContains` or change the implementation to `Expect.isTrue (report.Contains...`. I think the first option is preferable.

ThisFunctionalTom:
I just made a [pull request](https://github.com/Zaid-Ajaj/Fable.Mocha/pull/30) for it. 
If you need more functions from `Expect` module we would probably need to copy the implementation from Expecto into Fable.Mocha but most of the time this should not be a problem I guess.

TysonMN:
That PR seems like the right fix. Thanks @ThisFunctionalTom.

ThisFunctionalTom:
If you are in a hurry I guess we could extend the Expect module in FSharp.Hedgehog with a `#if FABLE` guard until the PR is merged.

TysonMN:
I don't think there is any hurry. I would prefer to wait for that PR to be merged and a release to be made.

ThisFunctionalTom:
Sorry, have not got time for this. The pull request is already merged but we still have some troubles because we do not use the Except module from Mocha. I will look into it on the weekend.

ThisFunctionalTom:
Finally had some time to look into it.

The first problem is that Fable.Mocha does not have all the assertions from the Expect module as Expecto does. The `stringContains` is now added but there are a lot more missing.

If we are using only a few assertions we can copy the implementation from Expecto into TestDsl (not a really big fan of this solution) or we can make a pull request in Expecto repository to add the Expect.fs file into .fable directory of the Expecto nuget package (the same way we did it in fsharp-hedgehog) so all functions from Expect will work out of the box with fable compiler.

The second problem is that we must open the correct namespace depending on `FABLE_COMPILER`. In one case it is `Expecto` namespace with submodule `Expect` and in the other case it is `Fable.Mocha` namespace with submodule `Expect`. I tried to abstract this away with TestDsl module but it does not work because each test file must open its own namespaces. 
We have 3 options here:
- use #if FABLE_COMPILER in each file and open different `Expect` modules like this
  ```fsharp
  # if FABLE_COMPILER
  open Fable.Mocha
  #else
  open Expecto
  #endif
  ```
- alias all used assertions in TestDsl module like I did for `isTrue` and `stringContains`
- if we succeed in making Expecto more fable friendly by adding Expect.fs into .fable directory I guess it should work for most/all of the assertions from the `Expect` module.

The pull request with the fixed `PropertyTests` can be found here:
https://github.com/ThisFunctionalTom/fsharp-hedgehog/tree/bug/327_counterexample_crashes

How should we proceed? What do you think?

TysonMN:
> The first problem is that Fable.Mocha does not have all the assertions from the Expect module as Expecto does.

I would prefer if any assertions we need were added to Fable.Mocha.

 > The second problem is that we must open the correct namespace depending on `FABLE_COMPILER`. In one case it is `Expecto` namespace with submodule `Expect` and in the other case it is `Fable.Mocha` namespace with submodule `Expect`.

We don't have very many test files. I think it is fine to add the code in your comment to the top of each file.

ThisFunctionalTom:
> > The first problem is that Fable.Mocha does not have all the assertions from the Expect module as Expecto does.
> 
> I would prefer if any assertions we need were added to Fable.Mocha.
> 
> > The second problem is that we must open the correct namespace depending on `FABLE_COMPILER`. In one case it is `Expecto` namespace with submodule `Expect` and in the other case it is `Fable.Mocha` namespace with submodule `Expect`.
> 
> We don't have very many test files. I think it is fine to add the code in your comment to the top of each file.

OK. I will change my pull request for your pull request (:trollface:) accordingly.

TysonMN:
> Would it be OK to copy the implementation of the assertions to our TestDsl.fs file into `Expect` module and once we have more functions in the module we can make a merge request for Fable.Mocha?

Yes. This seems fine.

TysonMN:
The most recent commit in this branch adds a workaround for the Fable compiler so that the tests pass.  This is now ready to be merged.

src/Hedgehog/Property.fs
```diff
@@ -56,15 +56,18 @@ module Property =
...
     let private mapGen
             (f : Gen<Journal * Outcome<'a>> -> Gen<Journal * Outcome<'b>>)
-            (x : Property<'a>) : Property<'b> =
-        toGen x |> f |> ofGen
+            (p : Property<'a>) : Property<'b> =
+        p |> toGen |> f |> ofGen
 
     let map (f : 'a -> 'b) (x : Property<'a>) : Property<'b> =
-        (mapGen << GenTuple.mapSnd << Outcome.map) f x
+        f |> Outcome.map |> GenTuple.mapSnd |> mapGen <| x
```
adam-becker, line 67: The `<|` here should be removed, since we went through the trouble of removing the others a while back 😄

src/Hedgehog/Property.fs
```diff
@@ -78,8 +81,11 @@ module Property =
             | Success x ->
                 GenTuple.mapFst (Journal.append journal) (k x))
 
+    let handle (e : exn) =
+        (Journal.singletonMessage (string e), Failure) |> Gen.constant
+
     let bind (k : 'a -> Property<'b>) (m : Property<'a>) : Property<'b> =
-        bindGen (toGen << k) (toGen m) |> ofGen
+        m |> toGen |> bindGen (fun a -> (try k a |> toGen with e -> handle e)) |> ofGen
```
adam-becker, line 88: There's enough going on with this line to warrant multiple lines.

src/Hedgehog/Property.fs
```diff
@@ -78,8 +81,11 @@ module Property =
             | Success x ->
                 GenTuple.mapFst (Journal.append journal) (k x))
 
+    let handle (e : exn) =
```
adam-becker, line 84: Should this be `private`?

src/Hedgehog/Property.fs
```diff
@@ -56,15 +56,18 @@ module Property =
...
-            (x : Property<'a>) : Property<'b> =
-        toGen x |> f |> ofGen
+            (p : Property<'a>) : Property<'b> =
+        p |> toGen |> f |> ofGen
 
     let map (f : 'a -> 'b) (x : Property<'a>) : Property<'b> =
-        (mapGen << GenTuple.mapSnd << Outcome.map) f x
+        f |> Outcome.map |> GenTuple.mapSnd |> mapGen <| x
+
+    let set (a: 'a) (property : Property<'b>) : Property<'a> =
```
adam-becker, line 69: Should this be `private`?

adam-becker (changes requested):
Found a few things, mostly around exposing new public API.

src/Hedgehog/Property.fs
adam-becker: Also, I'd extract the lambda to an inner function so the `try`/`with` can be on separate lines as well.

src/Hedgehog/Property.fs
TysonMN: Done

src/Hedgehog/Property.fs
TysonMN: Done

src/Hedgehog/Property.fs
TysonMN: Done

src/Hedgehog/Property.fs
TysonMN: If `Property<>` remains generic, then I think so, but since we are considering removing the generics, sure.  Done.

TysonMN:
@adam-becker, I think I addressed all the comment in your review.  I left them unresolved so that is easy for you to double check them.

When you are satisfied, I would prefer to squash the commits a bit.  Then we can merge.

TysonMN:
I decided to squash the commits.  I think this is ready to merge.

TysonMN:
Rebased on `master` (for clarity) even though there were no conflicts

adam-becker (approved):

adam-becker:
Looks good to me! Thanks @TysonMN

Issue #268, opened 2021-01-13
F# 5 applicative computation expression


adam-becker:
With F# 5 on the horizon, I think we should consider adding support for the new applicative computation expressions therein. This is mostly a tracking issue so we don't forget. Brief notes on this are [here][applicative].

[applicative]: https://devblogs.microsoft.com/dotnet/announcing-f-5/#applicative-computation-expressions

TysonMN:
Is this a duplicate of #267?

TysonMN:
For which of our monadic types is its monadic behavior different from its application behavior?  I do not know of any but I could be missing something.

TysonMN:
Oh, I realized how the applicative behavior would differ from the monadic behavior.

Hedgehog currently shrinks "multiple" generators sequentially.  This is required for the monadic behavior. The applicative behavior could be different; it could for example, cyclically shrink the generators until no improvement is possible (i.e. a Pareto optimal position).

I find this very interesting.

TysonMN:
@moodmosaic, does Haskell Hedgehog have this applicative behavior?

moodmosaic:
>cyclically shrink the generators until no improvement is possible

That should be possible according to @edsko's [article](https://well-typed.com/blog/2019/05/integrated-shrinking/#example-generating-pairs) (which is _the_ reference). Though I don't have ghci in front me.

adam-becker:
I was thinking of how this could differ, especially wrt `apply`'s implied behavior of being the "parallel" equivalent to `bind`'s "sequential" behavior. Perhaps all of the applied parameters receive the same seed and grow together? Unless that's what was already mentioned.

That would be an interesting way to control the integrated shrinking behavior of Hedgehog.

TysonMN:
My intuition is that the applicative behavior of `Random<>` should match its monadic behavior.

 > Perhaps all of the applied parameters receive the same seed [...]

I don't think that would be a good idea. Then applicatively generating a list would produce a list with the same value throughout.

> [...] and grow together?

I don't know what you mean by "grow".

TysonMN:
This is included in [version 0.12.1](https://www.nuget.org/packages/Hedgehog/0.12.1).

PR #338, opened 2021-09-08
Implement `Gen.mapN` variants with `Gen.apply`

adam-becker:
In anticipation of something like #268, I think we should implement the `mapN` variants with `apply` instead of `bind`. The parameters passed in aren't sequentially dependent so `bind` doesn't make sense, though it works.

There is also less closure capturing with this approach, for whatever that's worth.


TysonMN:
Excellent. I noticed that as well.

I approve (which my phone is currently unable to do programmatically).

PR #344, opened 2021-09-11
allow tests to access internal scope

TysonMN:
This is a change that I am splitting off of draft PR #336.  I think it is ready to merge now.

This PR allows the `Hedgehog.Tests` project to access types and functions in the `Hedgehog` project that have internal scope.  This is helpful because we can be more conservative about adding to the public API while still adding tests for this code.


adam-becker:
:+1:

PR #345, opened 2021-09-11
change access scope for two top-level modules from private to internal

TysonMN:
This is a change that I am splitting off of draft PR #336. I think it is ready to merge now.

This PR changes the access scope of two top-level modules from private to internal.  I didn't know that top-level modules could be private.  It seems to me that private and internal mean the same thing for top-level modules.  However, Visual Studio displayed the functions as transparent as though they were unused and returned no search results for references.  Strangely though, it was possible to navigate from a reference to one of these definitions.  None of these problems exist after changing the scope to internal.


adam-becker:
:+1:

PR #346, opened 2021-09-11
Simplifying property configuration

### LINKED ISSUE #343: Simplifying property configuration
### Discussed in https://github.com/hedgehogqa/fsharp-hedgehog/discussions/342

<div type='discussions-op-text'>

<sup>Originally posted by **adam-becker** September  9, 2021</sup>
I've been skimming over the property module, and I think it needs some TLC. We have no fewer than 8 `report` methods, and I think we can probably do better here. What I was thinking is that we could maybe make combinators that allow configuring a property, something like:

```fsharp
let property = // ...
property
|> Property.withTests 500
|> Property.withRecheck seed size
|> Property.check
```

This would translate into the LINQ interface like:

```csharp
var property = // ...
property
    .WithTests(500)
    .WithRecheck(seed, size)
    .Check();
```

I'm curious to hear thoughts on this. Apologies up front if this was suggested before.

/cc @moodmosaic @TysonMN @dharmaturtle </div>

dharmaturtle:
Closes #343

Okay, I got sucked into this.

Marking as a draft cause I still need to

* update tutorial/documentation
* update changelog

Buuuuut for the most part I think it's mostly done.

### DIFF
--- a/src/Hedgehog/Linq/Property.fs
+++ b/src/Hedgehog/Linq/Property.fs
@@ -72,17 +72,8 @@ type PropertyExtensions private () =
         Property.report property
 
     [<Extension>]
-    static member Report (property : Property, config : Hedgehog.PropertyConfig) : Report =
-        let (Property property) = property
-        Property.reportWith config property
-
-    [<Extension>]
     static member Report (property : Property<bool>) : Report =
         Property.reportBool property
-
-    [<Extension>]
-    static member Report (property : Property<bool>, config : Hedgehog.PropertyConfig) : Report =
-        Property.reportBoolWith config property
 
     [<Extension>]
     static member Check (property : Property) : unit =
@@ -90,17 +81,8 @@ type PropertyExtensions private () =
         Property.check property
 
     [<Extension>]
-    static member Check (property : Property, config : Hedgehog.PropertyConfig) : unit =
-        let (Property property) = property
-        Property.checkWith config property
-
-    [<Extension>]
     static member Check (property : Property<bool>) : unit =
         Property.checkBool property
-
-    [<Extension>]
-    static member Check (property : Property<bool>, config : Hedgehog.PropertyConfig) : unit =
-        Property.checkBoolWith config property
 
     [<Extension>]
     static member Recheck (property : Property, size : Size, seed : Seed) : unit =
@@ -108,17 +90,8 @@ type PropertyExtensions private () =
         Property.recheck size seed property
 
     [<Extension>]
-    static member Recheck (property : Property, size : Size, seed : Seed, config : Hedgehog.PropertyConfig) : unit =
-        let (Property property) = property
-        Property.recheckWith size seed config property
-
-    [<Extension>]
     static member Recheck (property : Property<bool>, size : Size, seed : Seed) : unit =
         Property.recheckBool size seed property
-
-    [<Extension>]
-    static member Recheck (property : Property<bool>, size : Size, seed : Seed, config : Hedgehog.PropertyConfig) : unit =
-        Property.recheckBoolWith size seed config property
 
     [<Extension>]
     static member ReportRecheck (property : Property, size : Size, seed : Seed) : Report =
@@ -126,17 +99,8 @@ type PropertyExtensions private () =
         Property.reportRecheck size seed property
 
     [<Extension>]
-    static member ReportRecheck (property : Property, size : Size, seed : Seed, config : Hedgehog.PropertyConfig) : Report =
-        let (Property property) = property
-        Property.reportRecheckWith size seed config property
-
-    [<Extension>]
     static member ReportRecheck (property : Property<bool>, size : Size, seed : Seed) : Report =
         Property.reportRecheckBool size seed property
-
-    [<Extension>]
-    static member ReportRecheck (property : Property<bool>, size : Size, seed : Seed, config : Hedgehog.PropertyConfig) : Report =
-        Property.reportRecheckBoolWith size seed config property
 
     [<Extension>]
     static member Render (property : Property) : string =
@@ -144,17 +108,8 @@ type PropertyExtensions private () =
         Property.render property
 
     [<Extension>]
-    static member Render (property : Property, config : Hedgehog.PropertyConfig) : string =
-        let (Property property) = property
-        Property.renderWith config property
-
-    [<Extension>]
     static member Render (property : Property<bool>) : string =
         Property.renderBool property
-
-    [<Extension>]
-    static member Render (property : Property<bool>, config : Hedgehog.PropertyConfig) : string =
-        Property.renderBoolWith config property
 
     [<Extension>]
     static member Where (property : Property<'T>, filter : Func<'T, bool>) : Property<'T> =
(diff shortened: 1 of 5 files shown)


src/Hedgehog/Linq/Property.fs
```diff
@@ -71,91 +71,46 @@ type PropertyExtensions private () =
         let (Property property) = property
         Property.report property
```
dharmaturtle, line 73: It might actually be nice to keep the `config` overloads. Thoughts?

src/Hedgehog/Property.fs
```diff
@@ -4,15 +4,15 @@ open System
...
 type Property<'a> =
-    | Property of Gen<Journal * Outcome<'a>>
+    | Property of PropertyConfig * Gen<Journal * Outcome<'a>>
 
 module Property =
 
-    let ofGen (x : Gen<Journal * Outcome<'a>>) : Property<'a> =
-        Property x
+    let ofGen (gen : Gen<Journal * Outcome<'a>>) : Property<'a> =
+        Property (PropertyConfig.defaultConfig, gen)
```
dharmaturtle, line 12: I gave the `PropertyConfig.defaultConfig` zero thought.

src/Hedgehog/Property.fs
```diff
@@ -254,9 +224,25 @@ module Property =
...
-        reportBoolWith config p
-        |> Report.render
+    /// Set the number of times a property is allowed to shrink before the test
+    /// runner gives up and displays the counterexample.
+    let withShrinks (shrinkLimit : int<shrinks>) (Property (config, gen) : Property<'a>) : Property<'a> =
+        let config = { config with ShrinkLimit = Some shrinkLimit }
+        Property (config, gen)
+
+    /// Restores the default shrinking behavior.
+    let withoutShrinks (Property (config, gen) : Property<'a>) : Property<'a> =
```
dharmaturtle, line 234: I think this should be called `withDefaultShrinks`.

src/Hedgehog/Property.fs
```diff
@@ -254,9 +224,25 @@ module Property =
         reportBool property
         |> Report.render
 
-    let renderBoolWith (config : PropertyConfig) (p : Property<bool>) : string =
-        reportBoolWith config p
-        |> Report.render
+    /// Set the number of times a property is allowed to shrink before the test
+    /// runner gives up and displays the counterexample.
+    let withShrinks (shrinkLimit : int<shrinks>) (Property (config, gen) : Property<'a>) : Property<'a> =
```
dharmaturtle, line 229: Might be nice to just take an `int` instead of an `int<shrinks>`

src/Hedgehog/Property.fs
```diff
@@ -4,15 +4,15 @@ open System
 
 [<Struct>]
 type Property<'a> =
-    | Property of Gen<Journal * Outcome<'a>>
+    | Property of PropertyConfig * Gen<Journal * Outcome<'a>>
```
TysonMN, line 7: What about making this into a record?  How about this?
```fsharp
{ Config: PropertyConfig
  Generator: Gen<Journal * Outcome<'a>> }
```

adam-becker: I'm okay with this, I think we should also mark the internals as `private`, and use lower camel case for the names:

```fsharp
type Property<'a> = {
    config: PropertyConfig
    generator: Gen<Journal * Outcome<'a>>
}
```

In any case, we should make 2 accessor functions, and only use them. They can be marked as `inline` if we have performance concerns.

```fsharp
let inline config (property : Property<'a>) : PropertyConfig =
    property.config

// This already exists as 'toGen'
let inline generator (property : Property<'a>) : PropertyConfig =
    property.generator
```

Thoughts?

src/Hedgehog/Property.fs
adam-becker: I'm okay with this, the current name `withoutShrinks` sounds as if no shrinking will occur.

src/Hedgehog/Property.fs
adam-becker: We'll need to audit the use of this function, if it's being used as `prop |> toGen |> doGenThing |> ofGen`, then we'll lose the original configuration.

tests/Hedgehog.Tests/ShrinkTests.fs
```diff
@@ -199,14 +199,11 @@ let shrinkTests = testList "Shrink tests" [
...
 
-        let propConfig =
-            PropertyConfig.defaultConfig
-            |> PropertyConfig.withShrinks shrinkLimit
         let report =
             property {
                 let! actual = Range.linear 1 1_000_000 |> Gen.int
                 return actual < 500_000
-            } |> Property.reportWith propConfig
+            } |> Property.withShrinks shrinkLimit |> Property.report
```
adam-becker, line 206: These two `|>` should be on separate lines for readability.

src/Hedgehog/Linq/Property.fs
```diff
@@ -184,4 +139,40 @@ type PropertyExtensions private () =
...
+    
+    [<Extension>]
+    static member WithTests (property : Property, testLimit : int<tests>) : Property =
+        let (Property property) = property
+        Property (Property.withTests testLimit property)
+    
+    [<Extension>]
+    static member WithConfig (property : Property, config : PropertyConfig) : Property =
+        let (Property property) = property
+        Property (Property.withConfig config property)
```
adam-becker, line 160: I think we can get rid of this, and only expose configuration with `Property.with*`

src/Hedgehog/Linq/Property.fs
adam-becker: I think we should only expose a single way to configure properties; this new API should be the only API.

adam-becker (changes requested):
Let me just say, I love the work done here. So many extra methods/functions are now gone! 🚀 

We'll need to look at this carefully, so the review may take a while but I have no doubt that this is the right way to go. Thank you very much for doing this @dharmaturtle!

src/Hedgehog/Property.fs
adam-becker: Can you expand on this?

src/Hedgehog/Property.fs
TysonMN: I am ok with removing the public access scope.  However, I think the access scope should be `internal` instead of private.

I think we should follow the [F# naming guidelines](https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting#use-pascalcase-for-type-declarations-members-and-labels) and use Pascal case for record field names.

The most common operation is mapping, not (just) getting.  As such, I think we should create a submodule inside Property like this (@dharmaturtle knows what I mean):
```fsharp
module Config =
  let get p = p.Config
  let set p c = { p with Config = c }
  let map f p = { p with Config = f p.Config }
```

src/Hedgehog/Linq/Property.fs
TysonMN: This is the same as `Property.Config.set` (as I strucutured/named it in another comment).  I think it is better to keep.  Might want to consider renaming to `setConfig` though.

src/Hedgehog/Property.fs
TysonMN: I think a better name that `withDefaultShrinks` (which is a syntactic name) is `withUnlimitedShrinking` (which is a semantic name)

src/Hedgehog/Linq/Property.fs
TysonMN: I am not exactly sure what is being discussed here.  However, I am not overly concerned with the C# API.

src/Hedgehog/Property.fs
TysonMN: Indeed.  All such calls should be replaced with `Property.Generator.map` (as I structured/named it in another comment).

src/Hedgehog/Linq/Property.fs
adam-becker: What would the value be in overriding a user's preferences?

TysonMN: It would be the user calling this function, it would be their preference.  For example, @dharmaturtle talks about how he runs PB tests with a time limit during development and with a test limit during integration.  I think the minimal amount of duplicated code to achieve this would be constructing the correct `PropertyConfig` instance in a global location and then referencing it from all the PB tests in order to set it, which (when using C#) would be achieved via this method.

adam-becker: You could also make an extension method of `Property` that configures them to all be the same and call that on each property you want it to apply to.

```csharp
public static Property ConfigureSpecifically(this Property property) =>
    property.WithTests(1000).WithoutShrinkLimit().Whatever();

// in a test

var someProperty = // ...
someProperty.ConfigureSpecifically().OverrideLocally().Check();
```

TysonMN: Good point.

I think I would still keep this myself, but I won't argue further in favor of keeping it.

src/Hedgehog/Property.fs
TysonMN: How about, it a PR before this one, we change `Property` from a single case discriminated union to a record with a single field as well as expressing all maps over that field as such?  Then I think it would easier to verify where `PropertyConfig.defaultConfig` should be used.

dharmaturtle:
Gonna let #353 go first as it's smaller, more mature, and may possibly establish designs/patterns that this will follow.

PR #347, opened 2021-09-11
improve function name takeSmallest to shrinkInput in Property

TysonMN:
This is a change that I am splitting off of draft PR #336. I think it is ready to merge now.

I think this is a better name for this (private) function.


adam-becker:
👍🏼

PR #348, opened 2021-09-11
Add error handling to Property.map

TysonMN:
This PR is similar to PR #328, which fixed bug #327 by adding error handling to `Property.bind`.  Specifically, it adds essentially the same error handling to `Property.map`.  The first commit adds a failing test and the second commit makes that test pass.

This work is also adjacent to draft PR #336.  That draft currently has code similar (but weaker than due to its use of `Property.bind`) to error handling code in this PR.


adam-becker:
:+1:

PR #349, opened 2021-09-11
Add contributing guidelines.

adam-becker:
To make getting started easier for newcomers, I've written some contributing guidelines. I don't want a very heavy process, but some process would make things clearer and prevent unnecessary work.

I think we could be a little clearer in the `Script.fsx` part.

We may want to start keeping track of benchmark output so we have a good baseline to compare new work to. Performance is going to be a metric I'm going to pay more attention to going forward.

/cc @moodmosaic @TysonMN


TysonMN (approved):

moodmosaic (approved):
👍🏋❤

PR #350, opened 2021-09-11
Update benchmark project

adam-becker:
Updated benchmark project to the latest `BenchmarkDotNet` version (`0.11.3` -> `0.13.1`). I also took the liberty to remove the explicit `FSharp.Core` reference, and I added another benchmark case for the `ForLoopTest`.

The `ForLoopTest` results show that we have some work to do there.. but that will be addressed in the future.

/cc @moodmosaic @TysonMN


TysonMN:
Approve

moodmosaic (commented):
LGTM 👍

PR #351, opened 2021-09-11
Deprecate gen functions

adam-becker:
To make the `Gen` API more clear, I've decided to deprecate the `Gen.int` and `Gen.float` functions.

- For `Gen.int` I've added a new function called `Gen.int32`, so the naming is consistent across integer types.
- For `Gen.float`, I don't see any reason for it since we already have `Gen.double`.

Instead of outright removing these functions, I've marked them with `Obsolete` attributes, to make the upgrade path a little friendlier. The plan is to remove them completely following the `0.11.0` release.

I'm still undecided on `Gen.byte` and `Gen.sbyte`.


src/Hedgehog/Gen.fs
```diff
@@ -496,7 +502,7 @@ module Gen =
               min
                 (14L * 60L)
                 ((ticks - DateTimeOffset.MinValue.Ticks) / TimeSpan.TicksPerMinute)
-            let! offsetMinutes = int (Range.linearFrom 0 (Operators.int minOffsetMinutes) (Operators.int maxOffsetMinutes))
+            let! offsetMinutes = int32 (Range.linearFrom 0 (Operators.int minOffsetMinutes) (Operators.int maxOffsetMinutes))
             return DateTimeOffset(ticks, TimeSpan.FromMinutes (Operators.float offsetMinutes))
```
adam-becker, line 506: The explicit `Operators.*` will also be removed once these functions are gone.

TysonMN:
Approve

Issue #326, opened 2021-03-24
Generalized formatting of generated values


adam-becker:
Following a quick discussion from #324, this issue is being created to provide generalized formatting for values from a `Gen<'a>`.

Ideally, we'd change the `Gen<'a>` type to be a record, and add a `formatter` field that looks like `'a -> string`. To maintain compatibility, this field should have a default value of `sprintf "%A"`. Another function should be exposed to override this default, `withFormatter : ('a -> string) -> Gen<'a> -> Gen<'a>`.

Any values from a generator that are formatted should use the provided formatter.

TysonMN:
> Ideally, we'd change the `Gen<'a>` type to be a record, and add a `formatter` field that looks like `'a -> string`.

I am not convinced that this is the right approach.  My first impression is that we should make a change like that to `Property<_>` or the functions of `Property`.  It is those functions that want to `string` representation.

How does Haskell-Hedgehog solve this problem?

adam-becker:
> > Ideally, we'd change the `Gen<'a>` type to be a record, and add a `formatter` field that looks like `'a -> string`.
> 
> I am not convinced that this is the right approach. My first impression is that we should make a change like that to `Property<_>` or the functions of `Property`. It is those functions that want to `string` representation.

Gen would have better knowledge of the `'a` being generated, I don't think `Property` is parameterized on the type being generated. I think of it as "gen can make a thing, and also a string representation of the thing".

> How does Haskell-Hedgehog solve this problem?

They probably just use type classes, like `Show` and don't have this problem at all. The equivalent in .NET would be to override `ToString`, which we can't do for types we don't control.

dharmaturtle:
I'm thinking about the interplay between `GenConfig` and `PropertyConfig`. Might we have something like `PropertyConfig.withGenConfig`, or is that too much 🤪

Also, would it be reasonable to eventually add `size` and/or `seed` to `GenConfig`?

...and if so, how would we explain to users why they're on `GenConfig` and not `PropertyConfig`? 😅

Issue #335, opened 2021-09-06
Use-case specific naming in Tree


TysonMN:
I would prefer if the `Tree` type, module, and function used language that was use-case agnostic.  Instead, for example, the function that returns the root of the tree is called `outcome`.

adam-becker:
I noticed this naming as well. Bad naming isn't limited to just the tree module, of course. But I'm okay with changing the names in this module to be more general. What names would you propose we use here going forward?

TysonMN:
From memory, there are the renames that I recall wanting.
`Tree.outcome` -> `Tree.root`
`Tree.shrinks` -> `Tree.children`

PR #353, opened 2021-09-11
Generalized formatting of generated values

adam-becker:
Opening this PR to get feedback early. This is to address #326.

I tried adding this at a `Property` level, but its type argument doesn't track the type being generated so it doesn't work there.

/cc @moodmosaic  @TysonMN

### DIFF
--- a/src/Hedgehog/Gen.fs
+++ b/src/Hedgehog/Gen.fs
@@ -361,6 +361,12 @@ module Gen =
         Random.sized sizedList
         |> ofRandom
 
+    /// Generates a `System.Collections.Generic.List<T>` using a 'Range' to determine the length.
+    let resizeArray (range : Range<int>) (g : Gen<'a>) : Gen<ResizeArray<'a>> =
+        list range g
+        |> map ResizeArray
+        |> withFormatter (List.ofSeq >> sprintf "%A")
+
     /// Generates an array using a 'Range' to determine the length.
     let array (range : Range<int>) (g : Gen<'a>) : Gen<array<'a>> =
         list range g |> map Array.ofList
--- a/src/Hedgehog/Gen.fs
+++ b/src/Hedgehog/Gen.fs
@@ -2,18 +2,36 @@ namespace Hedgehog
 
 open System
 
+type GenConfig<'a> = private {
+    Formatter : 'a -> string
+}
+
 /// A generator for values and shrink trees of type 'a.
 [<Struct>]
-type Gen<'a> =
-    | Gen of Random<Tree<'a>>
+type Gen<'a> = private {
+    Random : Random<Tree<'a>>
+    Config : GenConfig<'a>
+}
 
 module Gen =
 
-    let ofRandom (r : Random<Tree<'a>>) : Gen<'a> =
-        Gen r
+    let format (a : 'a) (gen : Gen<'a>) : string =
+        gen.Config.Formatter a
 
-    let toRandom (Gen r : Gen<'a>) : Random<Tree<'a>> =
-        r
+    let withFormatter (formatter : 'a -> string) (gen : Gen<'a>) : Gen<'a> =
+        {
+            gen with Config = { gen.Config with Formatter = formatter }
+        }
+
+    let ofRandom (r : Random<Tree<'a>>) : Gen<'a> = {
+        Random = r
+        Config = {
+            Formatter = sprintf "%A"
+        }
+    }
+
+    let toRandom (gen : Gen<'a>) : Random<Tree<'a>> =
+        gen.Random
 
     let delay (f : unit -> Gen<'a>) : Gen<'a> =
         Random.delay (toRandom << f) |> ofRandom
--- a/src/Hedgehog/Gen.fs
+++ b/src/Hedgehog/Gen.fs
@@ -2,32 +2,35 @@ namespace Hedgehog
 
 open System
 
-type GenConfig<'a> = private {
-    Formatter : 'a -> string
-}
-
 /// A generator for values and shrink trees of type 'a.
 [<Struct>]
 type Gen<'a> = private {
-    Random : Random<Tree<'a>>
     Config : GenConfig<'a>
+    Random : Random<Tree<'a>>
 }
 
 module Gen =
+    module Config =
+        let get (gen : Gen<'a>) : GenConfig<'a> =
+            gen.Config
+
+        let map (f : GenConfig<'a> -> GenConfig<'a>) (gen : Gen<'a>) : Gen<'a> =
+            { gen with Config = f (get gen) }
+
+        let set (config : GenConfig<'a>) (gen : Gen<'a>) : Gen<'a> =
+            map (always config) gen
 
     let format (a : 'a) (gen : Gen<'a>) : string =
-        gen.Config.Formatter a
+        let formatter = gen |> Config.get |> GenConfig.getFormatter
+        formatter a
 
     let withFormatter (formatter : 'a -> string) (gen : Gen<'a>) : Gen<'a> =
-        {
-            gen with Config = { gen.Config with Formatter = formatter }
-        }
+        gen
+        |> Config.map (GenConfig.setFormatter formatter)
 
-    let ofRandom (r : Random<Tree<'a>>) : Gen<'a> = {
-        Random = r
-        Config = {
-            Formatter = sprintf "%A"
-        }
+    let ofRandom (random : Random<Tree<'a>>) : Gen<'a> = {
+        Config = GenConfig.defaultConfig ()
+        Random = random
     }
 
     let toRandom (gen : Gen<'a>) : Random<Tree<'a>> =
--- a/src/Hedgehog/Gen.fs
+++ b/src/Hedgehog/Gen.fs
@@ -33,7 +33,7 @@ module Gen =
         |> withFormatter (Seq.toList >> sprintf "%A")
 
     let ofRandom (random : Random<Tree<'a>>) : Gen<'a> = {
-        Config = GenConfig.defaultConfig ()
+        Config = GenConfig.defaultConfig
         Random = random
     }
 
--- a/src/Hedgehog/Gen.fs
+++ b/src/Hedgehog/Gen.fs
@@ -28,6 +28,10 @@ module Gen =
         gen
         |> Config.map (GenConfig.setFormatter formatter)
 
+    let withListFormatter (gen : Gen<_>) : Gen<_> =
+        gen
+        |> withFormatter (Seq.toList >> sprintf "%A")
+
     let ofRandom (random : Random<Tree<'a>>) : Gen<'a> = {
         Config = GenConfig.defaultConfig ()
         Random = random
@@ -368,7 +372,7 @@ module Gen =
     let resizeArray (range : Range<int>) (g : Gen<'a>) : Gen<ResizeArray<'a>> =
         list range g
         |> map ResizeArray
-        |> withFormatter (List.ofSeq >> sprintf "%A")
+        |> withListFormatter
 
     /// Generates an array using a 'Range' to determine the length.
     let array (range : Range<int>) (g : Gen<'a>) : Gen<array<'a>> =
--- a/src/Hedgehog/GenConfig.fs
+++ b/src/Hedgehog/GenConfig.fs
@@ -1,0 +1,16 @@
+namespace Hedgehog
+
+type GenConfig<'a> = private {
+    Formatter : 'a -> string
+}
+
+module GenConfig =
+
+    let defaultConfig () =
+        { Formatter = sprintf "%A" }
+
+    let getFormatter (config : GenConfig<'a>) : ('a -> string) =
+        config.Formatter
+
+    let setFormatter (formatter : 'a -> string) (config : GenConfig<'a>) : GenConfig<'a> =
+        { config with Formatter = formatter }
--- a/src/Hedgehog/GenConfig.fs
+++ b/src/Hedgehog/GenConfig.fs
@@ -6,8 +6,9 @@ type GenConfig<'a> = private {
 
 module GenConfig =
 
-    let defaultConfig () =
-        { Formatter = sprintf "%A" }
+    let defaultConfig<'a> =
+        let formatter: 'a -> string = sprintf "%A"
+        { Formatter = formatter }
 
     let getFormatter (config : GenConfig<'a>) : ('a -> string) =
         config.Formatter
--- a/src/Hedgehog/Property.fs
+++ b/src/Hedgehog/Property.fs
@@ -98,26 +98,9 @@ module Property =
         |> bindGen kTry
         |> ofGen
 
-    let private printValue (value) : string =
-        // sprintf "%A" is not prepared for printing ResizeArray<_> (C# List<T>) so we prepare the value instead
-        let prepareForPrinting (value: obj) : obj =
-        #if FABLE_COMPILER
-            value
-        #else
-            let t = value.GetType()
-            // have to use TypeInfo due to targeting netstandard 1.6
-            let t = System.Reflection.IntrospectionExtensions.GetTypeInfo(t)
-            let isList = t.IsGenericType && t.GetGenericTypeDefinition() = typedefof<ResizeArray<_>>
-            if isList
-            then value :?> System.Collections.IEnumerable |> Seq.cast<obj> |> List.ofSeq :> obj
-            else value
-        #endif
-
-        value |> prepareForPrinting |> sprintf "%A"
-
     let forAll (k : 'a -> Property<'b>) (gen : Gen<'a>) : Property<'b> =
         let prepend (x : 'a) =
-            counterexample (fun () -> printValue x)
+            counterexample (fun () -> gen |> Gen.format x)
             |> set x
             |> bind k
             |> toGen
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -7,7 +7,8 @@
 
   "extensions": [
     "editorconfig.editorconfig",
-    "ionide.ionide-fsharp@4.17.0"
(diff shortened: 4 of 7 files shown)


src/Hedgehog/Gen.fs
```diff
@@ -2,18 +2,36 @@
...
-        Gen r
+    let format (a : 'a) (gen : Gen<'a>) : string =
+        gen.Config.Formatter a
+
+    let withFormatter (formatter : 'a -> string) (gen : Gen<'a>) : Gen<'a> =
+        {
+            gen with Config = { gen.Config with Formatter = formatter }
+        }
+
+    let ofRandom (r : Random<Tree<'a>>) : Gen<'a> = {
```
adam-becker, line 26: As noted in #346, this needs to be traced carefully so we don't end up overriding the configuration.

src/Hedgehog/Gen.fs
```diff
@@ -2,18 +2,36 @@
...
+    Formatter : 'a -> string
+}
+
 /// A generator for values and shrink trees of type 'a.
 [<Struct>]
-type Gen<'a> =
-    | Gen of Random<Tree<'a>>
+type Gen<'a> = private {
+    Random : Random<Tree<'a>>
+    Config : GenConfig<'a>
```
TysonMN, line 13: nit: I would like the order of these fields and the ones for `Property<>` to be the same

.devcontainer/Dockerfile
```diff
@@ -15,3 +15,5 @@ RUN groupadd -g 1000 dev \
  && chown -R dev:dev /home/dev
 
 USER dev
+
+ENV DOTNET_CLI_TELEMETRY_OPTOUT=1
```
TysonMN, line 19: Why this and the other docker changes?

src/Hedgehog/Gen.fs
```diff
@@ -2,18 +2,36 @@
...
-        Gen r
+    let format (a : 'a) (gen : Gen<'a>) : string =
+        gen.Config.Formatter a
 
-    let toRandom (Gen r : Gen<'a>) : Random<Tree<'a>> =
-        r
+    let withFormatter (formatter : 'a -> string) (gen : Gen<'a>) : Gen<'a> =
+        {
+            gen with Config = { gen.Config with Formatter = formatter }
+        }
```
TysonMN, line 24: I would prefer the structure and give in the PR for Property.

src/Hedgehog/Property.fs
```diff
@@ -101,26 +101,9 @@ module Property =
...
-            let t = value.GetType()
-            // have to use TypeInfo due to targeting netstandard 1.6
-            let t = System.Reflection.IntrospectionExtensions.GetTypeInfo(t)
-            let isList = t.IsGenericType && t.GetGenericTypeDefinition() = typedefof<ResizeArray<_>>
-            if isList
-            then value :?> System.Collections.IEnumerable |> Seq.cast<obj> |> List.ofSeq :> obj
-            else value
-        #endif
-
-        value |> prepareForPrinting |> sprintf "%A"
```
TysonMN, line 119: I don't see where you are preserving this behavior

adam-becker: `Gen.resizeArray`. This came up since the `Hedgehog.Linq` version of `Gen.List` generates the BCL version of list, and the formatting for that just prints the type name. In `Gen.resizeArray` a formatter is specified that converts the BCL list to an F# list for formatting.

.devcontainer/Dockerfile
adam-becker: Oops. I'll move them to a separate PR. The gist is that the latest C# extension has broken intellisense, and I don't like being spied on with telemetry. 😜

src/Hedgehog/Property.fs
TysonMN: I think you are assuming that the only way a user could generate a `ResizeArray<>` is by calling `Gen.resizeArray`.  That assumption is false.

src/Hedgehog/Gen.fs
adam-becker: Done

src/Hedgehog/Property.fs
adam-becker: I turned the list formatter into a function called `Gen.withListFormatter`, so any `IEnumerable` can get the same formatting behavior. If you just generate a list using `Gen.resizeArray`, then you're good to go by default.

A test case has been added for the new function.

TysonMN: I will try to be more specific.  The following test passes on `master` but fails on the branch in this PR (at commit 88ad7b19cc86e6698e4f561e90b0e718fb233c45).  I think this test should continue to pass.  I don't think we should purposefully introduce a regression.

```fsharp
fableIgnore "generated seq of five elements converted to C# list is not abbreviated in the failure report without calling Gen.withListFormatter" <| fun _ ->
    let report =
        property {
            let gen = Gen.int32 (Range.singleton 0)
            let! xs = Gen.seq (Range.singleton 5) gen |> Gen.map ResizeArray
            return false
        }
        |> Property.renderWith (PropertyConfig.withShrinks 0<shrinks> PropertyConfig.defaultConfig)
    Expect.isNotMatch report "\.\.\." "Abbreviation (...) found"
```

src/Hedgehog/GenConfig.fs
```diff
@@ -0,0 +1,16 @@
+namespace Hedgehog
+
+type GenConfig<'a> = private {
+    Formatter : 'a -> string
+}
+
+module GenConfig =
+
+    let defaultConfig () =
+        { Formatter = sprintf "%A" }
```
dharmaturtle, line 10: For me, when I see something with the signature `unit -> 'T`, I think it side effects somewhere. Since this doesn't, would it be reasonable to drop `()`?

src/Hedgehog/Gen.fs
```diff
@@ -4,16 +4,41 @@ open System
...
 [<Struct>]
-type Gen<'a> =
-    | Gen of Random<Tree<'a>>
+type Gen<'a> = private {
+    Config : GenConfig<'a>
+    Random : Random<Tree<'a>>
+}
 
 module Gen =
+    module Config =
```
dharmaturtle, line 13: `PropertyConfig` exists. In the interest of consistency, I think we should have this either be `GenConfig` or change `PropertyConfig` to `Property.Config`. I _believe_ [when I wrote the original](https://github.com/hedgehogqa/fsharp-hedgehog/pull/288) `PropertyConfig`, I was simply copying what Haskell did. I don't feel strongly in either direction.

src/Hedgehog/GenConfig.fs
TysonMN: This is running into F# compiler's inability to infer generic types for values.  The easiest workaround, as was done there, is to change it from a value to a function (by accepting the value `()`).  Another workaround is to explicitly specify the type parameters (so that the F# compiler doesn't have to do any inference).  I will give one possibility in a different comment thread (that includes line 9).

TysonMN: > I will give one possibility in a different comment thread (that includes line 9).

Actually, that is not possible.  Either a bug or missing feature in GitHub is that a new multiline comment thread cannot be made if the last line already has a comment thread.

Here is one way to explicitly specify the type parameters.  Maybe there is a way to do this inline, but I don't know of such code at the moment.

```fsharp
    let defaultConfig<'a> =
        let f : 'a -> string = sprintf "%A"
        { Formatter = f }
```

dharmaturtle: This works:

```f#
let f : 'a -> string = sprintf "%A" in { Formatter = f }
```

I'm not necessarily advocating for it.

adam-becker: > This is running into F# compiler's inability to infer generic types for values.

I know what you mean, but this is _slightly_ wrong. The F# compiler can't emit generic values because support for this doesn't exist in MSIL. The equivalent in C# would be:

```csharp
public GenConfig<T> DefaultConfig<T> = new GenConfig<T> {
    Formatter = // ...
}
```

src/Hedgehog/Property.fs
adam-becker: I know exactly what you mean, but the previous solution uses magic. Any non-magic solution is going to fail to cover that use case, but I still want the magic removed. The previous version sets a bad precedent of special casing certain types in our code. Once that starts, where will it end? Say we want special `Dictionary` formatting, or X type formatting? We had to add what feels like a hack to support a single user's desire to have a list formatted, to me this is the wrong approach. Speaking of, what are the approaches? As I see it, they are:

Approach 1: Override `.ToString()`. This works for types that users have created themselves, but fails for types like `System.Collections.Generic.List<T>`, the original use case. This solution only gets us so far, and means some types are special in Hedgehog which I dislike.

Approach 2: Reflection magic. This works really well as long as we build in the special cases, but there will _always_ be more cases we don't support and this approach will never end. Whenever .NET adds a new "gotta have it" type, we'll be asked to add formatting for it. Worse still, users have no way around it until _we_ "fix" it. I don't want to be on the hook for this. I prefer to solve problems in the general case and then allow users to do the right thing for their code. Which brings us to..

Approach 3: Allow configuring a formatter if you want one. This is the most general approach, and gives us the ability to support custom formatters if need be. Users can also format their types however they choose without having to sacrifice `.ToString()` to support their tests. If `.ToString()` is used, then the default formatter covers that case. In a scenario where .NET adds a new shiny users won't be blocked on us, they can format the new shiny however they like and move on. We can be asked to support this internally, and we'll have all the time in the world to decide if we want to or not. This approach also covers cases where the same type can be formatted differently for different tests. Say you only want to see a certain property's output for one, and a different property's output for another. How is this done today? Can reflection support that scenario?

If there are other approaches I've missed, please share so we can weigh them against my chosen approach.

>  I don't think we should purposefully introduce a regression.

This is only a regression if you see reflection hacks as a solution. Our API is still unstable and at the end of the day we get to support these cases however we want until `1.0.0`. Lets try to get the API we want while we still can!

src/Hedgehog/GenConfig.fs
TysonMN: What did I say that is wrong?

src/Hedgehog/Property.fs
TysonMN: I don't think you will change your mind about this, so I will stop trying trying.

src/Hedgehog/GenConfig.fs
adam-becker: > This is running into F# compiler's inability to infer generic types for values.
> What did I say that is wrong?

It can infer the types just fine, the resulting construct can't be represented in MSIL.

dharmaturtle: > I'm not necessarily advocating for it.

To be clear - I still am advocating for the removal of `unit` as a param. This will also have better symmetry with `PropertyConfig.defaultConfig`.

src/Hedgehog/Property.fs
adam-becker: > I don't think you will change your mind about this, so I will stop trying trying.

I don't mean to be domineering, I think this can be a good discussion. I want to get this right before we merge it in.

Do you think functionality like this would be better represented in experimental? Similar to the way that the `Gen` infrastructure was already built upon for `Gen.auto`?

src/Hedgehog/GenConfig.fs
TysonMN: Do you think it is possible for the F# compiler to be changed so that
```fsharp
let defaultConfig = { Formatter = sprintf "%A" }
```

and
```fsharp
let defaultConfig = { Formatter = let f : 'a -> string = sprintf "%A" in { Formatter = f } }
```

emit the same MSIL?

src/Hedgehog/Property.fs
TysonMN: Moving the existing use of reflection to experimental is also a regression, so no.  I understand that the major version is still `0` so "anything" is allowed, but that doesn't mean we should.

In this case, I think it is easy to retain the current behavior in the next version along side the feature added in this PR.  Then users can switch to this feature before upgrading again.

src/Hedgehog/GenConfig.fs
adam-becker: I tried your examples (all permutations I could think of) and still got errors..

```fsharp
type Blah<'a> = { Formatter : 'a -> string }

let defaultConfig =
    { Formatter = let f : 'a -> string = sprintf "%A" in { Formatter = f } }
// This expression was expected to have type
//     ''a -> string'
// but here has type
//     'Blah<'b>'

let defaultConfig =
    { Formatter = let f : 'a -> string = sprintf "%A" in f }
// Value restriction. The value 'defaultConfig' has been inferred to have generic type
//     val internal defaultConfig : Blah<'_a>    
// Either define 'defaultConfig' as a simple data term, make it a function with explicit arguments or, if you do not intend for it to be generic, add a type annotation.

let defaultConfig =
    let f : 'a -> string = sprintf "%A" in { Formatter = f }
// Value restriction. The value 'defaultConfig' has been inferred to have generic type
//     val internal defaultConfig : Blah<'_a>    
// Either define 'defaultConfig' as a simple data term, make it a function with explicit arguments or, if you do not intend for it to be generic, add a type annotation.
```

> emit the same MSIL?

I only now realized I've been misspeaking. MSIL is the bytecode that is JIT'd at runtime to a host architecture. That's not what I meant, I meant the binary format of .NET assemblies. Fields/Properties can't be declared as generic. In theory properties _could_ support this since they themselves are just methods. However, the .NET teams have regularly issued resistance to any binary format changes, understandably so.

src/Hedgehog/Property.fs
adam-becker: > Moving the existing use of reflection to experimental is also a regression, so no.

Understood.

> I understand that the major version is still 0 so "anything" is allowed, but that doesn't mean we should.

We've both routinely violated that when we wanted. 😎

Kidding aside, I do want to minimize impacts of our changes between releases and I do want to stabilize the API as we move closer to `1.0.0`. But, I'm also not going to let that stand in the way of making changes where I think they need to be made, and I fully support breaking the API further for things like #335 so that we can get things right before we can't anymore.

> In this case, I think it is easy to retain the current behavior in the next version along side the feature added in this PR. Then users can switch to this feature before upgrading again.

So we add this feature in the next release, then take it away in the release right after? It's my understanding that this version of the fix has not yet been released, sorry if I've gotten that wrong.

TysonMN: > > In this case, I think it is easy to retain the current behavior in the next version along side the feature added in this PR. Then users can switch to this feature before upgrading again.
> 
> So we add this feature in the next release, then take it away in the release right after? It's my understanding that this version of the fix has not yet been released, sorry if I've gotten that wrong.

I mean you could add the feature in this PR while maintaining the current behavior that uses reflection, release a new version, remove the use of reflection, release a new version.

> But, I'm also not going to let that stand in the way of making changes where I think they need to be made, and I fully support breaking the API further for things like #335 so that we can get things right before we can't anymore.

I am have changed my mind.  I now think we should more aggressive changes while it is easier to do so.

src/Hedgehog/GenConfig.fs
TysonMN: > I tried your examples (all permutations I could think of) and still got errors..

You forgot to add an explicit type parameter on the value.

Some of our code blocks above also forgot to add the explicit type parameter is some cases.  In comment https://github.com/hedgehogqa/fsharp-hedgehog/pull/353#discussion_r711604381, which was my first comment with a code block in this thread, I did not forget the explicit type parameter.

TysonMN: > Fields/Properties can't be declared as generic.

I would typically interpret this to mean "the type" of a field/property cannot be generic.  Of course that statement is false.  I think what you mean is that in both C# and F#, neither a field nor a property can "pick up" or "introduce" a new type parameter when referenced.  I think that statement is true.

> Do you think it is possible for the F# compiler to be changed so that
> 
> ```fsharp
> let defaultConfig = { Formatter = sprintf "%A" }
> ```
> 
> and
> 
> ```fsharp
> let defaultConfig<'a> = let f : 'a -> string = sprintf "%A" in { Formatter = f }
> ```
> 
> emit the same MSIL?

(I fixed two typos in the second code block compared to what I am quoting from https://github.com/hedgehogqa/fsharp-hedgehog/pull/353#discussion_r711622977)

I am not suggesting that a new type parameter be introduced when `defaultConfig` is accessed. I am suggesting that it is possible that F# to be changed/improved so it correctly infers the use of an already existing type parameter (instead of resolving that type parameter as `obj` like it does in its current behavior).

I think it should be possible, but I don't know for sure.  Maybe I am wrong about something here.

adam-becker: It can already infer the types, H&M type inference is pretty solid. F# just doesn't allow this because the CLR doesn't. The best you can do today as an F# programmer is make default config an inline function to remove the overhead of the call (but you then introduce/keep instantiation overhead).

In a perfect world F# could "memoize" copies of this property for each type argument its been given, and pretend the field is generic. Such that `defaultConfig<int32>` is only created once, as is `defaultConfig<string list>`, etc. While this is theoretically possible, I don't think it would ever happen without CLR support first.

Does that clear things up at all?

TysonMN: > Does that clear things up at all?

No, you are wrong.

> The best you can do today as an F# programmer is make default config an inline function to remove the overhead of the call [...]

This best one can do is what I gave in https://github.com/hedgehogqa/fsharp-hedgehog/pull/353#discussion_r711604381.

> F# just doesn't allow this because the CLR doesn't.

The code in the linked comment compiles and does what we want.  I don't care what CLI it generates.  F# can do things that the CLI doesn't support, such as statically resolved type parameters and units of measure.  Maybe this can't be accessed from C#.  That is a different issue.  I think we should first give F# the best possible API and then consider what can be done for C#.

PR #352, opened 2021-09-11
Remove 'Random.Builder'

adam-becker:
I removed the computation expression builder for `Random`. One function in `Gen` was using it, and that function has been rewritten to not need it.

PR #354, opened 2021-09-12
make Gen.{sampe, sampleTree} and Random.replicate lazy

### LINKED ISSUE #154: Generate lazily evaluated sequences

TysonMN:
Resolves #154

### DIFF
--- a/src/Hedgehog/Gen.fs
+++ b/src/Hedgehog/Gen.fs
@@ -1,4 +1,4 @@
-﻿namespace Hedgehog
+namespace Hedgehog
 
 open System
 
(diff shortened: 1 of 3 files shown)


adam-becker:
@TysonMN ready to merge?

TysonMN:
Yep.

Note, there will be a minor conflict between this PR and PR #352.

PR #357, opened 2021-09-12
remove Property.ofThrowing

TysonMN:
I don't think `Property.ofThrowing` needs to exist.  Its purpose was to convert an exception into a property with a failed outcome.  That conversion how happens in both `Property.map` and `Property.bind`, so I can't think of any reason why this function needs to exist.

In the process of removing this, I improved/simplified a `Property.SelectMany` method by replacing a call to `bind` with a call to `map`.


adam-becker:
:+1:

PR #358, opened 2021-09-14
Fixed import namespace in cshrp sample

lupin-de-mid:
fixed mispelled namespace


moodmosaic (approved):
Thank you, @lupin-de-mid 👍 ❤️

PR #361, opened 2021-09-20
Fix order of arguments to Expect.equal

TysonMN:
Here is a screenshot of the tooltip for `Expect.equal` that shows the branch in this PR has the arguments in the correct order.

2021-09-19_20-21-10_695_devenv


adam-becker:
👍🏼

PR #363, opened 2021-09-21
Utilize Property.set internally

TysonMN:


src/Hedgehog/Property.fs
```diff
@@ -72,7 +72,7 @@ module Property =
         let h = g |> Gen.map |> mapGen
         h x
 
-    let private set (a: 'a) (property : Property<'b>) : Property<'a> =
+    let internal set (a: 'a) (property : Property<'b>) : Property<'a> =
         property |> map (fun _ -> a)
```
adam-becker, line 76: I never caught this until looking at this diff, but should we be using `always` here?

adam-becker (approved):

src/Hedgehog/Property.fs
TysonMN: Can you elaborate?  What is `always`?

Issue #332, opened 2021-07-08
Way to start a recheck from a fully shrunk case?


altxt:
In some cases, it can be beneficial to inspect in a debugger the behavior of the function under test. If it is sufficiently complex, this is better done using the input data that is as simple as possible - i.e. shrunken. Debugging is often done by seting breakpoints, so it is reasonable to expect the breakpoints to be hit starting directly with the shrunk unput. But I was unable to find a way of making it so. The reproduction info that Hedgehog prints seems to lead to the failing case that was found first, not to the result of shrinking. Am I missing something?

Here is the repro in code. We check a function with clear failure conditions and we log both the input and the number of calls:

```cs
[Test]
public void Hedgehog_Recheck_StartsDirectlyFromShrunkenCase()
{
    int count = 0;
    // any input in [13 .. 20] range fails the test, with 13 being the simplest failing case
    Func<int, bool> testFunc = x =>
    {
        count++;
        Console.WriteLine($"x = {x}, count = {count}");
        return x < 13 || x > 20;
    };
    var property =
        from x in Property.ForAll(Gen.Int32(Range.Constant(1, 1000)))
        select testFunc(x);
    
    // for getting the seed
    //property.Check(PropertyConfig.Default.WithShrinks(100));

    // for the actual test
    property.ReportRecheck(41, new Seed(15041786197103353380UL, 17636625472901432099UL), PropertyConfig.Default.WithTests(1));
    Assert.AreEqual(1, count, "Function under test called more than once");
}
```

Initial run, getting the seed for recheck:
```
System.Exception : *** Failed! Falsifiable (after 41 tests and 1 shrink):
13
This failure can be reproduced by running:
> Property.recheck (41 : Size) ({ Value = 15041786197103353380UL; Gamma = 17636625472901432099UL }) <property>

   at Hedgehog.ReportModule.tryRaise(Report report)
   at HedgehogSimpleTests.Hedgehog_Recheck_StartsDirectlyFromShrunkenCase()

x = 900, count = 1
x = 897, count = 2
x = 503, count = 3
x = 513, count = 4
x = 557, count = 5
x = 440, count = 6
x = 197, count = 7
x = 264, count = 8
x = 308, count = 9
x = 121, count = 10
x = 665, count = 11
x = 453, count = 12
x = 225, count = 13
x = 610, count = 14
x = 116, count = 15
x = 983, count = 16
x = 739, count = 17
x = 388, count = 18
x = 989, count = 19
[...]

PR #364, opened 2021-09-21
Add BindReturn to Property CE

TysonMN:
This PR is makes progress on issue #332.  This PR should be merged before PR #336 (which is still only a draft).  After this PR is merged, I will force push an update to the branch in that PR that includes a rebase on top of the branch in this PR.

The main point of this PR is to add `BindReturn` to the `Property` computational expression.  As [the corresponding RFC says](https://github.com/fsharp/fslang-design/blob/main/FSharp-5.0/FS-1063-support-letbang-andbang-for-applicative-functors.md#:~:text=nodes-,BindReturn%20-%20adds,Bind2Return,-%2C)
> - `BindReturn` - adds support and/or efficiency for `let! ... return`

That is preciously what we need to avoid testing non-shrunken inputs during rechecking.

One "controversial" thing this PR does is remove the `Return(bool)` method from the `Property` CE.  I removed it because having both this method and `BindReturn` were giving me compiler errors.  Removing `Return(bool)` was the only solution that I found.  Maybe there is another one we could use instead.  However, `Return(bool)` is not necessary.  It is syntactic sugar that allows the user to omit putting
```fsharp
|> Property.failOnFalse
```
after their CE instance.  Therefore, I think it is reasonable to remove this.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -16,7 +16,7 @@ jobs:
     - name: Setup .NET Core
       uses: actions/setup-dotnet@v1
       with:
-        dotnet-version: 3.1.403
+        dotnet-version: 5.0.403
     - name: Install dependencies
       run: dotnet restore
     - name: Build
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -18,7 +18,7 @@ jobs:
     - name: Setup .NET Core
       uses: actions/setup-dotnet@v1
       with:
-        dotnet-version: 3.1.403
+        dotnet-version: 5.0.403
     - name: Install dependencies
       run: dotnet restore
     - name: Build
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## Version ?
 
 - Rename `Property.failOnFalse` to `Property.falseToFailure` ([#384][384], [@TysonMN][TysonMN])
+- Add `BindReturn` to the `property` CE ([#364][364], [@TysonMN][TysonMN])
+  - A breaking change.  Previously, returning a `bool` from a `property` CE (after using `let!`) caused the CE to have return type `Property<unit>`.  Now this results in a return type of `Property<bool>`.  The previous behavior can now be expressed by piping the `Property<bool>` instance into `Property.falseToFailure`.
 
 ## Version 0.11.1 (2021-11-19)
 
@@ -188,6 +190,8 @@
   https://github.com/hedgehogqa/fsharp-hedgehog/pull/381
 [380]:
   https://github.com/hedgehogqa/fsharp-hedgehog/pull/380
+[364]:
+  https://github.com/hedgehogqa/fsharp-hedgehog/pull/364
 [363]:
   https://github.com/hedgehogqa/fsharp-hedgehog/pull/363
 [362]:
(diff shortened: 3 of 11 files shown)


TysonMN:
> [This CE builder method was added "in F# 5".](https://docs.microsoft.com/en-us/dotnet/fsharp/whats-new/fsharp-50#applicative-computation-expressions) Therefore, I thought I would need to add compiler directives to avoid using this method if the version of `FSharp.Core` was less than 5.0. However, this doesn't seem to be a problem. Even if I pin the version of `FSharp.Core` to the lowest version supported by Hedgehog, I can still make use of this CE builder method. (I think that means it is a feature from the .NET 5 SDK that was released simultaneously with F# 5, but maybe I am wrong about this.) Anyway, the conclusion is that this PR doesn't add any compiler directives (about the version of `FSharp.Core`) because it doesn't seem necessary.

I think I was just being dumb.  After thinking through this again, I am now positive that F# 5 is part of the SDK.  So as I previously discovered and described, this PR has no impact on our on `FSharp.Core`.

adam-becker:
Before I merge this (since it removes `Return(bool)`) does this mean we are committing to doing #365 in `0.12.0`? I think the pill will be easier to swallow is there's a "new way" of doing things. I think the workaround you've mentioned would be asking too much of our users, so I just want to make sure we have a nice-ish upgrade path. Also, #365 has some unanswered questions, particularly how we would support `counterexample`. Once `Property` is no longer polymorphic, the `property` CE becomes more-or-less useless.

TysonMN:
> I think the workaround you've mentioned would be asking too much of our users, so I just want to make sure we have a nice-ish upgrade path.

This can be simplified by making it into a public function for users to consume.

> (since it removes `Return(bool)`) does this mean we are committing to doing #365 in `0.12.0`?

No.  I think that issue and this PR are unrelated...except that I already wrote this code and the code in PR #336 to work with a polymorphic `Property`, and I don't want to have to keep then conflict free with `master`.

> Also, #365 has some unanswered questions, particularly how we would support `counterexample`. Once `Property` is no longer polymorphic, the `property` CE becomes more-or-less useless.

Let's worry about that when focusing on #365.

adam-becker:
I think this PR is going to have to wait until we see what `Property` and its CE evolve into. My concern is that we go through with this and then end up making it harder to shape `Property` the way we want to.

If you feel this concern is unwarranted please set me straight, otherwise I'm going to focus on #365 first.

TysonMN:
#365 is just code clean up.  In contrast, this PR plus PR #336 adds a great feature for users.  I think we should prioritize adding value by focusing on this PR and PR #336 now and focusing on PR #365 later.

TysonMN:
I force pushed to add commit ce3ef0d44e2f3f8ce775c11b3cb68d8ee7651634 that replaces all existing occurrences of `bind ofBool` with `failOnFalse`.

TysonMN:
Separate from this PR, I merged in the code that adds and then uses `Property.failOnFalse`.

dharmaturtle (commented):
Ah, I'm a bit late, but I think `failOnFalse` should've been named `failureWhenFalse` for two reasons:

1. It's called `Failure` in Hedgehog.
2. `failwith` exists. My mind conflates `failOnFalse` with that, so seeing `property { ... } |> Property.failOnFalse |> Property.check` is slightly surprising. `fail` is (mostly) a verb, while `failure` is (mostly) a noun.

Oh well.

Did you intentionally _not_ use `checkBool` on your tests?

Otherwise LGTM!

TysonMN:
I agree that `failure` is better than `fail`.  Is `when` better than `on` though?  How about `Property. FailureOnFalse`?

Issue #322, opened 2021-02-24
Name all arguments and include all type annotations in public API?


TysonMN:
Do we have to name all arguments and include all type annotations in the public API?

For example, I would prefer to write
https://github.com/hedgehogqa/fsharp-hedgehog/blob/569ca6915e77b274ec685ee6b5c339fbfc128bec/src/Hedgehog/Property.fs#L66-L67

as
```fsharp
let map f =
    f |> Outcome.map |> GenTuple.mapSnd |> mapGen
```

If we decide to change this convention/style, know that we don't have to change the all the code at once.  It can happen slowly over time, especially as we are making changes in an area.

adam-becker:
The component design guidelines for F# explicitly mention not doing what you're suggesting here. All parameters should be named/typed. If not in the main source file, then in a separate `.fsi` file once the API stabilizes (which I think we should do as part of a `1.0.0` release).

The reason for this, is that the F# compiler won't generate the same code for point-free style functions. For example, this:

```fsharp
let add x y = x + y
```

Would compile to a method that looks like:

```csharp
int add(int x, int y) {
    return x + y;
}
```

On the other hand, this:

```fsharp
let add x = (+) x
```

Would compile to a method that looks like:

```csharp
FSharpFunc<int, int> add(int x) {
    return new FSharpFunc<int, int>(y => x + y);
}
```

This is not only ugly, it's also very inefficient. Now a simple add function has to allocate an `FSharpFunc`, and an object for its closure. In the event that the compiler _can_ turn the above into a 2 parameter method, the compiler generated names are utterly useless to consumers of the library (`_arg1`, `_arg2`, etc).

I think the code example you provided could just be written simply:

```fsharp
let map f x =
    Outcome.map f x
    |> GenTuple.mapSnd
    |> mapGen
```

This form is more readable, and also has the virtue of generating better code. My vote is to just write things simply, especially public-facing functions.

TysonMN:
> I think the code example you provided could just be written simply:
> 
> ```fsharp
> let map f x =
>     Outcome.map f x
>     |> GenTuple.mapSnd
>     |> mapGen
> ```
> 
> This form is more readable, and also has the virtue of generating better code. My vote is to just write things simply, especially public-facing functions.

What you are suggesting doesn't compile.

[...]

PR #368, opened 2021-09-21
Example of using a .fsi file for Tree.fs

adam-becker:
Specifically targeted at @TysonMN following a discussion about this in #322. I also took the opportunity to address #335.

This is a first pass, so I likely missed something.

### DIFF
--- a/src/Hedgehog/Hedgehog.fsproj
+++ b/src/Hedgehog/Hedgehog.fsproj
@@ -32,6 +32,7 @@ https://github.com/hedgehogqa/fsharp-hedgehog/blob/master/doc/index.md
     <Compile Include="Numeric.fs" />
     <Compile Include="Seed.fs" />
     <Compile Include="Seq.fs" />
+    <Compile Include="Tree.fsi" />
     <Compile Include="Tree.fs" />
     <Compile Include="OptionTree.fs" />
     <Compile Include="Range.fs" />
--- a/src/Hedgehog/Hedgehog.fsproj
+++ b/src/Hedgehog/Hedgehog.fsproj
@@ -23,6 +23,7 @@ https://github.com/hedgehogqa/fsharp-hedgehog/blob/master/doc/index.md
     <PackageIconUrl>https://github.com/hedgehogqa/fsharp-hedgehog/raw/master/img/hedgehog-logo.png</PackageIconUrl>
     <PackageTags>fsharp;f#;c#;csharp;dotnet;testing</PackageTags>
     <RepositoryUrl>https://github.com/hedgehogqa/fsharp-hedgehog</RepositoryUrl>
+    <OtherFlags>--warnon:3218</OtherFlags>
   </PropertyGroup>
   <ItemGroup>
     <AssemblyAttribute Include="System.Runtime.CompilerServices.InternalsVisibleToAttribute">
--- a/src/Hedgehog/Hedgehog.fsproj
+++ b/src/Hedgehog/Hedgehog.fsproj
@@ -23,6 +23,7 @@ https://github.com/hedgehogqa/fsharp-hedgehog/blob/master/doc/index.md
     <PackageIconUrl>https://github.com/hedgehogqa/fsharp-hedgehog/raw/master/img/hedgehog-logo.png</PackageIconUrl>
     <PackageTags>fsharp;f#;c#;csharp;dotnet;testing</PackageTags>
     <RepositoryUrl>https://github.com/hedgehogqa/fsharp-hedgehog</RepositoryUrl>
+    <!-- Warning for inconsistent parameter names between .fs and .fsi files -->
     <OtherFlags>--warnon:3218</OtherFlags>
   </PropertyGroup>
   <ItemGroup>
--- a/src/Hedgehog/Tree.fs
+++ b/src/Hedgehog/Tree.fs
@@ -5,25 +5,25 @@ type Tree<'a> =
 
 module Tree =
 
-    let root (Node (root, _)) =
+    let root tree =
+        let (Node (root, _)) = tree
         root
 
-    let children (Node (_, children)) =
+    let children tree =
+        let (Node (_, children)) = tree
         children
 
     let singleton value =
         Node (value, Seq.empty)
 
     let addChild child parent =
-        let (Node (x, xs)) = parent
-        Node (x, Seq.cons child xs)
+        Node (root parent, Seq.cons child (children parent))
 
-    let addChildValue value tree =
-        tree |> addChild (singleton value)
+    let addChildValue value parent =
+        parent |> addChild (singleton value)
 
     let rec cata f tree =
-        let (Node (x, xs)) = tree
-        f x (Seq.map (cata f) xs)
+        f (root tree) (Seq.map (cata f) (children tree))
 
     let depth tree =
         tree |> cata (fun _ -> Seq.fold max -1 >> (+) 1)
@@ -31,18 +31,18 @@ module Tree =
     let toSeq tree =
         tree |> cata (fun a -> Seq.join >> Seq.cons a)
 
-    let rec map f tree =
-        let (Node (x, xs)) = tree
-        Node (f x, Seq.map (map f) xs)
+    let rec map mapping tree =
+        let root = mapping (root tree)
+        let children = Seq.map (map mapping) (children tree)
+        Node (root, children)
 
-    let mapWithSubtrees f tree =
-        tree |> cata (fun a subtrees -> Node (f a subtrees, subtrees))
+    let mapWithSubtrees mapping tree =
+        tree |> cata (fun root subtrees -> Node (mapping root subtrees, subtrees))
 
-    let rec bind k tree =
-        match k (root tree) with
-        | Node (y, ys) ->
-            let xs = Seq.map (bind k) (children tree)
-            Node (y, Seq.append xs ys)
+    let rec bind mapping tree =
+        let newTree = mapping (root tree)
+        let newChildren = Seq.map (bind mapping) (children tree)
+        Node (root newTree, Seq.append newChildren (children newTree))
 
     let join trees =
         bind id trees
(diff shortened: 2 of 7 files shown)


TysonMN:
The parameter names are missing.  I think this makes the API worse, especially for functions like `addChild` that accept two parameters of the same type and is a non-commutative function.

Can the API include parameter names if an `.fsi` file is being used?

Specifically, `master` looks like

2021-09-21_20-56-07_697_devenv

while the branch in this PR looks like
2021-09-21_20-53-58_696_devenv

adam-becker:
> Can the API include parameter names if an `.fsi` file is being used?

Yep! Quoting the article about signature files:

> If parameter names in a signature file differ from the corresponding implementation file, the name in the signature file will be used instead, which may cause issues when debugging or profiling. If you wish to be notified of such mismatches, enable warning 3218 in your project file or when invoking the compiler (see --warnon under Compiler Options).

I can add these back in, but it's completely doable.

TysonMN:
I think we should put parameter names in the signature file and turn on that warning

adam-becker:
@TysonMN I've pushed a change to this branch that includes the parameter names and the warning.

src/Hedgehog/Hedgehog.fsproj
```diff
@@ -23,6 +23,7 @@ https://github.com/hedgehogqa/fsharp-hedgehog/blob/master/doc/index.md
     <PackageIconUrl>https://github.com/hedgehogqa/fsharp-hedgehog/raw/master/img/hedgehog-logo.png</PackageIconUrl>
     <PackageTags>fsharp;f#;c#;csharp;dotnet;testing</PackageTags>
     <RepositoryUrl>https://github.com/hedgehogqa/fsharp-hedgehog</RepositoryUrl>
+    <OtherFlags>--warnon:3218</OtherFlags>
```
TysonMN, line 26: Add a comment saying what this warning this is

src/Hedgehog/Tree.fs
```diff
@@ -1,120 +1,105 @@
...
+        Node (root parent, Seq.cons child (children parent))
 
-    let rec cata (f: 'a -> 'b seq -> 'b) (Node (x, xs): Tree<'a>) : 'b =
-        f x (Seq.map (cata f) xs)
+    let addChildValue value parent =
+        parent |> addChild (singleton value)
 
-    let depth (tree: Tree<'a>) : int =
+    let rec cata f tree =
+        f (root tree) (Seq.map (cata f) (children tree))
```
TysonMN, line 26: I think the pair that is a tree should still be explicitly decomposed.  Does the `.fsi` file prevent this?

src/Hedgehog/Tree.fs
```diff
@@ -1,120 +1,105 @@
...
-    ///
-    /// If you want to replace the shrinks altogether, try:
-    ///
-    /// Tree.unfold f (outcome oldTree)
-    ///
-    let rec expand (f : 'a -> seq<'a>) (Node (x, xs) : Tree<'a>) : Tree<'a> =
+    let rec map mapping tree =
+        let root = mapping (root tree)
+        let children = Seq.map (map mapping) (children tree)
+        Node (root, children)
```
TysonMN, line 37: Rename `mapping` back to `f` and decompose the tree in the input.

src/Hedgehog/Tree.fs
```diff
@@ -1,120 +1,105 @@
...
+
+    let mapWithSubtrees mapping tree =
```
TysonMN, line 39: Rename `mapping` back to `f`

src/Hedgehog/Tree.fs
```diff
@@ -1,120 +1,105 @@
...
+        tree |> cata (fun root subtrees -> Node (mapping root subtrees, subtrees))
+
+    let rec bind mapping tree =
+        let newTree = mapping (root tree)
+        let newChildren = Seq.map (bind mapping) (children tree)
+        Node (root newTree, Seq.append newChildren (children newTree))
```
TysonMN, line 45: Rename `mapping` to `f` and decompose the tree in the input.

Actually, I think there should be a way to express this as `cata`.  I will look into that.

src/Hedgehog/Tree.fs
```diff
@@ -1,120 +1,105 @@
...
+
+    let join trees =
```
TysonMN, line 47: `trees` is certainly better than `xss`, but I expect the type of `trees` would be something like `Tree<_> seq` or Tree<_> list`.  A better name would probably be `treeOfTrees`.

TysonMN:
> I still don't have good names for the parameters of `fold` and `foldForest` though.

They might not be used.  Make a pass and remove anything not in use.

src/Hedgehog/Tree.fs
TysonMN: `bind`/`join` is defined in terms of `cata` in [this post](https://blog.ploeh.dk/2019/06/10/tree-catamorphism)...in Haskell, so I need to get some help reading it.  That post also defines `apply` in terms of `cata`, which I will want soon when adding proper applicative support (instead of via `bind`/`join` as is the case now).

adam-becker:
> They might not be used. Make a pass and remove anything not in use.

This PR isn't about cleaning up the API, only documenting it. If there are functions/etc we think need to be removed they should be in their own PR.

src/Hedgehog/Tree.fs
adam-becker: Mapping is something of an F# idiom. `Seq.map`, `Seq.collect`, `Option.map`, etc all call these parameters `mapping`.

src/Hedgehog/Tree.fs
adam-becker: You can't destructure a parameter with the `.fsi` file and the warning we've enabled. In this case the compiler will call the parameter `_arg1` and we'll get the warning. Also, I'm not a big fan of continually destructuring this way because we already have functions to do this for us.

TysonMN:
I understand your desire to separate concerns, but because you didn't first (or refuse to also) delete unused functions, you are creating more work for yourself (especially with regard to _any_ time spent thinking of better names for a function that is unused).

TysonMN:
I think you misunderstand the primary purpose of those functions, which is to chain calls.  It is possible (but significantly more difficult...you have to use `Func.flatten`) to reference something twice.  In those cases, I think it is typically more readable to assign that to a local value.  As a bonus, it will also be more efficient.

Not deconstructing directly in the parameters isn't so bad though.  Just do it in the first line of code.

src/Hedgehog/Tree.fs
TysonMN: I think you misunderstand the primary purpose of those functions, which is to chain calls.  It is possible (but significantly more difficult...you have to use `Func.flatten`) to reference something twice.  In those cases, I think it is typically more readable to assign that to a local value.  As a bonus, it will also be more efficient.

Not deconstructing directly in the parameters isn't so bad though.  Just do it in the first line of code.

src/Hedgehog/Tree.fs
TysonMN: That is a good reason to call the parameter `mapping`, but I still think that `f` is a better name.

src/Hedgehog/Tree.fs
TysonMN: Wait, it isn't so bad.  It is "just"
```fsharp
let bind f =
  let g oldRoot newOldChildren =
    let Node (newRoot, newNewChildren) = f oldRoot
    Node (newRoot, Seq.concat newNewChildren newOldChildren)
  cata g
```

src/Hedgehog/Tree.fs
adam-becker: > I think you misunderstand the primary purpose of those functions, which is to chain calls. It is possible (but significantly more difficult...you have to use `Func.flatten`) to reference something twice. In those cases, I think it is typically more readable to assign that to a local value. As a bonus, it will also be more efficient.

I think you misunderstand my intentions continuously and it's starting to wear on me, we can talk offline if you'd like.

The purpose of using the calls is to make life easier if the `Tree` representation changes. I'm not in the business of just chaining calls together for the fun of it, or because I can. There is a practical reason to not destructure and I would very much appreciate if that were considered. There is also a middle ground here:

```fsharp
let rec cata f tree =
    let root = root tree
    let children = children tree
    let mapping = cata f
    f root (Seq.map mapping children)
```

I tend to optimize for readability first, and I think you do as well. We have different ideas of what that looks like, but I think we can be a bit more cordial in our approach.

adam-becker:
> I understand your desire to separate concerns, but because you didn't first (or refuse to also) delete unused functions, you are creating more work for yourself (especially with regard to _any_ time spent thinking of better names for a function that is unused).

These are also public API. Just because we aren't using them doesn't mean they aren't used. If we are going to do anything, we are going to mark them as `Obsolete` for the next version, then remove them. Either way, they will continue to be public API (with a better name) for at least the next version, so getting into an argument about this is moot.

src/Hedgehog/Tree.fs
TysonMN: > [...] it's starting to wear on me [...]
> [...] I think we can be a bit more cordial in our approach.

Yes, I am sorry. My language was much too harsh there.  I apologize.

> I think you misunderstand my intentions continuously [...]
>
> The purpose of using the calls is to make life easier if the Tree representation changes.  [...]  There is a practical reason to not destructure and I would very much appreciate if that were considered.

Indeed, I misunderstood your intention.  Based on your previous comments, I thought you wanted to restrict this PR to improving function names in `Tree.fs` and adding a `Tree.fsi` file.  I didn't consider that you also wanted to change the implementation to make it easier to change the tree representation.  I think that is a good improvement, I don't mind if such changes are also included in this PR.

TysonMN (approved):
I think this PR is ready to merge

TysonMN:
@adam-becker, do you think this PR is ready to merge?

TysonMN:
This PR adds an .fsi file, addresses #335, and partially encapsulates our implementation of `Tree<>`.  I would like these to be done in separate PRs.

PR #372, opened 2021-09-22
Update dev container and project files.

adam-becker:
Taken from #353. This removes the hard-coded UID/GID, fixes the startup command to also `dotnet restore`, and normalizes our test project files (`netcoreapp3.0` is going to be deprecated soon).


moodmosaic:
🚀

PR #378, opened 2021-10-13
Add test for concat order in Tree.bind

TysonMN:
There are two ways in which the old and new children in `Tree.bind` can be concatenated.  As [discussed here](https://well-typed.com/blog/2019/05/integrated-shrinking/#:~:text=Although%20this%20version%20of%20join%20still%20satisfies%20the%20monad%20laws%2C%20it%20is%20strictly%20worse.), one of those ways is better than the other.

There is another tests that sort of covers this, but it is very indirect and sometimes passes (when the order is wrong).  That tests is 1greedy traversal with a predicate yields the perfect minimal shrink`.

This PR adds a specific and deterministic test to ensure that we are using the better order.

### DIFF
--- a/src/Hedgehog/Tree.fs
+++ b/src/Hedgehog/Tree.fs
@@ -1,4 +1,4 @@
-﻿namespace Hedgehog
+namespace Hedgehog
 
 /// A rose tree which represents a random generated outcome, and all the ways
 /// in which it can be made smaller.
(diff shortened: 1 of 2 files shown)


TysonMN:
> Ah! I forgot that I can pass a list. Much better. I will make that change tomorrow.

Done

dharmaturtle:
`zip` and `equals` aren't internal. If they're going to be joining the public API, I believe they should get docs and explicit type annotations. And, if we're feeling aggressive, tests :)

TysonMN:
I understand your concern, but this PR adds a direct test of `bind`, which is more important than `zip` or `equals`.  For that reason, this code is an improvement over the code in `master` as is.

However, I could make them internal, which is still in scope for the test project.  Is that acceptable to you?

dharmaturtle:
Yep! That's why I phrased it that way 😺

PR #385, opened 2021-11-27
Change recheck API to accept recheck data as string

### LINKED ISSUE #383: Change recheck functions to accept Base64 encoding
Issue #332 is a feature request to only recheck the shrunken input.  To achieve this goal, PR #336 wants the recheck functions to also accept the path through the shrink tree to the shrunken input.  To make that change easier, we can modify the recheck functions to accept a Base64-encoded string of the recheck arguments.

[Here](https://stackoverflow.com/questions/11743160/how-do-i-encode-and-decode-a-base64-string) is how to encode/decode a Base64 string in C#.

This idea draws inspiration from [JetCheck](https://github.com/JetBrains/jetCheck), a property-based testing library in Java used by JetBrains on IntelliJ.

TysonMN:
Resolves #383

### DIFF
--- a/src/Hedgehog/Report.fs
+++ b/src/Hedgehog/Report.fs
@@ -123,17 +123,13 @@ module Report =
 
         | RecheckType.FSharp ->
             appendLinef sb "This failure can be reproduced by running:"
-            appendLinef sb "> Property.recheck %d ({ Value = %A; Gamma = %A }) <property>"
-                failure.RecheckData.Size
-                failure.RecheckData.Seed.Value
-                failure.RecheckData.Seed.Gamma
+            appendLinef sb "> Property.recheck \"%s\" <property>"
+                (RecheckData.serialize failure.RecheckData)
 
         | RecheckType.CSharp ->
             appendLinef sb "This failure can be reproduced by running:"
-            appendLinef sb "> property.Recheck(%d, new Seed { Value = %A; Gamma = %A })"
-                failure.RecheckData.Size
-                failure.RecheckData.Seed.Value
-                failure.RecheckData.Seed.Gamma
+            appendLinef sb "> property.Recheck(\"%s\")"
+                (RecheckData.serialize failure.RecheckData)
 
         sb.ToString().Trim() // Exclude extra newline.
 
--- a/src/Hedgehog/Report.fs
+++ b/src/Hedgehog/Report.fs
@@ -9,10 +9,16 @@ type RecheckType =
     | None
     | CSharp
     | FSharp
-
-type FailureData = {
+    
+    
+[<Struct>]
+type RecheckData = internal {
     Size : Size
     Seed : Seed
+}
+
+type FailureData = {
+    RecheckData : RecheckData
     Shrinks : int<shrinks>
     Journal : Journal
     RecheckType : RecheckType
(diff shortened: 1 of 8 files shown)


TysonMN:
@Sweeney1309, you could also review this

TysonMN:
The Fable build failed, probably because it doesn't have the Base64 functions I used.  Probably should just contact strings together for now.

src/Hedgehog/Report.fs
```diff
@@ -29,6 +35,45 @@ type Report = {
...
+            |> Convert.FromBase64String
+            |> Encoding.UTF8.GetString
+
+    let private separator = "\n"
+
+    let serialize data =
+        data.Size.ToString ()         + separator
+        + data.Seed.Value.ToString () + separator
+        + data.Seed.Gamma.ToString ()
+        |> Base64.encode
```
dharmaturtle, line 61:
```fsharp
[ string data.Size
  string data.Seed.Value
  string data.Seed.Gamma ]
|> String.concat separator
|> Base64.encode
```

Maybe?

src/Hedgehog/Report.fs
```diff
@@ -9,10 +9,16 @@ type RecheckType =
     | None
     | CSharp
     | FSharp
-
-type FailureData = {
+    
+    
+[<Struct>]
+type RecheckData = internal {
```
dharmaturtle, line 15: I'm not a big fan of the name RecheckData, from a consumer POV. Perhaps something like RecheckSeed? This doesn't need to be the name of the internal type... maybe just on the public API? e.g.

```fsharp
let recheck (recheckSeed: string) (p : Property<unit>) : unit =
```

Could perhaps make it even simpler:

```fsharp
let recheck (seed: string) (p : Property<unit>) : unit =
```

TysonMN: I like `recheckSeed`.  I think `seed` is too confusing.

TysonMN: But I think `recheckData` is better, even for the user.

src/Hedgehog/Report.fs
TysonMN: Sure.  Just include this change in a force push.

tests/Hedgehog.Tests/ReportTests.fs
```diff
@@ -0,0 +1,22 @@
+﻿module Hedgehog.Tests.ReportTests
+
+open Hedgehog
+
+open TestDsl
+
+
+let reportTests = testList "Report tests" [
+    testCase "Roundtrip RecheckData serilzation" <| fun () ->
```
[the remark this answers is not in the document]
TysonMN: Just include this change in a force push.

TysonMN:
I removed the use of Base64 (which means that branch name isn't great any more, but whatever), and now the Fable tests pass.

TysonMN:
And I change the separator to `_` so that failure message now looks like this
```
> Property.recheck "0_7219121334209301472_6351764504400104151" <property>
```

src/Hedgehog/Report.fs
```diff
@@ -29,6 +35,31 @@ type Report = {
...
+        try
+            let parts = s.Split([|separator|], StringSplitOptions.None)
+            let size = parts.[0] |> Int32.Parse
+            let seed =
+                { Value = parts.[1] |> UInt64.Parse
+                  Gamma = parts.[2] |> UInt64.Parse }
+            { Size = size
+              Seed = seed }
+        with e ->
+            raise (ArgumentException("Failed to deseralize RecheckData", e))
```
dharmaturtle, line 60: `deserialize`

dharmaturtle (approved):
One last typo, but otherwise LGTM!

src/Hedgehog/Report.fs
TysonMN: Done

PR #401, opened 2021-12-30
Fix Journal bug in Property.BindReturn

TysonMN:
This PR is stacked on top of PR #399, which is stacked on top of PR #398. Those PRs should be completed first.


dharmaturtle (approved):
I have thoughts about the name `printValue` but whatever lol

TysonMN:
Sure, but `printValue` is an existing and non-public function, so not a big deal.

PR #415, opened 2022-07-21
408 fix recheck

### LINKED ISSUE #408: `recheck` always fails
I have the following test:

```fsharp
    testCase "my test" <| fun () ->
        property {
            let! i = Gen.int32 (Range.linear 1 10)
            Expect.isTrue (i <= 5)
        }
        |> Property.check
```

It fails (as expected) with the following error:

```
    *** Failed! Falsifiable (after 60 tests):
    6
    Expecto.AssertException: Should be true. Actual value was false but had expected it to be true.
       at Expecto.Expect.isTrue@397.Invoke(String msg)
       at Expecto.Expect.isTrue(Boolean actual, String message)
       at Hedgehog.Tests.TestDsl.Expect.isTrue(Boolean value) in C:\Code\fsharp-hedgehog\tests\Hedgehog.Tests\TestDsl.fs:line 38
       at Hedgehog.Tests.PropertyTests.propertyTests@103-30.Invoke(Int32 _arg8) in C:\Code\fsharp-hedgehog\tests\Hedgehog.Tests\PropertyTests.fs:line 103
       at Hedgehog.Property.kTry@88.Invoke(a a) in C:\Code\fsharp-hedgehog\src\Hedgehog\Property.fs:line 89
    > Property.recheck "59_15683818216235017248_6890634259096173777_" <property>
```

I plug the suggested `recheck` into the test:

```fsharp
    testCase "my test" <| fun () ->
        property {
            let! i = Gen.int32 (Range.linear 1 10)
            Expect.isTrue (i <= 5)
        }
        |> Property.recheck "59_15683818216235017248_6890634259096173777_"
```

It fails with

```

    *** Failed! Falsifiable (after 1 test):
    6
    Expecto.AssertException: Should be true. Actual value was false but had expected it to be true.
       at Expecto.Expect.isTrue@397.Invoke(String msg)
       at Expecto.Expect.isTrue(Boolean actual, String message)
       at Hedgehog.Tests.TestDsl.Expect.isTrue(Boolean value) in C:\Code\fsharp-hedgehog\tests\Hedgehog.Tests\TestDsl.fs:line 38
       at Hedgehog.Tests.PropertyTests.propertyTests@103-30.Invoke(Int32 _arg8) in C:\Code\fsharp-hedgehog\tests\Hedgehog.Tests\PropertyTests.fs:line 103
       at Hedgehog.Property.kTry@88.Invoke(a a) in C:\Code\fsharp-hedgehog\src\Hedgehog\Property.fs:line 89
```

Great! Let's now make the test pass (trivially):

```fsharp
    testCase "my test" <| fun () ->
        property {
            let! i = Gen.int32 (Range.linear 1 10)
            Expect.isTrue true
        }
        |> Property.recheck "59_15683818216235017248_6890634259096173777_"
```

But it doesn't pass!

```
    *** Failed! Falsifiable (after 1 test):
    6
```

Is this intended behavior?

TysonMN:
Fixes #408

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Version TBD
+
+- Fix bug in `Property.recheck` where the result is always `Failed`. ([#415][415], [@TysonMN][TysonMN])
+
 ## Version 0.12.1 (2021-12-31)
 
 - Add `Tree.apply`. Change `Gen.apply` from monadic to applicative. Revert runtime optimization of `Gen.integral`. ([#398][398], [@TysonMN][TysonMN])
@@ -192,6 +196,8 @@
 [porges]:
   https://github.com/porges
 
+[415]:
+  https://github.com/hedgehogqa/fsharp-hedgehog/pull/415
 [401]:
   https://github.com/hedgehogqa/fsharp-hedgehog/pull/401
 [400]:
--- a/src/Hedgehog/Property.fs
+++ b/src/Hedgehog/Property.fs
@@ -163,11 +163,15 @@ module Property =
         let rec skipPassedChild children shrinkPath =
             match children, shrinkPath with
             | _, [] ->
-                Failed {
-                    Shrinks = 0<shrinks>
-                    Journal = root.Value |> fst
-                    RecheckInfo = None
-                }
+                let journal, outcome = root.Value
+                match outcome with
+                | Failure ->
+                    { Shrinks = 0<shrinks>
+                      Journal = journal
+                      RecheckInfo = None }
+                    |> Failed
+                | Success _ -> OK
+                | Discard -> failwith "Unexpected 'Discard' result when rechecking. This should never happen."
             | [], _ -> failwith "The shrink path lead to a dead end. This should never happen."
             | _ :: childrenTail, ShrinkOutcome.Pass :: shrinkPathTail -> skipPassedChild  childrenTail shrinkPathTail
             | childrenHead :: _, ShrinkOutcome.Fail :: shrinkPathTail -> followShrinkPath childrenHead shrinkPathTail
--- a/tests/Hedgehog.Tests/PropertyTests.fs
+++ b/tests/Hedgehog.Tests/PropertyTests.fs
@@ -97,6 +97,31 @@ let propertyTests = testList "Property tests" [
                 count =! 1
                 //render.Contains "actual: 1" =! true // comment out for now since it causes the Fable test to fail
 
+
+    testCase "recheck passes after code is fixed" <| fun () ->
+        let mutable b = false
+        let prop =
+            property {
+                let! _ = Gen.bool
+                Expect.isTrue b
+            }
+    
+        let report1 = Property.report prop
+        match report1.Status with
+        | OK -> failwith "Initial report should be Failed, not OK"
+        | GaveUp -> failwith "Initial report should be Failed, not GaveUp"
+        | Failed failure1 ->
+            b <- true // "fix code"
+            let report2 =
+                Property.reportRecheck
+                    (RecheckData.serialize failure1.RecheckInfo.Value.Data)
+                    prop
+            match report2.Status with
+            | OK -> ()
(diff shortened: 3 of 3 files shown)


TysonMN:
@AlexeyRaga, can you review?  In particular, can you test this with the code that lead you to write https://github.com/hedgehogqa/fsharp-hedgehog/issues/408#issuecomment-1191296726?

tests/Hedgehog.Tests/PropertyTests.fs
```diff
@@ -97,6 +97,36 @@ let propertyTests = testList "Property tests" [
                 count =! 1
                 //render.Contains "actual: 1" =! true // comment out for now since it causes the Fable test to fail
 
+
+    testCase "recheck passes after code is fixed" <| fun () ->
+        let expectedShrunkenInput = 1
+        let mutable target = 0
+        let prop =
+            property {
+                let! i = Range.constant 0 1_000_000 |> Gen.int32
```
AlexeyRaga, line 106: Since the range starts with `0`, can't it generate `i = 0` which would be `= target`?

AlexeyRaga:
@TysonMN looks good to me, thanks for fixing it!

TysonMN:
@AlexeyRaga, you keep deleting your comments ;)

I had a good response typed up for your most recent one, so here it is.

> Since the range starts with `0`, can't it generate `i = 0` which would be `= target`?

Yes, and that happens.  You can put a breakpoint there to confirm verify.

An integral generator always generates the "smallest" (in terms of shrinking) value first, which in this case is `0`.  But the property is checked 100 times (by default, which is what is being used here), so the second value generated is unlikely to be `0` again (there is only `1` chance in `1,000,000` of that).  This test will fail if `0` is generated every time.  The probably of that is `(1 / 1,000,000)^99`, so this test will never fail.

AlexeyRaga:
@TysonMN yes, I understand that. I initially just didn't understand the fact that you _wanted_ that test to fail when you run it first time :) But then I read further, understood what is going on and quickly deleted my comment hoping that you won't see my initial stupidity :)

AlexeyRaga:
Could you please coin a release once it is merged?

TysonMN:
Definitely.  I talked to @dharmaturtle, and he will review this tomorrow.

src/Hedgehog/Property.fs
```diff
@@ -163,11 +163,15 @@ module Property =
...
-                }
+                let journal, outcome = root.Value
+                match outcome with
+                | Failure ->
+                    { Shrinks = 0<shrinks>
+                      Journal = journal
+                      RecheckInfo = None }
+                    |> Failed
+                | Success _ -> OK
+                | Discard -> failwith "Unexpected 'Discard' result when rechecking"
```
dharmaturtle, line 174:
```suggestion
                | Discard -> failwith "Unexpected 'Discard' result when rechecking. This should never happen."
```

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## Version 0.12.3 (2022-07-21)
+
+- Fix bug in `Property.recheck` where the result is always failed.
```
dharmaturtle, line 3:
```suggestion
- Fix bug in `Property.recheck` where the result is always `Failed`. ([#408][408], [@TysonMN][TysonMN])
```

dharmaturtle (approved):
Passes in my use case as well. A few very minor suggestions, but otherwise LGTM!

As side commentary, I'm not fond of making the test dependent on the shrinking behavior, but I don't feel strongly about it. One possible solution is to steal from my example in #408:

```fsharp
        let mutable target = false
        property {
            let! _ = // some random unused gen
            Expect.isTrue target
        }
```

But again, I don't feel strongly about this.

TysonMN:
> I'm not fond of making the test dependent on the shrinking behavior, but I don't feel strongly about it. One possible solution is to steal from my example in #408:

Yes, must better.  I changed to your suggestion.

PR #421, opened 2022-10-09
Fix introductory examples

bmitc:
The example properties are of type `Property<bool>`, and thus the correct type annotation should be used as well as the correct render function. As they are now in the documentation (prior to this edit), neither of these example properties compile.


TysonMN:
Thanks!

bmitc:
No problem! Thanks for the merge!

PR #433, opened 2023-07-12
Speed up recheck

### LINKED ISSUE #432: Recheck slow when using choice combined with list
While trying to create a toy example of using Hedgehog to generate test cases to exercise a FSM, in this case a very [simple web journey]( https://github.com/JohnEffo/fsharp-hedgehog-xunit/blob/choice-example/examples/Hedgehog.Xunit.Examples.CSharp/WebJourney/GenRandomJourney.cs) I noticed that running a re-check with the provided seed was slow (around 6-10 seconds) the code below simulates the problem but the wait is more like a couple of seconds:

```CSharp
using System.Collections.ObjectModel;
using Gen = Hedgehog.Linq.Gen;
using Range = Hedgehog.Linq.Range;
using Linq;
public class ChoiceRecheck
{
    [Fact]
    public void Choice()
    {
        var low = Gen.Int32(Range.Constant(0, 5));
        var mid = Gen.Int32(Range.Constant(10, 50));
        var big = Gen.Int32(Range.Constant(100, 200));
        var large = Gen.Int32(Range.Constant(500, 1000));
        var chioce = Gen.Choice(new Collection<Gen<int>> { low, mid, big, large }).List(Range.Constant(100,200));

        var prop = Property.ForAll(chioce).Select(x => x.Any(Test ));
        //prop.Check();
        prop.Recheck("0_8474167946941752373_7006309460388453401_000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000");

    }

    public bool Test(
        int x) => x == 990;
}
```

Initial posted in #419 as problem seemed similar.
> **Note**: In the original linked problem , the generator originally had a constant range (as in the example above) on the list with a range around 100..400. With this setting the recheck was even slower. Since moving to an Linear range the speed is better. The difference is that now the list can shrink down to a much smaller size, where as before the minimum size was length 100, so I think that the size of the shrunk result may have an impact on the speed.

TysonMN:
Resolves #432

Review happened in https://github.com/hedgehogqa/fsharp-hedgehog/issues/432#issuecomment-1629565074

### DIFF
--- a/src/Hedgehog/Linq/Property.fs
+++ b/src/Hedgehog/Linq/Property.fs
@@ -134,6 +134,10 @@ type PropertyExtensions private () =
         property |> Property.falseToFailure |> Property.reportRecheck recheckData
 
     [<Extension>]
+    static member ReportRecheck (property : Property<bool>, recheckData: RecheckData) : Report =
+        property |> Property.falseToFailure |> Property.reportRecheck (RecheckData.serialize recheckData)
+
+    [<Extension>]
     static member ReportRecheck (property : Property<bool>, recheckData: string, config : Hedgehog.PropertyConfig) : Report =
         property |> Property.falseToFailure |> Property.reportRecheckWith recheckData config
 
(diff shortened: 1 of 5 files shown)


dharmaturtle:
Turns out master is now broken. [Fixed here.](https://github.com/hedgehogqa/fsharp-hedgehog/pull/435)

PR #435, opened 2023-07-12
up deps

dharmaturtle:
update dependencies and other miscellanea


TysonMN:
I trust you on this.  Merge when ready.

Drop the first two commits (`fable sanity check` and its revert)?

PR #441, opened 2023-07-25
Fix decimal inline conversion

### LINKED ISSUE #439: Range.linearBounded crashes on decimals
Repro code:

```fsharp
#r "nuget: Hedgehog, 0.13.0"

open Hedgehog

let report =
  property {
    // Works
    // let! x = Gen.decimal (Range.constant 0.0m 100.0m)

    // Crashes
    // System.OverflowException: Value was either too large or too small for an Int32.
    let! x = Gen.decimal (Range.linearBounded ())

    return x = x
  }
  |> Property.reportBool

Report.render report
|> printfn "%s"
```

njlr:
Closes #439

### DIFF
--- a/src/Hedgehog/Numeric.fs
+++ b/src/Hedgehog/Numeric.fs
@@ -207,7 +207,7 @@ type FromBigInt =
         double
 
     static member FromBigInt (_: decimal, _: FromBigInt) : bigint -> decimal =
-        int32 >> decimal
+        decimal
 
 #if !FABLE_COMPILER
     static member FromBigInt (_: nativeint , _: FromBigInt) : bigint -> nativeint =
--- a/tests/Hedgehog.Tests/RangeTests.fs
+++ b/tests/Hedgehog.Tests/RangeTests.fs
@@ -219,4 +219,13 @@ let rangeTests = testList "Range tests" [
             (Range.exponentialBounded () : Range<sbyte>)
             |> Range.bounds 99
         actual =! (-128y, 127y)
+
+    testCase "exponentialBounded uses the full range of a data type - example 4" <| fun _ ->
+        let actual =
+            (Range.exponentialBounded () : Range<decimal>)
+            |> Range.bounds 99
+// TODO: should also work for Fable
+#if !FABLE_COMPILER
+        actual =! (Decimal.MinValue, Decimal.MaxValue)
+#endif
 ]


TysonMN:
I removed the assert from Fable.

@ThisFunctionalTom, are you still interested in maintaining Fable support?

ThisFunctionalTom:
> I removed the assert from Fable.
> 
> @ThisFunctionalTom, are you still interested in maintaining Fable support?

I can look into it. Is there a problem with Fable?
I could also update the compiler to the newest version.
I'll try updating, compiling and running tests and see what comes out.

TysonMN:
This is the build that I saw for the original version of this PR:
https://github.com/hedgehogqa/fsharp-hedgehog/actions/runs/5660360646/job/15460970572

The diff was very confusing at first.  Eventually I realized that there is a new line between each digit.  So instead of a number like
```
1234
```

the diff is comparing
```
1
2
3
4
```

The expected values are the min and max values for decimal.

---------------------------------

As I just worked on writing the above, now I see that the Fable build is also broken for `master`.  Here is an example:
https://github.com/hedgehogqa/fsharp-hedgehog/actions/runs/5705984309/job/15461213961

TysonMN:
> This is the build that I saw for the original version of this PR: https://github.com/hedgehogqa/fsharp-hedgehog/actions/runs/5660360646/job/15460970572
> 
> The diff was very confusing at first. Eventually I realized that there is a new line between each digit. So instead of a number like
> 
> ```
> 1234
> ```
> 
> the diff is comparing
> 
> ```
> 1
> 2
> 3
> 4
> ```
> 
> The expected values are the min and max values for decimal.
> 
> As I just worked on writing the above, now I see that the Fable build is also broken for `master`. Here is an example: https://github.com/hedgehogqa/fsharp-hedgehog/actions/runs/5705984309/job/15461213961

I just pushed a fix for that

TysonMN:
@ThisFunctionalTom, thanks for PR #444.

Can this test assertion I excluded from Fable be included now?

If so, can you make a PR for that?

ThisFunctionalTom:
> @ThisFunctionalTom, thanks for PR #444.
> 
> Can this test assertion I excluded from Fable be included now?
> 
> If so, can you make a PR for that?

Sadly it still does not work.

I think I found a bug in fable compiler or Fable.Core libraries. Following test does not work:

```fsharp
    testCase "Fable Decimal/BigInt bug" <| fun _ ->
        let actual = Decimal.MinValue |> bigint |> decimal
        actual =! Decimal.MinValue
```

I am going to see if I can fix it myself or, if not, I will open an issue.

PR #442, opened 2023-07-25
Add function Gen.timeSpan

### LINKED ISSUE #440: Add a Gen.timeSpan function?
Could be this?

```fsharp
  let timeSpan (range : Range<TimeSpan>) : Gen<TimeSpan> =
    Gen.int64 (range |> Range.map (fun x -> x.Ticks))
    |> Gen.map TimeSpan.FromTicks
```

Would a PR be accepted?

njlr:
Closes #440


TysonMN:
Fable doesn't support the property `TimeSpan.Ticks`, so I removed all this code from Fable

PR #446, opened 2023-08-06
Anchor works in the website too

arialdomartini:
In the page https://hedgehogqa.github.io/fsharp-hedgehog/ the links in the table of content do not work, because of mismatches in the case (e.g. `getting-started` instead of `Getting-started`).

This PR fixes this.

I also noticed there was a duplicated link, which I removed.


TysonMN:
Thanks! :)

PR #447, opened 2023-08-10
Linq xml comments

JohnEffo:
Add XML documentation to Hedgehog.Linq so that C# consumers have a better experience.


dharmaturtle (approved):
lgtm!

PR #452, opened 2025-04-11
Specify the YoloDev.Expecto.TestSdk package version to 0.14.3

jwChung:
Starting from version 0.15.*, the YoloDev.Expecto.TestSdkpackage targets .NET 8. Since the test project is using .NET 6, it fails to build with newer versions. To ensure compatibility and successful builds, the package version is specified to 0.14.3, which supports .NET 6.


moodmosaic:
Thank you, @jwChung 👍

PR #456, opened 2025-11-22
Perf optimisation

AlexeyRaga:
An easy way to make gen for most of the cases `2.5x` faster.
We are `~20x` slower than FSCheck, and after this optimisation we are "just" `7.5x` slower.

Before:
```
| Method                 | N      | Mean         | Error      | StdDev     | Ratio | RatioSD |
|----------------------- |------- |-------------:|-----------:|-----------:|------:|--------:|
| HedgehogGenSampleInt32 | 1000   |    404.56 us |   5.059 us |   4.732 us | 22.75 |    0.27 |
| FsCheckGenSampleInt32  | 1000   |     17.79 us |   0.090 us |   0.075 us |  1.00 |    0.01 |
|                        |        |              |            |            |       |         |
| HedgehogGenSampleInt32 | 10000  |  4,016.28 us |  17.300 us |  15.336 us | 22.83 |    0.14 |
| FsCheckGenSampleInt32  | 10000  |    175.90 us |   1.058 us |   0.884 us |  1.00 |    0.01 |
|                        |        |              |            |            |       |         |
| HedgehogGenSampleInt32 | 100000 | 40,881.16 us | 186.056 us | 174.037 us | 21.15 |    0.30 |
| FsCheckGenSampleInt32  | 100000 |  1,932.94 us |  30.679 us |  27.196 us |  1.00 |    0.02 |
```

After:
```
| Method                 | N      | Mean         | Error      | StdDev     | Ratio | RatioSD |
|----------------------- |------- |-------------:|-----------:|-----------:|------:|--------:|
| HedgehogGenSampleInt32 | 1000   |    152.88 us |   1.669 us |   1.393 us |  8.35 |    0.11 |
| FsCheckGenSampleInt32  | 1000   |     18.32 us |   0.215 us |   0.191 us |  1.00 |    0.01 |
|                        |        |              |            |            |       |         |
| HedgehogGenSampleInt32 | 10000  |  1,519.81 us |  26.226 us |  24.532 us |  8.48 |    0.17 |
| FsCheckGenSampleInt32  | 10000  |    179.26 us |   3.047 us |   2.544 us |  1.00 |    0.02 |
|                        |        |              |            |            |       |         |
| HedgehogGenSampleInt32 | 100000 | 15,473.26 us | 240.543 us | 277.010 us |  7.54 |    0.14 |
| FsCheckGenSampleInt32  | 100000 |  2,051.74 us |  11.659 us |  10.905 us |  1.00 |    0.01 |
```

The rest seems to be an overhead of Hedgehog's integrated shrinking, which is accepted by design.

### DIFF
--- a/tests/Hedgehog.Benchmarks/Program.fs
+++ b/tests/Hedgehog.Benchmarks/Program.fs
@@ -49,7 +49,7 @@ type ScaledBenchmarks () =
 
 [<EntryPoint>]
 let main argv =
-    // BenchmarkRunner.Run<Benchmarks> () |> ignore
-    // BenchmarkRunner.Run<ScaledBenchmarks> () |> ignore
+    BenchmarkRunner.Run<Benchmarks> () |> ignore
+    BenchmarkRunner.Run<ScaledBenchmarks> () |> ignore
     BenchmarkRunner.Run<GenBenchmarks>() |> ignore
-    0 // return an integer exit code
+    0


src/Hedgehog/Seed.fs
```diff
@@ -153,6 +153,47 @@ module Seed =
...
+                        loop seed1
+                loop seed
+
+    /// Generates a random int64 in the specified range.
+    let rec nextInt64 (lo : int64) (hi : int64) (seed : Seed) : int64 * Seed =
+        if lo > hi then
+            nextInt64 hi lo seed
+        else
+            let range = uint64 (hi - lo)
+            // If range is UInt64.MaxValue, we can just use the full 64 bits from nextUInt64
```
moodmosaic:
```suggestion
            // If range is UInt64.MaxValue, we can just use the full 64 bits from nextUInt64.
```

tests/Hedgehog.Benchmarks/Program.fs
```diff
@@ -49,7 +49,7 @@ type ScaledBenchmarks () =
 
 [<EntryPoint>]
 let main argv =
-    BenchmarkRunner.Run<Benchmarks> () |> ignore
-    BenchmarkRunner.Run<ScaledBenchmarks> () |> ignore
+    // BenchmarkRunner.Run<Benchmarks> () |> ignore
```
moodmosaic:
```suggestion
```

tests/Hedgehog.Benchmarks/Program.fs
```diff
@@ -49,7 +49,7 @@ type ScaledBenchmarks () =
...
+    // BenchmarkRunner.Run<ScaledBenchmarks> () |> ignore
```
moodmosaic:
```suggestion
```

moodmosaic (commented):
Thank you! The core optimization strategy is sound. I spotted a few things in the implementation that if true should result in even better perf.

src/Hedgehog/Seq.fs
```diff
@@ -8,13 +8,16 @@ let inline cons (x : 'a) (xs : seq<'a>) : seq<'a> =
...
-        if x = y then
-            ys0
-        else
-            cons x ys0
+    seq {
+        match Seq.tryHead ys0 with
+        | None -> yield x
+        | Some y ->
+            if x = y then
+                yield! ys0
```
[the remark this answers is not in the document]
AlexeyRaga: I benchmarked both against each other, and they are the same. Undone to the "old" one, no reasons to change.

src/Hedgehog/Random.fs
```diff
@@ -75,8 +75,15 @@ module Random =
     let inline integral (range : Range<'a>) : Random<'a> =
         Random (fun seed size ->
             let (lo, hi) = Range.bounds size range
-            let x, _ = Seed.nextBigInt (toBigInt lo) (toBigInt hi) seed
-            fromBigInt x)
+            if typeof<'a> = typeof<int32> then
+                let x, _ = Seed.nextInt32 (unbox (box lo)) (unbox (box hi)) seed
+                unbox (box x)
```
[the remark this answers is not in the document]
AlexeyRaga: I tried SRTP, but haven't noticed difference, and it was a bit harder to account for a "default" case...
We can try again later, as an iteration.

PR #470, opened 2025-12-04
Tagged journal

AlexeyRaga:
### What

Separate concerns of writing a journal and formatting the report.

This also simplifies report building (and unifies between hedgehog and hedgehog.xunit), and simplifies exceptions handling for repors.

### How

We do it by allowing "tagged" journal entries, so that:

- adding to the journal is not concerned about transforming values to strings
- report formatter has a bit more information about the intent and the meaning of each of the report entries

### The essence of the change

Instead of 

```fsharp
[<Struct>]
type Journal =
    | Journal of seq<unit -> string>
```

We now have

```fsharp
type JournalLine =
    | TestParameter of name: string * value: obj     // Individual test method parameter
    | GeneratedValue of value: obj                   // forAll generated values (no name)
    | Counterexample of message: string              // Property.counterexample user messages
    | Exception of exn: exn                          // Original exception, unwrap at render
    | Cancellation of message: string                // OperationCanceledException messages
    | Text of message: string                        // Plane text messages (info, etc.)

[<Struct>]
type Journal =
    | Journal of seq<unit -> JournalLine>
```

Now the code that writes the journal is not concerned about casting to strings, or curating exceptions, etc.
Report generator can do it.

### Generated report example

Even if I have a property definition that combines both xUnit and explicit techniques,  I can have a good report:

```csharp
// we are using Property attribuye and provided test method parameters
[Property]
public Property<bool> Foo(List<byte> myBytes, byte myInt) =>
    // and we also use explicit hedgehog-style generators in the same property
    from x in Gen.Alpha.String(Range.Constant(1, 20)).ForAll()
    from y in Gen.Bool.ForAll()
    from _ in Property.CounterExample(() => $"{x} is longer than {myInt}")
    select TestFail(myBytes.Count);
```

```
Hedgehog.Xunit.PropertyFailedException
  *** Failed! Falsifiable (after 1 test and 6 shrinks):

You can reproduce this failure with the following Recheck Seed:
  "0_691509993339586027_438402981521466921_0-0-0-0-0-0"

Test parameters:
  myBytes = []
  myInt = 0uy

Generated values:
  "a"
  false

Counterexamples:
  a is longer than 0

Actual exception:
System.InvalidOperationException: I fail with 0
   at Hedgehog.Xunit.Tests.CSharp.Async.TestFail(Int32 x) in /Users/alexey/src/github/fsharp-hedgehog/tests/Hedgehog.Xunit.Tests.CSharp/Async.cs:line 58
   at Hedgehog.Xunit.Tests.CSharp.Async.<>c__DisplayClass7_0.<Foo>b__3(<>f__AnonymousType0`2 <>h__TransparentIdentifier0, Unit _) in /Users/alexey/src/github/fsharp-hedgehog/tests/Hedgehog.Xunit.Tests.CSharp/Async.cs:line 55
   at <StartupCode$Hedgehog>.$Property___2.SelectMany@424-5.Invoke(TCollection b) in /Users/alexey/src/github/fsharp-hedgehog/src/Hedgehog/Linq/Property.fs:line 424
   at Hedgehog.Outcome.Map@28-3.Invoke(a x)
   at Hedgehog.Outcome.cata[a,b](Outcome`1 outcome, FSharpFunc`2 failure, FSharpFunc`2 discard, FSharpFunc`2 success) in /Users/alexey/src/github/fsharp-hedgehog/src/Hedgehog/Outcome.fs:line 21
   at Hedgehog.Outcome.Map[a,b](FSharpFunc`2 f, Outcome`1 result) in /Users/alexey/src/github/fsharp-hedgehog/src/Hedgehog/Outcome.fs:line 25
   at Hedgehog.FSharp.Property.applyWithExceptionHandling@171.Invoke(Tuple`2 tupledArg)

<xUnit stack trace hidden>
```

### DIFF
--- a/src/Hedgehog.Xunit/InternalLogic.fs
+++ b/src/Hedgehog.Xunit/InternalLogic.fs
@@ -213,12 +213,11 @@ module private PropertyBuilder =
                 | e -> box e
 
         let createJournal args =
-            let parameterEntries =
-                Array.zip parameters (List.toArray args)
-                |> Array.map (fun (param, value) -> 
-                    fun () -> TestParameter (param.Name, value))
-                |> Array.toSeq
-            Journal.ofSeq parameterEntries
+            args
+            |> Seq.zip parameters
+            |> Seq.map (fun (param, value) -> fun () -> TestParameter (param.Name, value))
+            |> Array.ofSeq // not sure if journal will do multiple enumerations
+            |> Journal.ofSeq
         
         let wrapWithExceptionHandling (result: obj) : Property<unit> =
             match result with


src/Hedgehog.Xunit/InternalLogic.fs
```diff
@@ -219,16 +189,8 @@ module private PropertyBuilder =
             else
                 testMethod
 
-        try
-            methodToInvoke.Invoke(testClassInstance, args |> Array.ofList)
-        with
-        | :? TargetInvocationException as tie when not (isNull tie.InnerException) ->
```
moodmosaic: This was moved to the call site in `InternalLogic.fs`. Is this intentional? The unwrapping is no longer recursive here. If a `TargetInvocationException` wraps another `TargetInvocationException`, it won't fully unwrap anymore.

moodmosaic (commented):
This is strictly better UX. Thank you! Since the Journal type signature changed it should require a major version bump.

PR #474, opened 2025-12-24
Simplify vars and envs

AlexeyRaga:
## Simplifying `Var`s and `Env`s

Instead of storing values in `Env`, `Var`s now store concrete values:

```fsharp
type Var<'T> =
    internal
    | Symbolic of 'T option
    | Concrete of 'T
```

This allows simplifying `Var` itself, but also removes the need of reflective state traversal for reporting the state.

### Justification

In .NET we don't have HKT, so we don't `HTraverse` between symbolic and concrete state anyway. Instead, state always holds `Var<'T>`, so there is no need in runtime correlation between `Env` and `Var`.

This means that during execution stage we can simply create a concrete `Var` (with value) and pass it to `Update`.
`Update` will still not be able to distinguish between symbolic and concrete variables, as expected.

`Env` is still being received by execution-stage methods (`Require`, `Execute`, `Ensure`) and is still required by `Var.Resolve` because it serves the role of the _capability_ and provides some compile-time safety: if you don't have `Env` you can't have access to concrete state.

### Results

With this approach we:

- Have much simple `Var` (win for maintenance)
- Can print state naturally (win for devex debugging)
- Don't need heavy reflection traversal of state in order to print in into the report (win from any angle)
- Have less code lines :)

### Report example

image


AlexeyRaga:
@moodmosaic kind of have it in tests: https://github.com/hedgehogqa/fsharp-hedgehog/blob/master/tests/Hedgehog.Stateful.Tests.FSharp/SequentialCounterSpec.fs ;)

PR #476, opened 2026-01-05
Fix preconditions check

AlexeyRaga:
`Precondition` and `Require` wasn't properly checked while shrinking

### DIFF
--- a/src/Hedgehog.Stateful/Sequential.fs
+++ b/src/Hedgehog.Stateful/Sequential.fs
@@ -390,7 +390,4 @@ module Sequential =
                 return (combinedJournal, Failure)
         }
 
-        // Create property from async computation that returns (Journal, Outcome)
-        // The lazy ensures execution only happens when forced (during followPath in recheck)
-        Gen.constant (lazy (Async.RunSynchronously executionAsync))
-        |> Property.ofGen
+        Property.ofAsyncWithJournal executionAsync
--- a/src/Hedgehog/Property.fs
+++ b/src/Hedgehog/Property.fs
@@ -118,7 +118,12 @@ module Property =
     /// The property succeeds when the async computation completes successfully, and fails if it throws an exception.
     /// This enables testing of asynchronous F# code.
     let ofAsync (asyncComputation : Async<'T>) : Property<'T> =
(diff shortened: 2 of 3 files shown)


moodmosaic (commented):
LGTM! Left one comment/question.

moodmosaic (commented):
I really appreciate all the thought you're putting into this. Let me share some perspective that might be helpful.

I haven't done large-scale, consensus-critical, model-based testing with HH at the scale I've used fast-check's approach. I've used HH in projects, but not at that level. If you can point me to a codebase using HH's model-based testing at large scale, I'd love to see it and might be able to tell more. If you've personally used HH's model-based testing at large scale, that's more than enough validation! 

But if you haven't (and are porting it here to use in .NET based on the design), I'd like to ask us to take a collective step back. I'm concerned we might be adding unnecessary complexity for no proven reason.

fast-check's approach is pragmatic and simple. fsharp-hedgehog doesn't necessarily need to replicate HH's approach 1:1 — we can take the best ideas from multiple ecosystems.

I've been using fast-check against PoX-4 (Proof of Transfer v4, the Stacks blockchain's consensus mechanism) and SIP-031 (Stacks Improvement Proposal for signer key rotation) when I was at Stacks Foundation. These are consensus-critical, production blockchain contracts. The efforts can be seen here:

- SIP-031 initial set of commands: https://github.com/stacks-network/stacks-core/pull/6321/changes#diff-adb5e0d70e3ba2936b90a93235b5d012948d51c1895868b418cb026277b95fb3
- All PoX-4 commands: https://github.com/stacks-network/stacks-core/tree/e6c240e0c0dc763b1cff8fd5fca7b4c237da8bdb/contrib/boot-contracts-stateful-prop-tests/tests/pox-4

I am not proposing or favoring JS/TS or fast-check over Haskell/Hedgehog/.NET/x/y/z. All I am saying is I'd rather port something I know works in practice and at scale, rather than porting something because "that's how it's done there in HH".

The two-stage generation/execution model with symbolic/concrete distinction adds real complexity. What concrete benefits does it provide that justify that complexity? Especially if we don't have evidence of it being battle-tested at scale?

PR #477, opened 2026-01-08
Remove Hedgehog.Stateful

AlexeyRaga:
## Description

Hedgehog to implement fast-check approach later at some point as Hedgehog Haskell approach appears to be complex.


moodmosaic (commented):
LGTM! The added combinators are fine to be part of this PR yes?

AlexeyRaga:
Yes, combinators standardise name, and will be useful for me in Stateful.

PR #478, opened 2026-01-12
Add Hedgehog.NUnit and tests

AlexeyRaga:
Introduce Hedgehog.NUnit

### DIFF
--- a/src/Hedgehog.NUnit/PropertyAttribute.fs
+++ b/src/Hedgehog.NUnit/PropertyAttribute.fs
@@ -69,12 +69,17 @@ and HedgehogTestMethod(mi: IMethodInfo, parentSuite: Test) =
         TestExecutionContext.CurrentContext.CurrentResult <- testResult
 
         try
+            use _ = new TestExecutionContext.IsolatedContext()
+
+            // Run SetUp methods
             try
-                use _ = new TestExecutionContext.IsolatedContext()
                 x.RunSetUp()
-                x.RunPropertyTest(context, testResult)
             with ex ->
                 x.HandleException(ex, testResult, FailureSite.SetUp)
+
+            // Only run the test if SetUp succeeded
+            if testResult.ResultState = ResultState.Inconclusive then
+                x.RunPropertyTest(context, testResult)
         finally
             x.RunTearDown(testResult)
 


src/Hedgehog.NUnit/GenAttribute.Prelude.fs
```diff
@@ -0,0 +1,176 @@
...
+[<AutoOpen>]
+module private RangeHelpers =
+    /// Choose between constant, linear, and exponential range based on the range size.
+    /// - For ranges > 1000: use exponential to ensure boundary values are tested
+    /// - For ranges > 100: use linear for balanced shrinking
+    /// - For ranges <= 100: use constant (no shrinking needed for small ranges)
+    let inline chooseRangeInt32 (min: int) (max: int) : Range<int> =
+        let rangeSize = int64 max - int64 min
+        let origin = if min <= 0 && 0 <= max then 0 else min
+        match rangeSize with
```
moodmosaic: Perhaps these ranges can be named constants? E.g.

```fs
[<Literal>]
let private LargeRangeThreshold = 1000L
[<Literal>]
let private MediumRangeThreshold = 100L

let inline chooseRangeInt32 (min: int) (max: int) : Range<int> =
    let rangeSize = int64 max - int64 min
    let origin = if min <= 0 && 0 <= max then 0 else min
    match rangeSize with
    | size when size > LargeRangeThreshold -> Range.exponentialFrom origin min max
    | size when size > MediumRangeThreshold -> Range.linearFrom origin min max
    | _ -> Range.constantFrom origin min max
```

src/Hedgehog.NUnit/InternalLogic.fs
```diff
@@ -0,0 +1,256 @@
...
+
+let withTests =
+    function
+    | Some x -> PropertyConfig.withTests x
+    | None -> id
+
+let withShrinks =
+    function
+    | Some x -> PropertyConfig.withShrinks x
+    | None -> PropertyConfig.withoutShrinks
```
moodmosaic:
```suggestion
    | None -> id
```

Shouldn't we leave default shrinking as configured by PropertyConfig.defaults unless explicitly overridden?

src/Hedgehog.NUnit/AutoGenConfig.fs
```diff
@@ -0,0 +1,34 @@
+namespace Hedgehog.NUnit
+
+open System
+open Hedgehog
+
+module internal AutoGenConfig =
+    let instantiate (configType: Type) (configArgs: obj array) =
```
moodmosaic: This searches methods (not properties) and expects “exactly one” public static member returning `IAutoGenConfig`. The error string mentions “property” but the code looks for methods, IIUC. Are we sure this works?

src/Hedgehog.NUnit/PropertyContext.fs
```diff
@@ -0,0 +1,56 @@
...
+                |> AutoGenConfig.merge ctx.AutoGenConfig
+            | None -> ctx.AutoGenConfig
+
+        { ctx with
+            AutoGenConfig = config
+            Tests = attr.Tests |> Option.orElse ctx.Tests
+            Shrinks = attr.Shrinks |> Option.orElse ctx.Shrinks
+            Size = attr.Size |> Option.orElse ctx.Size }
+
+    let fromMethod (method: MethodInfo) =
```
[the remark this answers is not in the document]
AlexeyRaga: Yes, that's how it works.
The first parameter of `orElse` is applied when the second parameter is `None`
```fsharp
    let inline orElse ifNone option =
        match option with
        | None -> ifNone
        | Some _ -> option
```

So yes, it is "take value from the attribute, then from the context"

src/Hedgehog.NUnit/AutoGenConfig.fs
AlexeyRaga: Yes, it does work for both properties and methods. I checked that before (when moving this code into this repo), and just triple-checked.
It works because the property gets compiled with a getter method on an instance.

Issue #450, opened 2023-10-06
Support global deterministic seed


cmeeren:
Not sure if this is feasible, but I'll try:

I like deterministic builds. Being able to easily re-build and re-deploy a previous build has saved me many times. (Let's keep aside the roll-forward vs. roll-back debate. It's irrelevant anyway; the below also applies to a roll-forward.)

I have at times been thwarted by Hedgehog properties failing on CI builds where it did not fail locally, and where re-running makes it pass. Or vice versa; failing locally, but unexpectedly passing on CI. The reason is that a rare combination of generated values surfaces a bug (most often this is in the test/generator code, though occasionally it is in the production code). This combination does not occur on every run.

I would like to bring determinism back to my test suite, by having Hedgehog use the same seed every time, so to speak. So that, sure, when it runs a property 100 times, the values are seemingly random, but subsequent runs of the test produces the same values for each respective iteration. I want to be able to run a test as many times as I want, and never get different results, for the aforementioned reasons.

The only conceptual drawback I see to this is that making the seed random on every run increases the chance of catching a bug. But personally, that is not a way I want to catch a bug. It indicates a poorly written property that does not sufficiently try to reproduce relevant edge cases. I value deterministic builds more. If I want increased confidence in a property, I just increase the property's test count.

moodmosaic:
Thanks! The direction looks good to me. Adding an optional `Seed` to `IPropertyConfig` and threading it through `PropertyArgs.init` is exactly the right approach.

PR #482, opened 2026-03-23
Make seed part of config and pass into recheck data

njlr:
* Adds the type `SeedConfig` for defining starting seed configuration
 * Extends `IPropertyConfig` with a `SeedConfig`
 * Adds helpers for setting the `SeedConfig` in a `IPropertyConfig` 
 * Uses `SeedConfig` to set initial `RecheckData` in `reportWith`
 * Fixes seed in `PropertyConfig.defaults`
 * Adds a snapshot test for determinism

Relates to https://github.com/hedgehogqa/fsharp-hedgehog/issues/450

### DIFF
--- a/src/Hedgehog/Property.fs
+++ b/src/Hedgehog/Property.fs
@@ -175,7 +175,7 @@ module Property =
         let applyWithExceptionHandling (j, outcome) =
             try
                 (j, outcome |> Outcome.map f)
-            with 
+            with
             | :? TestReturnedFalseException ->
                 // Don't include internal exception in journal - it's just a signal.
                 (j, Failure)
@@ -210,12 +210,12 @@ module Property =
     let private bindGen
             (f : 'a -> Gen<Lazy<PropertyResult<'b>>>)
             (m : Gen<Lazy<PropertyResult<'a>>>) : Gen<Lazy<PropertyResult<'b>>> =
-        
+
         // Use GenLazy.bind pattern (like before async was introduced) but handle PropertyResult.
         m |> GenLazy.bind (fun propertyResultA ->
             // This function is called with the FORCED PropertyResult (lazy was forced by GenLazy.bind).
             // This happens during tree construction for proper shrinking via Gen.bind.
-            
+
             match propertyResultA with
             | PropertyResult.Sync (journalA, outcomeA) ->
                 // Synchronous case: pattern match and continue (original behavior).
@@ -243,7 +243,7 @@ module Property =
 #else
                 // Block to get the result (this happens during generation phase).
                 let journalA, outcomeA = Async.RunSynchronously asyncResultA
-                
+
                 // Now handle just like the sync case.
                 match outcomeA with
                 | Failure -> shortCircuit journalA Failure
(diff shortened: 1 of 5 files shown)


tests/Hedgehog.Tests/GenTests.fs
```diff
@@ -336,4 +356,25 @@ let genTests = testList "Gen tests" [
         firstValues.Count > 10 |> Expect.isTrue
         secondValues.Count > 10 |> Expect.isTrue
         thirdValues.Count > 10 |> Expect.isTrue
+
+    testCase "Seed is deterministic" <| fun () ->
```
moodmosaic, line 367: This is great! Though I think more cases are warranted:

- same fixed seed, same results
- same fixed seed in sync and async path give same results
- `withRandomSeed` differs from fixed-seed behavior
- `recheck` path preserves configured seed semantics

src/Hedgehog/PropertyConfig.fs
```diff
@@ -1,17 +1,32 @@
...
+
 module PropertyConfig =
 
     /// The default configuration for a property test.
     [<CompiledName("Default")>]
     let defaults: IPropertyConfig =
         { TestLimit = 100<tests>
-          ShrinkLimit = None }
+          ShrinkLimit = None
+          SeedConfig = FixedSeed (Seed.from 0UL) }
```
moodmosaic: This changes the default config from “effectively random every run” to always deterministic.

I'd have to thoroughly check but if the shrinking/recheck model depends on initial seed plus shrink path, changing default seed behavior affects:

- reproduced counterexamples
- golden tests
- user workflows that expect “rerun gives new samples”

---

This is a significant semantic change. Every user who calls Property.check, Property.checkBool, or any function that flows through PropertyConfig.defaults will now:

- Always test the same inputs across runs (unless they opt into `withRandomSeed`)
- Potentially miss bugs that only manifest with certain random seeds
- Get different behavior from what they've been relying on without changing any of their own code

AFAIK, in most property-based testing systems, the default is random seeds with the seed printed on failure for reproducibility.

The deterministic-by-default philosophy might exist, but flipping an established default can be surprising to the users, if not risky.

src/Hedgehog/Property.fs
```diff
@@ -476,7 +476,8 @@ module Property =
     /// The report includes the number of tests run, discards, and failure information with shrunk counterexamples.
     /// This blocks until all tests complete.
     let reportWith (config : IPropertyConfig) (p : Property<unit>) : Report =
-        p |> reportWith' (PropertyArgs.init ()) config
+        let seed = SeedConfig.init config.SeedConfig
```
moodmosaic, line 479: That's good: it means only one concrete seed per property run, not reevaluating randomness multiple times.

moodmosaic (changes requested):
Thank you for the PR!

The core feature is a good addition, but the default should be RandomSeed to preserve existing behavior. The withSeed API lets users opt into determinism explicitly, which is the right trade-off for a property testing library.

tests/Hedgehog.Tests/GenTests.fs
njlr: Sorry for delay. Hopefully all addressed now?

moodmosaic: Thanks! I'll do another round of review 🙏 👍

moodmosaic (approved):
Thank you, @njlr 👍 This is in good shape to merge.

PR #485, opened 2026-06-25
Add SourceLink

### LINKED ISSUE #484: SourceLink
Would be great if the package supported [SourceLink](https://github.com/dotnet/sourcelink) so that I can browse the original F# sources when debugging.

That is all.

jwosty:
Resolves #484

* Changes rollForward policy in global.json to allow repo to be opened with any .net sdk > 8.0.416
* Adds SourceLink, via embedded pdb to match Hedgehog.Xunit and Hedgehog.Nunit (turns out they already had sources bundled): https://github.com/dotnet/sourcelink#embed-in-assembly

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "rollForward": "disable",
-    "version": "8.0.416"
+    "version": "8.0.416",
+    "rollForward": "latestFeature"
   }
 }
--- a/src/Hedgehog/Hedgehog.fsproj
+++ b/src/Hedgehog/Hedgehog.fsproj
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
+    <IsPackable>true</IsPackable>
     <TargetFramework>net8.0</TargetFramework>
     <Version>2.0.3</Version>
     <Description>Release with confidence.</Description>
@@ -8,6 +9,8 @@
     <Copyright>Copyright © 2017</Copyright>
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <DebugType>Embedded</DebugType>
+    <EmbedAllSources>True</EmbedAllSources>
     <PackageId>Hedgehog</PackageId>
     <PackageDescription>
 Hedgehog automatically generates a comprehensive array of test cases, exercising your software in ways human testers would never imagine.


global.json
```diff
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "rollForward": "disable",
-    "version": "8.0.416"
+    "version": "8.0.416",
+    "rollForward": "latestFeature"
```
moodmosaic, line 4: Do we need to change this?

src/Hedgehog/Hedgehog.fsproj
```diff
@@ -1,13 +1,16 @@
...
+    <IsPackable>true</IsPackable>
     <TargetFramework>net8.0</TargetFramework>
     <Version>2.0.3</Version>
     <Description>Release with confidence.</Description>
     <Authors>Jacob Stanley;Nikos Baxevanis</Authors>
     <Copyright>Copyright © 2017</Copyright>
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <DebugType>Embedded</DebugType>
+    <EmbedAllSources>True</EmbedAllSources>
```
moodmosaic, line 13: Also consider adding `<PublishRepositoryUrl>true</PublishRepositoryUrl>` and (CI-gated?) `<ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>` for deterministic, properly-pathed builds, as I think the latter is what makes SourceLink links actually resolve correctly in published packages(?)

moodmosaic (commented):
Thank you!

Does this rely on the .NET 8 + built-in SourceLink, or should Microsoft.SourceLink.GitHub be referenced explicitly?

global.json
jwosty: For the question of SourceLink support, no. I just ran into an issue building it on my machine immediately after cloning because I had .NET 8.0.422 installed so it failed because it was too high. I could have installed a lower .NET 8, but I figured I'd give a courtesy contribution to reduce friction for new contributors a bit. I'm happy to revert if requested.

src/Hedgehog/Hedgehog.fsproj
jwosty: Hm, this is actually my first time hearing about that property and I've published a few packages with SourceLink (and seen some others) working just fine without that. I'm happy to add that if you'd still like me to do that though.

moodmosaic: Yeah we can leave those out. Good point.

global.json
moodmosaic: Yeah that's a good point actually 💯

moodmosaic:
@jwosty, thank you! What are the next steps?

jwosty:
> Thank you!
> 
> Does this rely on the .NET 8 + built-in SourceLink, or should Microsoft.SourceLink.GitHub be referenced explicitly?

Yep, it uses the built-in sourcelink and doesn't need to reference any packages.

> @jwosty, thank you! What are the next steps?

Nothing - it should work next time it is deployed

